PR #24, opened 2017-01-25
Adding new model based on hpyerparameter tuning and adding some L2 re…

dangerwheeler:
…gularization


dangerwheeler:
D'oh! This is not the model I wanted to share. I'll close and resubmit.

Issue #23, opened 2017-01-25
Update build instructions. 


wroscoe:
Include
* How to pick a car + tested models. 
* link to ami image
* Wiring diagram + pictures

yconst:
Are the current build instructions up to date? Installing to a RPi2 via the docs at https://github.com/wroscoe/donkey/blob/master/docs/01-build_a_car.md and then trying to run drive_pi.py shows an error about missing docopt. Any ideas?

wroscoe:
docopt should be installed when you run `pip install -e ../donkey[pi]` in your virtual environment.

That error should be fixable by installing docopt  in your virtual environment
`pip install docopt`

I'm updating the build instructions to be clearer but the current version should work.

wroscoe:
@yconst  Here is the in progress build doc. Please suggest update and add pictures. https://docs.google.com/document/d/11IPqZcDcLTd2mtYaR5ONpDxFgL9Y1nMNTDvEarST8Wk/edit#

yconst:
Thanks for sharing! I'll sure chime in if I have any suggestions, although I'm still getting to grips with the entirety of the project!

alanwells:
@wroscoe Thanks for this great project! I've got my car built up (based on an [ExceedRC Magnet 1/16 scale truck)](http://www.nitrorcx.com/51c853-savared-24-ghz.html) and steering mostly works (angles are a little off wheels move left/right), but throttle doesn't work at all. The error code on the ESC indicates abnormal throttle signal, so I'm thinking it's a PWM calibration issue. I noticed there was some stubbed out calibration code in the `actuators.py`, do you have any suggestions on where to start on figuring out the correct values? 

Also, I have a few notes from going thru the donkey install process on car and server. Do you prefer to receive them as suggestions to the Google Doc or in comments here?

Here are a few pics of the build. The 1/16th scale is a bit cramped but it worked out OK. Happy to take better pics for the build document if it's helpful:

![img_0435](https://cloud.githubusercontent.com/assets/1134287/22678759/628aad02-ecb1-11e6-86b9-80c2f3f9e5ca.jpg)

![img_0437](https://cloud.githubusercontent.com/assets/1134287/22678767/6f2b16d2-ecb1-11e6-9ce4-0fc7404400d7.jpg)

wroscoe:
That car looks awesome. 

[...]

Issue #21, opened 2017-01-22
Move session handling inside remote server.


wroscoe:
Currently sessions are passed to the remote sever on creation. This limits the server to only handling one session at a time and prevents switching. 

* updates required to scripts/serve.py & donkey/remotes.py

Issue #20, opened 2017-01-22
Implement single mixer class instead of separate throttle/steering classes


yconst:
A single "mixer" class that handles command distribution to all actuators is more flexible than separate "throttle" and "steering" classes. 
For instance in differential or skid steering one should be aware of both the throttle as well as the steering value to assign correct (PWM) values to the actuators

wroscoe:
Good point. I think the best place to add this would be as a new vehicle class that provided this awareness for skid steering vehicles in the drive loop. This library has so far focused on cars that steer by angling the front wheels but this would be a great addition. @adammconway may also have thoughts.

adammconway:
Hi @yconst are you considering implementing a car based on differential drive using Donkey? am I understanding the request properly? thanks!

yconst:
@adammconway yes indeed, I have a differential steering frame available and am thinking of "emulating" the constraints in "regular" car steering by limiting the difference in throttle to the motors on each side/making it proportional to collective throttle.

adammconway:
@yconst that is awesome, so glad you are looking to join us.  

@wroscoe feels like differential drive should be a different setup... this goes back to how we abstract out the various hardware.  Downstream if we choose to add odometry and other sensors we will also have further differences.  I know that sensors and actuators are separate now, but I think it is possible we will see interplay between them as we add certain sensors.

wroscoe:
@yconst this would be great. We were looking to find a vehicle that could compete on the RGB(colored lines) track. 

The place to put a new vehicle build would be below the BaseVehicle class here; https://github.com/wroscoe/donkey/blob/master/donkey/vehicles.py#L3 You could inherit the BaseVehicle class and change the drive loop to use the logic specific to the differential drive vehicle. 

@adammconway There is no real difference between sensors and actuators. The files are used only for organization.

yconst:
@wroscoe , @adammconway I took the liberty of modifying the actuator code to accommodate for mixing. I've pushed the changes in my fork currently:
https://github.com/yconst/donkey/blob/actuator_mixer/donkey/actuators.py

I'm not sure, but I felt an additional level of abstraction between "vehicle" and "hardware" would make for better distribution of tasks. Thus came the mixer class.

There are also changes to the vehicle and demo classes to accommodate the new actuators class. 

Please keep in mind, this is all just a concept, nothing is tested yet.

wroscoe:
Great!. I made a pull request for this change and will merge once we have both types of cars working on this branch.

[...]

Issue #19, opened 2017-01-19
Web: enable auto pilot switching. 


wroscoe:
Currently the only way to change auto pilots is by restarting the server using different CLI variables. This is slow and doesn't facilitate quick iteration.

A faster way to test poilots would be to switch them from the control webpage. To do this the following would need to be changed. 
* remotes.py - change pilots from a singleton to a dictionary, include model and description
* monitor.html - add pilots control.  
* main.js - add ajax to update current pilot.

PR #29, opened 2017-01-28
Add differential drive and update docs.

### LINKED ISSUE #27: Refactor imports?
Donkey is importing all modules in the donkey/ directory, which leads to a lot of unneeded module imports, and breaks in many cases (such as running drive_pi.py, which requires envoy and Keras, even thought they are not used). Can this be architected differently so that only modules needed are imported?

wroscoe:
* Creates an abstraction of mixers over actuators to support differential drive vehicles (untested).
* Moves pilot and session control into the web server.
* Adds web interface to switch pilots
* Adds web interface to view sessions and delete photos.
* Fixes joystick so full throttle is possible at sharp angles.
* Moves most documentation to a google doc
* Rename demos folder to scripts
* Add line detection implementation of CompoundEye using opencv

Issues
fixes #27 #23 #32 #21 #20 #19 #1


wroscoe:
Session view.
image


Pilot Switching in vehicle monitor
image

PR #39, opened 2017-02-13
Add hookup to tawnkramer/sdsandbox simulator

alanswx:
I think it will be useful to test our models on a simulator.  This makes the simulator seem like a virtual car.


wroscoe:
This merge had some of the big model files that I just deleted so created a separate branch with the joystick code and simulator code and merged it instead.

PR #40, opened 2017-02-13
Support differential drive vehicles.

wroscoe:
Currently donkey only supports mechanical steering vehicles like RC cars. This pull request:
*  refactors the mixers
* adds actuators for DC motors using Adafruit motor hat
* normalizes throttle and angle values to be between -1 and 1

Example differential drive vehicle:


wroscoe:
It works.
gif 1

PR #44, opened 2017-02-21
Session image select UI updates

alanwells:
This PR adds:
- a more performant method of selecting multiple images on the session page
- select all capability for images on the session page
- removes a selector that always marked the `Pilots` navigation tab as active.

@wroscoe Looks like my editor has a different whitespace setting than yours. Let me know if you'd like me to redo this without the whitespace changes.

### DIFF
--- a/donkey/remotes.py
+++ b/donkey/remotes.py
@@ -150,7 +150,7 @@ class DonkeyPilotApplication(tornado.web.Application):
 
         settings = {'debug': True}
 
-        super().__init__(handlers, debug=True)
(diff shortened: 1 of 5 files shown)


wroscoe:
Thanks Alan. This looks good but how do we avoid the white space / non-changes in the future. It's hard to see what actually changed with all that other noise?

alanwells:
I think it must be a setting in my editor that is conflicting with whatever your editor does for linebreaks. I'll close this and submit a clean PR after I can figure out what caused it.

PR #53, opened 2017-02-28
Draggable multiselect and turn icons

alanwells:
@wroscoe I added back the click-to-drag multiselect capabilty with some filters that make it more performant. This feels pretty snappy on my computer for session pages with up to 500 images. Beyond that it lags pretty badly, so I updated the pagination code to break the pages at 500 instead of 1000.

Also adds iconography to indicate turn direction (requested by @adammconway).

### DIFF
--- a/donkey/remotes.py
+++ b/donkey/remotes.py
@@ -444,7 +444,7 @@ class SessionView(tornado.web.RequestHandler):
         imgs = [dk.utils.merge_two_dicts({'name':f.name}, dk.sessions.parse_img_filepath(f.path)) for f in os.scandir(path) if f.is_file() ]
         img_count = len(imgs)
 
-        perpage = 1000
+        perpage = 500
         pages = math.ceil(img_count/perpage)
         if page is None: 
             page = 1
--- a/donkey/templates/session.html
+++ b/donkey/templates/session.html
@@ -13,20 +13,18 @@
 			
 			<div class="row">
 				<div class="col-md-12 session-thumbnails">
-					{% for img in session['imgs'] %}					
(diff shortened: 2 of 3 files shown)


alanwells:
@adammconway here's a screenshot of the icons to indicate turn direction. does this match what you were thinking of in your comment from sunday?

screen shot 2017-02-28 at 7 59 15 am

wroscoe:
@alanwells What you have is an improvement but I think something on the image would be easier to debug. See otavios debug screen. I've found that on my computer, looping through the images is the easiest because you can see the motion. 

https://dl.dropboxusercontent.com/u/21554/car/2016/debugSteerColor.html

adammconway:
Alan, that is a big improvement.  It will make it much easier to go through the data.  
Cheers,
Adam

adammconway:
@wroscoe Hey Will, I like what Otavio did as well.  It shows the wired outputs.  I learned that while doing non-track training (like the sidewalk) seeing and managing the wired outputs would be nice so that you can see whether the car wants to take a left down a driveway, etc.  I know I am a broken record on this (not trying to be irritating and love the work you are doing.)

wroscoe:
When you say "wired" do you mean discreet / categorical or the line on the image showing the steering direction?

adammconway:
I mean categorical.  The little bar chart on the bottom.  Maybe you already made that change...

alanwells:
@wroscoe @adammconway Otavio's debug screen is slick. UI improvements to the vehicle control page for training seem more urgent to me (at least a quick first pass on better formatting for small screens) but definitely would like to circle back to the session page at some point in the future.

PR #56, opened 2017-03-02
Changes in Dockerfile and startup script

yconst:
I've modified the Dockerfile to mount the donkey dir from the local fs, instead of git cloning it within the container. The advantage is that local changes are reflected in the container after restarting it. Check Dockerfile comments for more.

I've also added a convenience script to build & run the container.

So now the only command to build & run donkey server with Docker is:
`./start-server.sh`

Any comments/improvements more than welcome.


wroscoe:
wow, this sounds great! @alanwells can you test on iOS?

wroscoe:
This works well on ubuntu. It takes ~15 min when I run it the first time and ~15 seconds when I run it the second time.

wroscoe:
@yconst  This is a docker noob question. Can I access the docker directories from outside the docker instance. For example I'd like to add some models I have saved on my local machine to the mydonkey/models of the docker instance. How do I do this?

alanwells:
@wroscoe I'm a docker noob too, but I looked into that question and didn't find a way to do that. My workaround was to mount directory from outside the docker instance as a volume in the docker instance. using the `-v` arg on `docker run`, for example:
```
docker run -p 8886-8887:8886-8887 -v /Users/alanwells/Projects/Personal/donkey:/root/donkey -it  alanwells/donkey-server-jessie bash
```

alanwells:
@wroscoe confirmed the updated dockerfile works on OS X, but I think you have to be careful running via the `./start-server.sh` script. The script creates a new container every time, which means that you'll lose all your training data and models that were created inside the container unless you manually copy them out of the docker instance.

@yconst what's the right way to use for multiple server sessions without losing training data and models? Do you recommend restarting a stopped container that was previously created, or is there a way modify  `./start-server.sh` so that it doesn't blow away the data in the container?

wroscoe:
@alanwells Good catch. I already merged it because it is the quickest way to get a server going. In the instructions I'll keep the local server version as an option. Another way around this issue would be to add download links for the models and sessions created in the web app.

yconst:
@wroscoe @alanwells Indeed maybe it's better to have a parameter in the script to force rebuild, otherwise conditionally build the container, e.g. sth like docker run -p 8887:8887 donkey || docker build ....

Re. sharing, maybe [Data Volumes](https://docs.docker.com/engine/tutorials/dockervolumes/) can be a solution? I'm a Docker noob myself too, but it kinda seems as the thing we're looking for for sharing stuff outside the container.

I'll try to look into these issues during the weekend.

Thanks for the feedback!

yconst:
As much as I'd like to solve the data sharing issue within the Dockerfile (via the VOLUME directive), it seems this is currently unsupported. So for now we might have to resort to having the -v /path/in/container:/path/in/host flag when running the container.

yconst:
@alanwells also I think in your example we no longer need the -it flag, as the serve.py script is run via the ENTRYPOINT directive, which makes the script the default executable in the container. Unless of course we still wish to have regular shell access.

yconst:
So turns out there are two ways to persistence in Docker containers:

1. What @alanwells suggested, i.e. adding the -v flag in the startup script. This also allows us to specify which directory will be mounted in the container, so that we may have a convenient location on the host to store models/images etc.

2. The VOLUME directive in the Dockerfile. This is IMO more portable, but has the drawback that we may not specify the folder location in the host, i.e. it will be assigned by Docker (but its location is sill unchanging, so data can be persisted).

I'm planning to make this fixes/changes, so @wroscoe @alanwells  if you have any ideas do share.

wroscoe:
@yconst The volume approach would require less setup which is good. Can you print the location of the volume on docker startup? Its important that the user find his previous sessions/models when he starts up a docker image a second time. Can we specify the volume to be the same place each time?

yconst:
@wroscoe The dir where the VOLUME is stored is the same each time the docker image starts. This is to ensure persistence. However, I don't know what happens if the container is rebuilt. I believe we should avoid this anyway, as mentioned above.

wroscoe:
Yep. This sounds practical.

PR #64, opened 2017-03-05
Small screen UI with control by deviceOrientation

alanwells:
@wroscoe @adammconway Here's my first pass on the small screen UI with support for steering using device orientation. I think it's ready for testing but expect that we'll want to tune the dead zone parameters here before merging to master.

Also, there is some inconsistency on how the start/stop vehicle button is handled. Currently, if the vehicle is in autopilot, the state of that button is not respected. I was going to hook up some client-side logic to revert to user pilot and set throttle + angle to 0,0 when the stop button is hit, but that feels annoying if you just want to briefly stop your car and resume with the same pilot settings. I think the right way to get consistent behavior from the brake button here would be to send the `brakeOn` state to the server, and have the server set throttle and steering to 0,0 whenever the brake is on (regardless of pilot type).

I haven't had time to work on visual indicators for angle/throttle, as I wanted to focus on getting the control by device orientation hooked up first. I'll work on those in a subsequent PR.


wroscoe:
I've merged this into dev but I've noticed one issue on my phone.  My phone is in landscape view and I'm using Chrome on Android.

When I click "Start Vehicle" with the phone tilted slightly toward me the angles are around .9 and decrease when I tilt the phone more toward me and max out at 1 when I tilt the phone away from me. 

When I click "Start Vehicle" with the phone tilted slightly away me the angles are around -.9 and decrease when I tilt the phone more away from me and max out at -1 when I tilt the phone away from me.

alanwells:
@wroscoe that's interesting, I think that means the deviceOrientation output of your device is the inverse of my iOS device. Could you capture a screenshot that shows the deviceOrientation debug info on the bottom of the page when the controls are showing .9 and -0.9?

alanwells:
@wroscoe I tried to reproduce this on my Nexus 7 tablet running Chrome and wasn't able to, so screenshots would be really helpful here. Also, could you try rotating your device 180º around the Z axis (as labeled in this graphic) to see if it corrects the throttle controls?
1398214917alpha

wroscoe:
Rotating doesn't seem to have any affect. I'll past screenshots later but my Gamma (rotation around Y) reads:
0 when phone is flat.
-45 when phone is tilted 45 deg towards me
45 when phone is tilted 45 deg away from me.

alanwells:
@wroscoe yeah, that's the opposite of my device. I'll need to add a calibration factor to correct for it. I'm curious if it's related to device manufacturer - what type of phone are you using?

wroscoe:
I'm using a Moto Droid 2. 

So in your phone, when you lay your phone flat, is gamma 90?

wroscoe:
I just checked on Alan S. iPhone (who is sitting next to me at this conference) and it is doing the same thing. The gamma <> throttle mapping doesn't allow for a continuous range between -1 and 1.

alanwells:
@wroscoe I get gamma 90 when the phone is flat (parallel to the ground) but my gamma values at 45 degrees are the opposite of yours. When I hold my iPhone in landscape orientation with the home button closest to my right hand, I get the following throttle and gamma values:

- Phone perpendicular to ground: gamma 90, throttle 0
- Phone tilted 45 degrees towards me: gamma 45, throttle -.40 
- Phone tilted 45 degrees away from me: gamma -45, throttle .40

I can get a nice continuous range of -1 to 1 by going from the phone parallel to the ground (screen down, where throttle = -1), rotating it away from me (upwards - through perpendicular, where throttle = 0) to parallel to the ground (screen up, where throttle = 1).


When I flip the phone to landscape orientation with the home button closest to my left hand, I get the following values (which seem to match yours) but I still get a continuous range between -1 and 1 (it's just the opposite of above, tilting the phone away from me decreases throttle instead of increases it).
- Phone perpendicular to ground: gamma 90, throttle 0
- Phone tilted 45 degrees towards me: gamma -45, throttle .40 
- Phone tilted 45 degrees away from me: gamma 45, throttle -.40

But you're saying if you flip your phone to the other landscape orientation, it doesn't fix the problem?

PR #68, opened 2017-03-11
Small screen UI - 2nd revision

alanwells:
- Adds visual indicators for throttle and steering
- Adjusts device orientation steering calculations to be relative to starting point instead of hardcoded device positions
- Toggle for switching between joystick and device orientation control modes
- Steering and throttle angles now behave as expected in both landscape orientations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ data/*
 .ipynb_checkpoints
 *.egg-info/
 .git.bfg-report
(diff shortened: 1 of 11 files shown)


alanwells:
@wroscoe there may be some conflicts here with the latest gamepad PR. If that's the case, let me know and I'll circle back to resolve any conflicts that arise after you merge that one.

PR #69, opened 2017-03-11
Max throttle UI control

alanwells:
Enables user to set a max throttle not to be exceeded, retains user throttle control from 0 to max throttle. Once we have odometry implemented, I think we could change this to a control that keeps the vehicle at a fixed velocity. cc @adammconway

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ data/*
 .ipynb_checkpoints
 *.egg-info/
 .git.bfg-report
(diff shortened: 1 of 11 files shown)


wroscoe:
@alanwells these conflicts look like just duplicate commits from the last pull request. Can you fix?

alanwells:
@wroscoe sorry about that, fixed.

PR #96, opened 2017-07-15
Donkey2 with Odometer

alanwells:
@wroscoe Not sure whether you want to pull an example like this into the main repo. Thought I'd open it up the PR so you can tell me whether I followed the patterns as you were thinking.


wroscoe:
I think in this alpha it makes sense to have everything in one place. Later we can figure out how to break out the add-on parts into their on library or package system. 

The code is super clear. Thanks for contributing.

PR #101, opened 2017-07-27
Marcos dev branch

wroscoe:


wroscoe:
I updated the tub to use json in a bit different way than you did so now there is a conflict. I'll try to fix and merge this tomorrow.

PR #113, opened 2017-08-22
Fixed issue with keras.py and tub.py to allow it run

ricorx7:
I had problems with the code running from a clean install.  It did not work with 'val_loss' in keras.py.  Had to change to 'loss'.  In tub.py, when you write the JSON data, it throws an exception and crashes.  I put a try/except to ignore when it gives and error for now.  Not sure what the value is suppose to be yet.  

I updated my code to latest and it made the pull request now 65 files.  Not sure what happen there.

I also documented here a little more instructions to get Donkeycar 2 working.
You can take what you need from it to update your documentation.
http://www.sroboto.com/2017/08/donkeycar-2-install-setup-and-run.html


wroscoe:
I can't view what has changed because it looks like everything has changed. https://github.com/wroscoe/donkey/pull/113/files  I appreciate you submitting changes but they have to be more incremental for me to merge them. Also, I fixed the val_loss bug recently.

PR #117, opened 2017-09-05
some small fixes, adds joystick part

tawnkramer:
There's a few files in the root you probably don't want. But the changes in donkeycar may be ones you'd like. There's probably a better fix to the json serialization crash than commenting out the pilot output in the template scripts. Let me know if you have questions. -Tawn


wroscoe:
This looks good @tawnkramer. What's the rightway to include the sixpair in the Joystick part but have it as an optional part. Maybe the `donkey install could install the part?

tawnkramer:
Oh shoot. I forgot about sixpair. I included that for my own convenience when getting things running for people. It's only needed once on setup. I'm open to whatever might work in your framework. It could be invoked from donkey maybe. Like donkey sixpair. But it needs to be compiled, and has a dependency on libusb. So not simple. 

If you want to pass on that, we can give it some thought for the future.

tawnkramer:
Sixpair sends the address of the Bluetooth host controller to the PS3 sixaxis so it knows to pair with that Pi. That only needs to be done once. Maybe not as a part. Doesn't need to run each time. Joystick.py is self contained and doesn't rely on that.

wroscoe:
Thanks. I saw @ricorx7 method to use curl.

wroscoe:
Also what are the steps to install on windows? Something to go here. http://docs.donkeycar.com/guide/install_software/

wroscoe:
This fork will be added to [dev branch](https://github.com/wroscoe/donkey/tree/dev) then merged to master.

tawnkramer:
Bottom half of this page, from Windows Donkey Setup on down. Requires the
win_environment.yml file.

https://github.com/tawnkramer/donkey/blob/master/windows_howto.md

PR #119, opened 2017-09-09
Ricorx7 fork

wroscoe:


wroscoe:
These changes were very close to Tawns that I merged into #dev. I added the joystick install docs here. https://github.com/wroscoe/donkey/blob/dev/docs/parts/controllers.md

PR #129, opened 2017-09-12
Tub view

kennethjiang:
Usage: `donkey tub /path/to/data/folder`


wroscoe:
Before merging this we need to find a way to share the web libraries like jquery and bootstrap between the controller and this. I'm thinking all these could live in folder web_assets in the root.

wroscoe:
Also, all the docopt CLI commands were switched back to car.py. These should remain manage.py.

kennethjiang:
Since donkey2 adopted a very modulized design, sharing static assets between tub viewer and web controller would introduce coupling that may bite us down the road. For instance, now tub viewer can't upgrade to bootstrap without impacting web controller and vice versa. Also mixing custom javascripts between tub viewer and web controller may introduce some unintended naming conflicts.

Keep asset folders to themselves only cost some extra storage on disk, which is an acceptable tradeoff IMO.

kennethjiang:
Also just pushed a couple commits to add hotkey. It is pretty usable now. :)

PR #133, opened 2017-09-15
fixed mac env yml and conda install instructions

tawnkramer:


tawnkramer:
Thanks for integrating this. There was a problem installing Tensorflow that this fixes. Is there something I can do to push this fix to the docs website?

PR #135, opened 2017-09-16
Add more button controls for PS3 controller

alanwells:
Recording is manually toggled in this version. Additional controls added:

        button map name => PS3 button => function
        * top2 = PS3 dpad up => increase throttle scale
        * base = PS3 dpad down => decrease throttle scale
        * base2 = PS3 dpad left => increase steering scale 
        * pinkie = PS3 dpad right => decrease steering scale
        * trigger = PS3 select => switch modes
        * top = PS3 start => toggle constant throttle
        * circle = PS3 circrle => toggle recording
        * triangle = PS3 triangle => increase max throttle
        * cross = PS3 cross => decrease max throttle

### DIFF
--- a/donkeycar/parts/controllers/joystick.py
+++ b/donkeycar/parts/controllers/joystick.py
@@ -211,6 +211,7 @@ class JoystickPilot():
         self.steering_scale = steering_scale
         self.throttle_scale = throttle_scale
         self.recording = False
+        self.constant_throttle = False
 
         #init joystick
         self.js = Joystick(dev_fn)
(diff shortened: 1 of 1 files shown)


wroscoe:
@tawnkramer I'm unable to test/check this pull request as I don't have a controller. Please merge it when you're ready. Thanks.

alanwells:
@tawnkramer some follow up thoughts to your comments:
- The issue that I hit with the automatic recording is that I often want to do a few practice laps (unrecorded to avoid having to clean up the data later) before I start capturing training data. I know I can just stop the drive script and restart so that a new tub gets created (making it easier to delete the practice laps), but it's nice not to have to stop. I agree w/ you about it being a huge bummer when you thought you were recording and it turns out you weren't. How about something like the following:
  - Revert to the behavior that you had as the default (recording starts when you drive and stops when you stop, automatically)
  - Add a toggle to enable/disable this automatic recording feature. This would allow for disabling it when desired, but only if you choose to flip the toggle. This would leave the safer (automatic) behavior as the default but still allow for disabling recording when you don't want it on.

- I love the idea of an LED part that indicates recording status. I think that could be done in a separate PR, but would be happy to add it here if you'd like to see it as part of this change.
- Regarding counting number of images - I don't think we need a new part from this. I think we can get the image index value out of the tub and just print it to console inside the tub part?

PR #136, opened 2017-09-17
incorrect text commands, formatting issues

adammconway:


adammconway:
Quick text edits to docs

PR #139, opened 2017-09-18
Rotary encoder updates

alanwells:
- Define tick distance in millimeters instead of meters
- Fixes a bug where velocity would not be correctly calculated when vehicle was stopped.
- Print the top speed to console at shutdown
- Add debug flag for console output


alanwells:
@wroscoe just noticed you dropped the rotary encoder template from the repo. No need to re-add, I just didn't want it to be outdated (and incompatible with the fixes I checked in for the rotary encoder). I'll push up a commit that reverts the addition.

wroscoe:
Great. I think the right place to put it right now is in a gist and make an example cars page in the docs. Its a pain to have to update each template if some part of the API changes. We can figure out a better way to save / update templates later. Thanks

alanwells:
@wroscoe ok, removed. thanks!

tawnkramer:
Joystick template could be a gist too. Right now just a single line change
to main template.

wroscoe:
We could also put these in the main donkey2 image but commented out.

tawnkramer:
Makes sense. I will do that.

gang65:
Hello @alanwells.
I just notice that you have implemented Rotary encoder, which I would like to add to my Donkey Car.
Could you please explain where I could find latest source code of this part (I noticed that source code was removed)?
What kind of hardware did you use to build Rotary encoder?

My general idea, it to use Rotary Encoder values (Velocity) as a Input for Convolutional neural network:
https://github.com/wroscoe/donkey/blob/dev/donkeycar/parts/keras.py#L86

I expect the NN to output more accurate angle and throttle if we provide actual velocity as an additional input for the network.

alanwells:
@gang65 @wroscoe here are the parts I used:
- This rotary encoder from Sparkfun: https://www.sparkfun.com/products/10932
- A 3.3V to 5V level shifter like this: https://www.adafruit.com/product/1875
- Motor mount from polulu (has to be modified a bit with drill or dremel to fit above encoder): https://www.pololu.com/product/2676
- Plastic gear to mate with the existing gear on the Exceed Magnet driveshaft: https://www.mcmaster.com/#57655k24/=1e0ioui
- Tool steel (cut a short portion to connect the plastic gear above to the rotary encoder using the plastic coupler that is included with the encoder): https://www.mcmaster.com/#3023a223/=1e0ipde

My setup looks like this after assembly:

img_1062

alanwells:
@gang65 original reply above had the wrong McMaster link. Correct link is https://www.mcmaster.com/#57655k25/=1e0itla (also edited in msg above).

alanwells:
@gang65 I haven't been racing or contributing code lately so I haven't made any attempts to keep the rotary encoder part working as the Donkey software has evolved. You can see how I was using it in my fork (which is very far behind this repo now):
- part code: https://github.com/alanwells/donkey/blob/master/donkeycar/parts/sensors/rotary_encoder.py
- drive loop using rotary encoder to log velocity and use velocity to set throttle via a PID: https://github.com/alanwells/donkey/blob/master/donkeycar/templates/donkey2_aw.py#L18

jadonk:
For a cheaper solution, can anyone on this thread give some suggestions on how something like one of https://www.pololu.com/category/201/encoders would be integrated mechanically?

PR #143, opened 2017-09-19
Nasty bug in multi-tub training

kennethjiang:
This is to fix a nasty bug that prevented training from moving beyond the 1st tub. It was probably the reason why your model didn't work @wroscoe, and none of the Donkey cars did well Saturday. I feel so bad about it!

Also this fix would require everyone to regenerate `manage.py`.

One drawback of this implementation is that up to `batch_size` records in each tub may not be used at all. This is because now `get_record` will throw `StopIteration` exception when there aren't enough records to fill a whole batch so that `itertools.chain` can move on to next tub.


wroscoe:
This pull request has much more than your proposed change. https://github.com/wroscoe/donkey/pull/143/commits/4280ae4a09a65f34f19b64b6e6ab1d8fefc9f977#diff-d2e1a798b3ef27676227316c5ebdcf71R191

 Also the proposed change would stop training after the data is run through once. 

Can you fix and resubmit?

kennethjiang:
Sorry I completely screwed up this PR. Had to open another one #144 using a different branch. Sorry for the sloppiness.

PR #145, opened 2017-09-21
Fix a crash when simulation tries to load a model

kennethjiang:
Again this bug fix requires users to re-generate `manage.py` (only if they are using simulation)


kennethjiang:
I am piggybacking this PR for a couple minor bug fixes too. Somehow github shows a lot of commits that have already been merged. Just switch to "File changed" to see the diffs, which should be pretty clean.

wroscoe:
more fixes are fine. Try rebasing... https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch See solution 2.

kennethjiang:
I gave up on rebasing - every time when I tried to rebase the same conflicts (like `manage.py` vs `car.py`, along with dozens of others) kept on coming back. I have run into the similar situation in other projects and we never figured out why and had to fall back to merge.

I guess I will just create a new branch and new PR.

kennethjiang:
Created a new branch but it was the same: a bunch of merged commits still show up. This might have to do with those commits being squashed when last PR was merged.

wroscoe:
If the diff is clean I'll merge this one. Next one just start from a clean clone or use rebase.

PR #148, opened 2017-09-25
Check and fix

tawnkramer:
Added a `check` option to manage.py that will report any problems with the tub records. With the `--fix` option added it will remove those problem records.


tawnkramer:
this is a fix/work around for https://github.com/wroscoe/donkey/issues/118. Not sure what caused it in the first place though.

PR #154, opened 2017-10-03
Grad-CAM visualization

kennethjiang:


kennethjiang:
Thanks to @tawnkramer 's pointer I also included another visualization method.

Also I change the command to be more general `python manage.py visualize`

tawnkramer:
is the Grad-CAM visualization working? Did you pull in the visual back prop source?

kennethjiang:
It can generate Grad-CAM-enhanced video, although the heatmap doesn't seem to make sense. However considering Grad-CAM is quite popular, I decided to keep it in here just in case some other people would get a different result, or find out what is wrong with it.

Yes everything needed for visual back prop is pulled in.

wroscoe:
@kennethjiang This looks valuable and I know users would use this to debug their models. 

Before we merge this code we need to..
- change this pull request to be into the dev branch. 
- add a couple tests so we can know if it breaks.
- confirm the output colors are accurate. 

@kennethjiang can you take this on?

kennethjiang:
Sure I can re-do the PR based off of `dev`. Also add some automated testing.

However Grad-CAM doesn't look right at all. I guess I might as well just remove it. Visual back-prop looks better. But I'm not 100% sure about it correctness either. Not really understanding its internals, I did very little more than just adopting @tawnkramer 's script to Donkey 2 structure.

PR #155, opened 2017-10-06
Back propagation visualization

kennethjiang:
I rebased to `dev` branch but not sure why there is an extra commit showing up in the list. However the diff is correct.


tawnkramer:
Is there a missing file here? What's causing validation failure

kennethjiang:
Yes some stupid errors caused by merging between branches - too many branches to manage. :(

PR #156, opened 2017-10-09
Movie from tub

tawnkramer:
added utility to create a movie from tub dir. This adds a new dependency on moviepy. I tried to get things to work with just opencv, but failed to work as well as moviepy. docs updated to include new operation, as well as a few others in the `donkey` command that needed detailing.

### DIFF
--- a/donkeycar/management/base.py
+++ b/donkeycar/management/base.py
@@ -141,6 +141,40 @@ class CalibrateCar(BaseCommand):
             c.run(pmw)
 
 
+class MakeMovie(BaseCommand):    
+    
+    def parse_args(self, args):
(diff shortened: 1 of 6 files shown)


donkeycar/management/base.py
```diff
@@ -149,6 +213,7 @@ def execute_from_command_line():
             'findcar': FindCar,
             'calibrate': CalibrateCar,
             'tub': TubManager,
+            'movie': MakeMovie,
```
wroscoe: these commands should be actions (ie. makemovie)

tawnkramer:
I agree. Makes more sense with a verb in there. That's in now.

PR #164, opened 2017-10-11
Add plot_predictions method to template

alanwells:
This makes it easy to create plots for checking model performance (very similar to some of the charts that @wroscoe has in some of his notebooks). 

Here's an example:
screen shot 2017-10-11 at 10 32 18 am

### DIFF
--- a/donkeycar/templates/donkey2.py
+++ b/donkeycar/templates/donkey2.py
@@ -8,6 +8,7 @@ Usage:
     manage.py (calibrate)
     manage.py (check) [--tub=<tub1,tub2,..tubn>] [--fix]
     manage.py (analyze) [--tub=<tub1,tub2,..tubn>] (--op=<histogram>) (--rec=<"user/angle">)
+    manage.py (plot_predictions) [--tub=<tub1,tub2,..tubn>] (--model=<model>)
 
 Options:
     -h --help     Show this screen.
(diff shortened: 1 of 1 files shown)


wroscoe:
This looks awesome but can we find a place to put it outside the manage.py script. This is the easiest place to put things but prevents it from being used elsewhere.

This could be its own independent function that had defaults to read in "user_angle" and "user_throttle". It would also be dependent on the model having only two outputs which is likely to change. This could be overcome by using the output names of the model? For keras you can use `model.output_names` to get a list of the outputs.

alanwells:
@wroscoe @tawnkramer what do you guys think about creating a new category of parts for visualization? It would provide a place to put @tawnkramer's `analyze` method from the template, my prediction plotting, and the visual backprop/GradCam work that @kennethjiang has been working on.

wroscoe:
My preference would be to put as much as possible into a utils or library file that could be called by an access function in the managment area. Long term we should make this part of the webapp again. That is a larger refactor though.

alanwells:
@wroscoe @tawnkramer Over the weekend I created a revised version of this function and the `analyze` function that doesn't break when records are deleted:

```
def anaylze(cfg, tub_names, op, record):
    '''
    look at the tub data and produce some analysis, for example:
    manage.py analyze --tub=<tub1,tub2,..tubn> --op=histogram --rec="user/angle"
    '''
    import json
    import glob

    tubs = gather_tubs(cfg, tub_names)

    if op == 'histogram':
        import matplotlib.pyplot as plt
        samples = []
        for tub in tubs:
            record_paths = glob.glob(os.path.join(tub.path, 'record_*.json'))
            
            for record_path in record_paths:
                with open(record_path, 'r') as fp:
                    json_data = json.load(fp)
                sample = json_data[record]
                samples.append(float(sample))

        plt.hist(samples, 50)
        plt.xlabel(record)
        plt.show()

def plot_predictions(cfg, tub_names, model_name):
    '''
    Plot model predictions for angle and throttle against data from tubs.

    '''
    import matplotlib.pyplot as plt
    import pandas as pd
    from PIL import Image
    import json
    import glob
    import numpy as np

    tubs = gather_tubs(cfg, tub_names)
    
    model_path = os.path.expanduser(model_name)
    model = dk.parts.KerasCategorical()
    model.load(model_path)

    user_angles = []
    user_throttles = []
    pilot_angles = []
    pilot_throttles = []

    for tub in tubs:
        record_paths = glob.glob(os.path.join(tub.path, 'record_*.json'))

        for record_path in record_paths:
            with open(record_path, 'r') as fp:
                json_data = json.load(fp)
            
            image_filename = json_data["cam/image_array"]
            image_path = os.path.join(tub.path, image_filename)
            img = Image.open(image_path)
            img = np.array(img)

            user_angle = float(json_data["user/angle"])
            user_throttle = float(json_data["user/throttle"])
            pilot_angle, pilot_throttle = model.run(img)

            user_angles.append(user_angle)
            user_throttles.append(user_throttle)
            pilot_angles.append(pilot_angle)
            pilot_throttles.append(pilot_throttle)

    angles_df = pd.DataFrame({'user_angle': user_angles, 'pilot_angle': pilot_angles})
    throttles_df = pd.DataFrame({'user_throttle': user_throttles, 'pilot_throttle': pilot_throttles})

    fig = plt.figure()

    title = "Model Predictions\nTubs: " + tub_names + "\nModel: " + model_name
    fig.suptitle(title)

    ax1 = fig.add_subplot(211)
    ax2 = fig.add_subplot(212)

    angles_df.plot(ax=ax1)
    throttles_df.plot(ax=ax2)

    ax1.legend(loc=4)
    ax2.legend(loc=4)

    plt.show()
```
Is it helpful to submit those as a PR against the manage.py template or would you guys prefer in to wait until these can be refactored into somewhere else in the codebase?

tawnkramer:
Thanks Alan,

I made similar changes to accommodate tub edits with non continuous records
and checked that in yesterday. I renamed analyze to histogram, since it
simplifies the verbiage to invoke, and flattens command hierarchy.

PR #165, opened 2017-10-12
Donkey sim

tawnkramer:
This adds sim server to integrate with Unity Donkey simulator. The simulator is a separate application that will generate data in the tub format. Then models can be trained, using real or virtual images. These models can be validated by having the simulator ask the donkey how to drive given sensory input via websocket connection to donkey code.

Documentation and test code added. Simulator application is served from my Google drive.


tawnkramer:
I had troubles getting the default categorical model to drive these tracks. I had much better luck with a modified Linear model with cropping and image normalization. I'd like to consider how to switch back and forth between the two. Perhaps with a --type=<categorical|linear> defaulting to categorical as we do now.

sagnibak:
Hi,

I was looking for the source files of the donkey car simulator. I don't think they have been released anywhere. Please point me to them if I am wrong. I want to add a few features to it.

Thanks,
Sagnik B.

tawnkramer:
https://github.com/tawnkramer/sdsandbox
Branch: donkey

PR #166, opened 2017-10-14
Prevent the generators from being cached

kennethjiang:
Currently the dataset returned from generators will be cached by `itertools.cycle` after 1st epoch. This becomes a problem when I want to use data augmentation to generate unlimited amount of data.

This PR fixes the issue stated above.


tawnkramer:
Sounds useful, but caching was also useful. Can we conditionally keep the
caching? Was a noticeable speed improvement.

kennethjiang:
@tawnkramer Good point. With cache training runs much faster. Will make a command argument for that.

kennethjiang:
Added a command line argument to cache training data.

PR #170, opened 2017-10-25
V 2.2.0 - better tubs, opencv, bug fixes

wroscoe:
* Use pandas for tub index. 
* Option to load preloaded model.

Minor changes:
* sort tubs in tub editor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,4 @@ lag_log.csv
 .cache
 site/*
 build
-
+.idea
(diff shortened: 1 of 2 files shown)


wroscoe:
@tawnkramer for these releases into master,  we should queue up the changes, increment the versions and then release, try to take out unneeded code, update docs. I'm going to create a new V2.2.0 that will have breaking changes.  :+1:

PR #171, opened 2017-10-26
Extra steps for Mac installation of simulator

michaelbanfield:
I had some issues generating logs when running the simulator locally. I checked

/Users/<username>/Library/Logs/Unity/Player.log

and found the following

`
UnauthorizedAccessException: Access to the path "/private/var/folders/.../d/donkey_sim.app/log/log_car_controls.txt" is denied.
`

This is caused by Sierra's sandboxing of untrusted applications, and will probably be an issue for all Mac users.

### DIFF
--- a/docs/guide/simulator.md
+++ b/docs/guide/simulator.md
@@ -12,6 +12,11 @@ You will need a specific build per platform:
 
 Extract this compressed file. It will create a folder with executable. Double click that executable to launch the simulator.
 
+
+#### Extra Mac Steps
+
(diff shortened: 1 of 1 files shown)


tawnkramer:
Interesting, but we don't actually need that file.
Were you getting any files written to the donkey_sim.app/log/ dir?
I'm not running on Sierra.

michaelbanfield:
Just to clarify, that folder is created by Sierra to sandbox the application. It basically only lets the application operate in a read only filesystem, so no logs can be written

https://lapcatsoftware.com/articles/app-translocation.html

PR #184, opened 2017-11-19
fix error in tar version

luismesas:
`donkeycar` python package version was declared as `2.2.1` but tar file was pointing to an older version.


wroscoe:
I'll fix this in the next release. The committed 404 errors.

PR #185, opened 2017-11-19
fix exception when not command line arguments

luismesas:
fix: an exception error when no command is passed any command line argument
fix: typo in the error description


wroscoe:
Any code change pull requests should be made against the dev branch. This is to allow enough time for manual testing since our test suite is not sufficient. The change looks good.

Issue #188, opened 2017-11-24
Configure joystick buttons via config


ebswift:
New users might find it intimidating changing the manager and controller code to suit a different joystick, it should be able to be setup in config.

tawnkramer:
And you fixed your own bug! Exemplary!

PR #189, opened 2017-11-24
Resolve Issue #188 - add ability to setup joystick buttons in config.py

ebswift:
Added the ability for the user to setup all the joystick buttons in config.py.


ebswift:
Sorry, setting back to dev

PR #190, opened 2017-11-24
Fix Issue 188: Added ability for users to configure joystick buttons in config.py

ebswift:
* Added joystick buttons to config_defaults.py
* Added joystick parameters to donkey2.py
* Added joystick parameters to controller.py


tawnkramer:
Thanks for this contribution. I feel like this is moving in the right direction, but needs to be taken further. There needs to be an arbitrary map of buttons to actions. And that map needs to be in user space. I will work on something like that soon.

ebswift:
That'd be great, thanks Tawn.

PR #191, opened 2017-11-27
added more documentation to s3uploader

pierreaws:


wroscoe:
This would be better implemented as a separate part. Leaving this as a fork.

PR #195, opened 2017-12-07
fixed issue with ImgCrop in cv.py

wheatgrinder:
updated cv.py to fix issue with ImgCrop class


wroscoe:
This pull request has many extra commits. I added this change directly to master.

PR #209, opened 2018-01-26
Add Adam's store.

adammconway:
Adding as an option, Adam's store where people can get the donkey chassis 3D Printed.


tawnkramer:
your first store url has an extra STL at the end of it

adammconway:
Tawn, I think I changed this correctly.  It is a little confusing doing the PRs from the docs.
AC

PR #243, opened 2018-04-19
Add instructions to expand filesystem for the Pi

dnswus:


wroscoe:
Thanks. Just pushed these to docs.donkeycar.com

PR #246, opened 2018-04-30
Replace Keras library with TensorFlow API

joekickass:
Recent versions of TensorFlow has adopted the Keras API. Replacing
the Keras library with the TensorFlow Keras implementation lessens
the risk of library mismatch and makes upgrading TensorFlow easier.

Also trim whitespace from changed files.


joekickass:
I've only run unit test on the patch, no validation on the car yet. Should probably test both training a new model and running it, as well as running an old model trained before applying the patch.

wroscoe:
Thanks for this! I'll look to merge into the dev branch.

wroscoe:
make your second pr against dev

- Will

joekickass:
will do

PR #259, opened 2018-05-17
Adding Robocar store as an official store

adammconway:


adammconway:
Will, do you mind approving and pushing this?

PR #266, opened 2018-06-14
Improved throttle

tawnkramer:
These changes have led to improve throttle prediction over the last few months. I feel they would benefit the community.

This changes the default model to use categorical throttle. It will be backwards compatible with older models, but will create new models in the new format going forward. 

I found I needed to add some small dropout to get improved predictions on both angle and throttle. It does converge in slightly more epochs as a result. ( about 2-4 on average ).

I started with the master branch when making these changes. Let me know if you'd prefer to start with the dev branch so fewer problems merging.

### DIFF
--- a/donkeycar/parts/camera.py
+++ b/donkeycar/parts/camera.py
@@ -115,11 +115,11 @@ class MockCamera(BaseCamera):
     '''
     Fake camera. Returns only a single static frame
     '''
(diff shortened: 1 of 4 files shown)


tawnkramer:
I guess this change is not wanted. No feedback, so I will remove pull request.

dlarue:
As if it hasn't been mentioned a half dozen times in the help channel on Slack or something. :/

PR #269, opened 2018-06-21
Deploy to pypi only once when using build matrix

joekickass:
Makes use of TravisCI build stages. Note that this feature is still
in beta.


wroscoe:
Thanks! This confused me for a while yesterday. I thought the error was incorrect versioning

joekickass:
Sure, no problem.

Yeah, Travis features and syntax is a bit confusing and hard to test before commiting.

Hope it works 😉

PR #277, opened 2018-07-10
Fix source code of parts directory, to follow PEP8 rules

gang65:
I have started building my own car. 
I would like to contribute to this great project.
As a start point, I fixed source code from parts directory to follow PEP8 specification:
https://www.python.org/dev/peps/pep-0008/

In my opinion it is improving consistency of the source code.


wroscoe:
thanks

PR #278, opened 2018-07-11
Add flop_tub script for data augmentation

r7vme:
Script to augment data by flopping image and steering angle.

### DIFF
--- a/docs/guide/train_autopilot.md
+++ b/docs/guide/train_autopilot.md
@@ -27,6 +27,18 @@ folder from the raspberry pi.
 rsync -r pi@<your_pi_ip_address>:~/mycar/data/  ~/mycar/data/
 ```
 
+## Data augmentation
+
+Data augmentation is an optional step, but can be beneficial to prevent [overfitting](https://en.wikipedia.org/wiki/Overfitting). One of the straightforward approaches is to horizontally flop the image and set steering angle to the opposite value.
+
+Script below requires `Pillow` python module. Easiest way to install it is `pip install Pillow`.
+
+To do that execute:
+```bash
+./donkeycar/templates/flop_tub.py <path to the tub folder>
+```
+
+This command will create new tub with suffix `_flopped`. It also will make sure to change the steering angle value to the opposite.
 
 ## Train a model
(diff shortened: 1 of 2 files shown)


r7vme:
I saw similar questions in Slack, so decided to put my script upsteam.

gang65:
Nice work.
If you would like to learn more about overfitting, you could check link:
https://www.youtube.com/watch?v=u73PU6Qwl1I

I think you could add link with theoretical introduction of overfitting problem to documentation. It will help newcomers understand the problem.

To avoid overfitting in DonkeyCar, the drop method:
https://keras.io/layers/core/#dropout
is used:
https://github.com/wroscoe/donkey/blob/dev/donkeycar/parts/keras.py#L92

docs/guide/train_autopilot.md
```diff
@@ -27,6 +27,16 @@ folder from the raspberry pi.
 rsync -r pi@<your_pi_ip_address>:~/mycar/data/  ~/mycar/data/
 ```
 
+## Data augmentation
+
+Data augmentation is an optional step, but can be beneficial to prevent overfitting. One of the straightforward approaches is to horizontally flop the image and set steering angle to the opposite value.
```
gang65: Please add some external link for overfitting problem, like: https://en.wikipedia.org/wiki/Overfitting 
or 
https://www.youtube.com/watch?v=u73PU6Qwl1I

r7vme:
@gang65 thanks for review, added link to wikipedia.

r7vme:
@gang65 yep, doing it with python is a way to go. Done, please review.

r7vme:
@gang65 added `pip install Pillow` into doc. thanks for review.

r7vme:
@wroscoe can you take a look and merge if fine?

wroscoe:
Hi @r7vme. Thanks for posting your work. Putting it as a pull request makes it searchable for everyone. I'm not going to merge this implementation because I don't want to add another dependencey and I'd like to do the augmentation using tensors instead of an image library. This will make training easier / faster for people. 

I think something using the technique described here: https://github.com/filmo/donkey/blob/data_gen_office/donkeycar/parts/datagenerator.py but refactored to be easier to access / configure.

r7vme:
Agree, having this built-in makes more sense. Thanks

PR #280, opened 2018-07-25
Fixed conda install path

michaeljsmalley:
The conda install path was pointing to `envs/mac.yml`, but nowhere before that instruction does it tell the reader to `cd` into the `install` subdirectory first, so for the command to work, `install` must be prepended to the path.


michaeljsmalley:
This same documentation patch may also be warranted for the Windows and Linux sections of the install guide, but I didn't have the time to test those platforms this evening.

PR #282, opened 2018-08-06
Increase camera ISO to 1600, to reduce blurring caused by fast move.

gang65:
With this fix, the NN is learning much more effective.
More information about exposure mode are available at:
http://picamera.readthedocs.io/en/release-1.10/api_camera.html#picamera.camera.PiCamera.exposure_mode


wroscoe:
Closing because camera code now lives in this repo.. https://github.com/autorope/donkeypart_picamera

PR #296, opened 2018-08-24
Update Mac instructions for installing tensorflow

distractedDave:
For tenserflow 1.8 that is in donkeycar 2.5.1 you need this new file.
if not you will get error when trying to train model  ModuleNotFoundError: No module named 'tensorflow.python.keras


kelvinkoko:
I hit the same issue and the above update fixed the issue, I think the guide should be updated.

PR #303, opened 2018-09-01
env path update on doc

Muosvr:


wroscoe:
fixed by other pull request

PR #306, opened 2018-09-02
Update install_software.md

icuic:
The default branch is "dev", we must switch to "master" branch manually. Other wise, there would be unexcept errors during the fellow process.


wroscoe:
If people are cloning rather than installing from pip they should be using the dev branch.

PR #309, opened 2018-09-12
added assertions for add parts method arguments

zrebie:
Added assertions for inputs/outputs/threaded. When those are set by mistake as strings, they will be incorrectly put to the dictionary in memory object.


donkeycar/templates/donkey2.py
```diff
@@ -41,7 +40,7 @@ def drive(cfg, model_path=None, use_joystick=False, use_chaos=False):
     V = dk.vehicle.Vehicle()
 
     clock = Timestamp()
-    V.add(clock, outputs='timestamp')
+    V.add(clock, outputs=['timestamp'])
```
gang65: Nice catch.
With this fix the timestamp is properly saved to json file.

Previously the `.json` file looks like (`timestamp` is `null`):
> {"user/angle": -0.6392101809747612, "timestamp": null, "cam/image_array": "18500_cam-image_array_.jpg", "user/throttle": 0.63, "user/mode": "user"}

donkeycar/vehicle.py
```diff
@@ -39,6 +42,9 @@ def add(self, part, inputs=[], outputs=[],
             run_condition: boolean
                 If a part should be run at all.
         """
+        assert type(inputs) is list, "inputs is not a list: %r" % inputs
+        assert type(outputs) is list, "outputs is not a list: %r" % outputs
+        assert type(threaded) is bool, "threaded is not a boolean: %r" % threaded
```
gang65: Nice improvement.
With these assertions, the is no possibility, to have `inputs` and `outputs` different than `list`

Issue #97, opened 2017-07-15
Make an image_undistort part.


wroscoe:
This is a request for the 2.1 version that uses parts. 

For many of the computer vision approaches an undistorted image is needed to determine the real angle of the line. Given that their are so many cars that use the same camera it would be helpful to have a part that would undistort the image for the default camera. 

Most logic can be copied from here: https://github.com/wroscoe/udacity_projects/blob/master/P4_advanced_lanelines/Code%20and%20Writeup.ipynb

PR #310, opened 2018-09-12
Added a CalibratedPiCamera Part

Labonneguigue:
I extended the `calibrate` base command to also
take pictures of a chessboard with the camera to
calibrate it and being able to undistort it afterwards.

Opened the PR for initial review. 

#97 

Still need to:

- [ ] Write "How to" in the readme
- [ ] Properly test it


wroscoe:
I like a lot this but it should be written as its own part like this one. https://github.com/autorope/donkeypart_bluetooth_game_controller  Also default donkeycar doesn't assume cv2 is isntalled.

PR #320, opened 2018-10-05
Change the Throttle back to right analog stick on PS3 controller

dnswus:
I used `jstest` to check the axes number for the vertical movement of the right analog stick on PS3 controller. Seems the button mapping has changed in Raspbian Stretch.

Reference: https://retropie.org.uk/forum/topic/17400/changed-button-numbers-on-ps3-controller-on-stretch-4-4-image


wroscoe:
Joystick has been moved out of this repo into donkeyparts

Issue #14, opened 2016-12-30
Improve data augmentation with shifted images


wroscoe:
Here is a great example: https://chatbotslife.com/using-augmentation-to-mimic-human-driving-496b569760a9#.b6jyrypqu

alanswx:
I don't know if this improves things, or if this augmentation is quite right, but you can do a lot of it with one call:

             X_train, X_validation, y_train, y_validation = train_test_split(X, Y, test_size=0.20, random_state=42)
             datagen = ImageDataGenerator(
                rotation_range=1,
                    width_shift_range=0.2,
                    height_shift_range=0.2,
                    shear_range=0.2,
                    zoom_range=0.02,
                    horizontal_flip=False)


             # compute quantities required for featurewise normalization
             # (std, mean, and principal components if ZCA whitening is applied)
             datagen.fit(X_train)

             # fits the model on batches with real-time data augmentation:
             m.fit_generator(datagen.flow(X_train, y_train, batch_size=32, shuffle=True),
                    samples_per_epoch=len(X_train)*3, nb_epoch=20,callbacks=callbacks_list,
                    validation_data=(X_validation, y_validation)
                    )

wroscoe:
Parts of that will work. We'll want to validate on non-augmented data.

alanswx:
That snippet does validate on non-augmented data. (I believe)

wroscoe:
You're right @alanswx but it doesn't switch the steering angles when you flip the image horizontally.

alanswx:
that is why i turned off the flipping code -- once i get my car driving i will check in a class that does all the augmentation

wroscoe:
I've moved the augmentation code to the donkeybarn repo. https://github.com/autorope/donkeybarn/blob/master/donkeybarn/cv/augment.py

PR #324, opened 2018-10-13
Horizontal flip and shift added for data augmentation

Labonneguigue:
In response to #14 

* Horizontal flipping of the image and changes the sign of the steering.

* Horizontal shifting of the image is uniformly distributed over a range of 50 pixels. The effect of the shift onto the steering value is controllable by the `pixel_change` argument.

⚠️ Shifting requires opencv

shifted


wroscoe:
I like this and think its needed but don't want to include cv2 in the requirements. Planning to include augmentation in this repo that will be what's used for training and not used on the car. https://github.com/autorope/donkeybarn/tree/master/donkeybarn

Thanks!

Issue #322, opened 2018-10-13
Sanitizing User Input when Calibrating: "ValueError: invalid literal for int() with base 10"


rampageservices:
Hello,

I use a ssh program that works in powershell installed via www.scoop.sh.
This does not happen in OSX terminal ssh.

I run into a ValueError whenever I press the backspace during calibration.
While my situation is a edge case, the bigger issue is that It appears there is no sanitation done on user input for this step of the process.

The code in question is the following:
`int(input('Enter a PWM setting to test(0-1500)'))`

My computer science professors used to call this sloppy ninja code. You are casting user input as an int without checking the value first. At the very least, the input should be checked to see if it is within bounds, 0-1500. 

In the best case scenario, only digits should be assigned to the pwm variable.

See this stack overflow question for a few basic examples:
https://stackoverflow.com/questions/17336943/removing-non-numeric-characters-from-a-string

Thanks!

`(env) pi@donkeypi:~/mycar $ donkey calibrate --channel 0
using donkey v2.5.1 ...
Enter a PWM setting to test(0-1500)401
Enter a PWM setting to test(0-1500)402
Enter a PWM setting to test(0-1500)403
Enter a PWM setting to test(0-1500)4034^H^H4
Traceback (most recent call last):
  File "/usr/local/bin/donkey", line 11, in <module>
    sys.exit(execute_from_command_line())
  File "/usr/local/lib/python3.5/dist-packages/donkeycar/management/base.py", line 445, in execute_from_command_line
    c.run(args[2:])
  File "/usr/local/lib/python3.5/dist-packages/donkeycar/management/base.py", line 156, in run
    pmw = int(input('Enter a PWM setting to test(0-1500)'))
ValueError: invalid literal for int() with base 10: '4034\x08\x084'`

scitronboy:
Also, In my opinion, people should be able to read the prompt and enter a number between 0-1500. If they don't, I think python's error message is good enough to fix the problem.

rampageservices:
Sure, I can invest the time to fix it and submit a pull request.
However, the point of creating an issue is to discuss the issue before any action is taken.
Maybe, someone has already written their own fix, maybe someone feels strongly that they want to submit the pull request. 

Further, you are entitled to your opinion but I don't agree with your sentiment.

If Donkey car wants to see new user adoption the essential function of calibration needs to be bulletproof. Otherwise, users will get frustrated and no longer work on the project having never even gotten to the fun part of actually seeing the car drive itself.

[...]

PR #325, opened 2018-10-16
Added Validation of user input for PWM calibration

Labonneguigue:
Validation that the user inputs an integer and that the integer is indeed in range.

#322


Labonneguigue:
I see your point but I disagree with the idea. I don't think we want this 
`a = 'lkdfhisoe78347834 (())&/&745  '`
 to be a valid input.
Regex are beautiful but I think we should use them when needed. When the input is mistakenly not directly castable to int, we should reject the input.

scitronboy:
I agree with you @Labonneguigue , I think regexs are overkill and are unnecessary for this. Also, a user might have intended to input something different than the regex interprets it as.

scitronboy:
For example, if a user wants to input `1239`, but the user accidentally inputs `123i` because `i` is just below `9` on a keyboard, then the regex would remove the `i` and act like the user inputed `123`, which would _not_ be what the user wanted.

rampageservices:
No , 123 WOULD be what the user wanted.
It would strip the back space characters which is the original problem that
I was having if you reference the original issue.

scitronboy:
No, not necessarily. The user would _not_ want `123` if the user had meant to input `1234`.

PR #334, opened 2018-10-28
changed button names for PS3

mayorquinmachines:
Default button references for PS3

Since the default controller.py code was written for PS3 Sixaxis controllers,
we changed the update() method's logic to use PS3 Sixaxis button names.


wroscoe:
I've moved this controller out of the base donkey build. It now lives here and I made an issue referencing this pr.

PR #335, opened 2018-10-30
Add SerialController part.

Schoonology:
This part allows the control of the Donkey Car via the RX/TX pins on the Raspberry Pi.

This part was requested from a couple members of the Slack, originally created for my [Autonomous Robot Challenge](https://www.hackster.io/contests/arm2018) entry.

### DIFF
--- a/donkeycar/parts/serial.py
+++ b/donkeycar/parts/serial.py
@@ -1,0 +1,37 @@
+import serial
+
+class SerialController():
+    def __init__(self):
+        self.conn = serial.Serial(
+                '/dev/ttyS0',
+                baudrate=9600
+        )
+        self.steering = 0.0
+        self.throttle = 0.0
+        self.mode = 'user'
+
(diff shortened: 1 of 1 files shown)


wroscoe:
This looks awesome. I'm moving non core parts out of this repo so I've made a reverence to this pull here... https://github.com/autorope/donkeyparts/issues/3

Schoonology:
Fantastic! Thanks for the note, too; I'll move any parts I build in the future to that repo.

Schoonology:
Actually, now I see that's just an Issue on the other repo. Would you like me to resubmit this as a PR to `autorope/donkeyparts`?

wroscoe:
Even better would be if you were to make a separate repo for the part like this (https://github.com/autorope/donkeypart_bluetooth_game_controller) that could be installed via pip. I could then fork it into autorope to help people find it. Thanks!

PR #338, opened 2018-11-04
remove non core parts

wroscoe:
This pull removes some of the non core parts to reduce the size of the repo. These parts that are still used will be added as separate repos/python packages that can be installed separately.


depperson:
These changes break the simulation documentation (causes issue #339).  It appears that simulation.py completely changed, indicating some new method is required.  Please update the docs both to include the new simulation details as well as to checkout a known working git tag so that it doesn't break in the future.

mtedder:
Why was this change made if it breaks the simulator?

PR #345, opened 2018-11-09
Remove --path from createcar docs

elp2:
The --path argument is not valid, it's a positional argument.

createcar: error: unrecognized arguments: --path


wroscoe:
I just merged a duplicate PR.

PR #359, opened 2018-12-09
Remove unneeded nut reference

jamestharpe:
Also "the nut" is never used later in the instructions and there is no nut in the Magnet kit.


tawnkramer:
this change is folded into donkey 3.0 branch, thanks!

PR #362, opened 2018-12-09
USB Battery does not fit on plate underside

jamestharpe:
Attaching the USB battery to the underside of the plate as suggested make it impossible to attach the plate to the car. Attaching it to the back of the roll cage worked for me.


adammconway:
The location of the battery is clarified in this video.  I will update the docs as well.

jamestharpe:
Which video?

PR #368, opened 2019-01-22
Update install_software.md

chris-han:
I'm following the doc to install on my Pi : https://github.com/autorope/donkeycar/blob/dev/docs/guide/install_software.md#get-the-raspberry-pi-working. There's section says "If you are using the prebuilt image specified above, then your Pi is ready to go. You should see a mycar and donkey directory." I think it should be placed after the "Create your car app" section because I don't see the 'mycar' directory comes with the image


tawnkramer:
thanks for your fixes!

PR #372, opened 2019-02-24
Wait 0.1s after initializing the PWM controller.

eike-welk:
The ESC "Tamiya TBLE-02" makes a little leap otherwise.
This leap does not happen at the first time after power on,
but on all subsequent starts of the Donkeycar driving program.


eike-welk:
This little change fixes problems that I had with my Tamiya TT-02. The ESC that was sold with it (TBLE-02), seems to need this little wait, when the PWM-Signal stops and starts again.

tawnkramer:
Thanks for your fix!

PR #376, opened 2019-03-16
Fixed camera streaming for firefox

serge-m:
Fixed camera streaming for Firefox (65.0.1)

**Trigger**

streaming from raspberry camera doesn't work in firefox

**changes**

Added `\n` to `boundarydonotcross` header for each frame. Now it works for both chrome and firefox

@wroscoe @scitronboy


serge-m:
Before fix:
Screenshot from 2019-03-16 15-19-23


After fix:
Screenshot from 2019-03-16 15-21-21

tawnkramer:
Thanks for this fix!

PR #379, opened 2019-03-26
BUGFIX fixed get_record_gen when shuffle==False

serge-m:
BUGFIX fixed get_record_gen when shuffle==False, added test


tawnkramer:
thanks for your fix!

PR #381, opened 2019-04-18
Feature/logitech gamepad 710

teemukinnunen:
Tested and it works. Kind of. Something wrong with the throttle, but constant speed works. And it is enough for us for now.


tawnkramer:
I'm rolling these changes into donkey 3. I will order a logitech gamepad to test code.

PR #386, opened 2019-06-06
Fix not defined error

tall-josh:
This is my first ever Pull requrest. Sorry if I'm not doing it right.

`record_dict` is referenced before it is defined. Unpacked the `iterrows()` tuple in the for loop to `_, row` and called `record_dict = self.read_record(row.to_dict())`

Cheers, Josh

### DIFF
--- a/donkeycar/parts/datastore.py
+++ b/donkeycar/parts/datastore.py
@@ -290,11 +290,11 @@ class Tub(object):
             df = self.get_df()
 
         while True:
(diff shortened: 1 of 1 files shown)


tawnkramer:
Thanks for that fix Josh. Can you tell me what platform and what version of Pandas you are using?

tawnkramer:
thanks for your fix!

PR #387, opened 2019-06-14
Removed version requirement for tornado

Ezward:
- removed the explicit version from tornado in setup.py, mac.yml and other env files.
- fixed web.py to remove the old asynchronous annotations and convert to async await
- now explicitly creates the event loop for the thread in update()
- merged in web_controller/templates folder from master


tawnkramer:
couldn't solve conflicts easily. opening a new pull request with just web controller edits.

PR #389, opened 2019-06-16
Tk/donkey3

adammconway:
Some minor edits I found while installing the nano.  Also added the video to the hardware docs.  Will finish the Nano hardware docs later


tawnkramer:
lets pull against tk/donkey3 for now. Or just directly push to it.

PR #393, opened 2019-06-20
Adding tensorflow.python prefix to "keras" imports where required

Schmill:
Fix "salient" operation by adding tensorflow.python prefix to "keras" imports.
Fix "ShowCnnActivations" by adding tensorflow.python prefix to "keras" imports.


Schmill:
Added a minor update to fix typo in help text of ShowPredictionPlots 'type' argument

tawnkramer:
Thanks for this fix!

PR #398, opened 2019-06-21
Update tubclean webinterface

Schmill:
+ If clip is playing then speed changes from dropdown take effect immediately.
+ Clicking an image in the clip timeline will jump to that frame.


tawnkramer:
Awesome! Really makes this usable. Can’t wait to try it.

Schmill:
I'll be interested to hear what you think, I tried it in a number of scenarios (single clip paused / playing, multiple clips paused / playing, selecting thumbs from the active clip, selecting thumbs from the non-active clip) so hopefully it will all "just work" but let me know if there are any issues.

PR #401, opened 2019-06-24
Add axis names and button names to Joystick.

tikurahul:
* This step helps discover mappings not already pre-existing and
provides a way to dicover and modify defaults.

* Adds Forza mode for the XboxOne controller.

### DIFF
--- a/docs/parts/controllers.md
+++ b/docs/parts/controllers.md
@@ -223,5 +223,25 @@ quit
 
 Now that your controller is trusted, it should automatically connect with your Raspberry Pi when they are both turned on.  If your controller fails to connect, run the bluetoothctl steps again to reconnect.
 
+## Discovering / Modifying Button and Axis Mappings for Game Controllers
 
+To discover and modify your default button mappings (for your controllers) you can use the `Joystick` class defined in `donkeycar.parts.controller`.
 
+After setting up `Donkey` and activating your `virtualenv` you can do the following.
+First launch a `python` shell session.
+
+```python
+from donkeycar.parts.controller import Joystick
+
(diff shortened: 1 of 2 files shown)


tawnkramer:
Hi Rahul. Thanks for the extra details in the documentation. And added defaults in Joystick class. But I don't want to add the extra command to tie the shoulder button to the throttle by default. The PS3 and PS4 controllers are prone to accidental pressing of the shoulder analog buttons when you set them down on a table. Can you remove that, please?

tikurahul:
The `right_trigger` addition is only applicable to the `XboxJoystickController` which also happens to be in the same file. PS3, PS4 defaults remain the same. For the XboxOne controller, it's actually a nicer way to control the car. 

https://github.com/autorope/donkeycar/blob/9e2bbd777e1464b34a81258934388d160d549260/donkeycar/parts/controller.py#L1049

adammconway:
We had lots of thrashing on this feature, lots of PRs etc.  The solution might be to make it automatic.  We just look at whatever value is bigger, the stick or the trigger and use that.  That way it can be either/or

tikurahul:
The triggers work the way they should. Rather than the `right_trigger` do the work for both acceleration and deceleration we assign `right_trigger` to acceleration and `left_trigger` to deceleration. We do this by mapping raw axis values from `-1.` to `1.` to their `magnitude`.  This magnitude then gets scaled appropriately by `set_throttle`. 

We reverse the `magnitude` for the `left_trigger`. 
Tests: Tested on my Jetson Nano.

tawnkramer:
Can you resolve w base so I can merge?

tikurahul:
I rebased by changes with the `master` branch.

PR #409, opened 2019-06-27
Detect loading of unrecognised model file types.

Schmill:
Add 'else' clause to loading model files to capture unknown file extension types (i.e. NOT .h5 or .json) to avoid running with an "empty" model if the filetype is not recognised.


tawnkramer:
Thanks!

PR #410, opened 2019-06-27
Fixes issue where images come out as BGR instead of RGB for Jeston Nano

adricl:


adricl:
This fix is required as open cv requires GStreamer to have videos in BGR so it has to be done in post.

tawnkramer:
This adds an import cv2 that is a problem to import for the entire camera module. Continuous integration logs will run clean when this fixed. Thanks.

PR #412, opened 2019-07-01
Update install_software.md

vivianpiketmedia:
Added link to broader description on SSD in order to boost the readability for beginners.


frenck:
Friendly notification. This user tries to put his links in all kinds of projects, which seems to be the only thing this GitHub user account does. I guess in order to drive traffic to his website.

PR #414, opened 2019-07-03
CSI camera: Config option to flip the cam frame orientation

jithu83:
If the Jetson Nano's camera is mounted in a rotated position, provide an option
to flip the camera frame.

CSICamera nano part is implemented using gstreamer. Passing
the flip parameter to the gstreamer pipeline will do this.

Signed-off-by: Jithu Joseph <jithu83@gmail.com>


tawnkramer:
Look good. Thanks!

PR #419, opened 2019-07-07
Support for TensorRTLinear model.

tikurahul:
* This PR adds support for Tensor RT.
* There are probably more minor improvements I can make to `tensorrt.py`, but I think this is a good starting point.
* `freeze_model.py` exports metadata about the model which helps us reconstruct the TensorRT model.
* Make sure `pycuda.autoinit` is never loaded during training, as it interferes with `tensorflow-gpu`.

Test: Tested this on my Jetson Nano.

----

* **Note** - This PR removes a `Cropping2D` layer from the linear model as its unsupported by the TensorRT runtime. If you are using Cropping2D use `opencv` to do the cropping as a pre-processing step.

----

#### Setup TensorRT on your Ubuntu Machine
Follow the instructions [here](https://docs.nvidia.com/deeplearning/sdk/tensorrt-install-guide/index.html#installing-tar). Make sure you use the `tar` file instructions unless you have previously installed CUDA using `.deb` files.

#### Convert Model to `uff` on your Ubuntu Machine

* Use the script in `donkeycar/scripts/freeze_model.py` to freeze your linear model. This will convert the linear model to a protobuf file, and use the `convert-to-uff` utility to convert it to a `uff` file.

```bash
python freeze_model.py --model=../../Training/Race/models/Linear.h5 --output=../../Training/Race/models/Linear.pb
convert-to-uff ../../Training/Race/models/Linear.pb
# Produces Linear.uff
```
* Copy (`scp`) the `Linear.uff` file to the Jetson Nano.

#### Jetson Nano

* Setup some environment variables so `nvcc` is on `$PATH`. Add the following lines to your `~/.bashrc` file.

```bash
export CUDA_HOME=/usr/local/cuda
export PATH=$CUDA_HOME/bin:$PATH
export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
```

```bash
source ~/.bashrc
nvcc --version
```

You should see something like:
```bash
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2018 NVIDIA Corporation
Built on ...
Cuda compilation tools, release 10.0, Vxxxxx
```
* Switch to your `virtualenv` and install `pycuda`.

```
workon donkeycar
pip install pycuda
```

* After this you will also need to setup `PYTHONPATH` such that your `dist-packages` are included as part of your `virtualenv`. Add this to your `.bashrc`. This needs to be done because the python bindings to `tensorrt` are available in `dist-packages` and this folder is usually not visible to your virtualenv. To make them visible we add it to `PYTHONPATH`.

```bash
export PYTHONPATH=/usr/lib/python3.6/dist-packages:$PYTHONPATH
```

#### Running the model on the Jetson Nano.

* In `config.py` make sure you pick `tensorrt_linear` as the model type. 

```
DEFAULT_MODEL_TYPE = `tensorrt_linear`
```

* Finally you can do

```bash
# After you scp your `uff` model to the Nano
python manage.py drive --model=./models/Linear.uff
```

### DIFF
--- a/donkeycar/parts/keras.py
+++ b/donkeycar/parts/keras.py
@@ -325,8 +325,8 @@ def default_n_linear(num_outputs, input_shape=(120, 160, 3), roi_crop=(0, 0)):
     
     img_in = Input(shape=input_shape, name='img_in')
     x = img_in
-    x = Cropping2D(cropping=(roi_crop, (0,0)))(x) #trim pixels off top and bottom
-    #x = Lambda(lambda x: x/127.5 - 1.)(x) # normalize and re-center
+    # Remove Cropping2D here as TensorRT does not seem to support a strided_slice operation.
+    # TODO: Add this to the pre-processing step.
     x = BatchNormalization()(x)
     x = Convolution2D(24, (5,5), strides=(2,2), activation='relu', name="conv2d_1")(x)
     x = Dropout(drop)(x)
--- a/donkeycar/parts/tensorrt.py
+++ b/donkeycar/parts/tensorrt.py
@@ -1,0 +1,91 @@
+from collections import namedtuple
+from donkeycar.parts.keras import KerasPilot
+import json
+import numpy as np
+import pycuda.driver as cuda
+import pycuda.autoinit
+from pathlib import Path
+import tensorflow as tf
+import tensorrt as trt
+
+HostDeviceMemory = namedtuple('HostDeviceMemory', 'host_memory device_memory')
+
+class TensorRTLinear(KerasPilot):
+    '''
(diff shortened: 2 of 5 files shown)


tawnkramer:
Awesome!! Very nice comments. Have you added these to the docs? Thanks for all this work!

tawnkramer:
Do you feel like it’s ready to be merged w dev?

tikurahul:
Yes. I tested this all weekend. I think given this is opt-in it’s safe to go into `dev`. That way we can iterate on feedback.

PR #421, opened 2019-07-09
crop and normalize outside of model

tawnkramer:
this creates a part to pre-process image before inference to apply crop, as well as normalize images. It also crops and normalizes during training.

### DIFF
--- a/donkeycar/config.py
+++ b/donkeycar/config.py
@@ -73,9 +73,10 @@ def load_config(config_path=None):
         
     
     #derivative settings
-    cfg.TARGET_H = cfg.IMAGE_H - cfg.ROI_CROP_TOP - cfg.ROI_CROP_BOTTOM
-    cfg.TARGET_W = cfg.IMAGE_W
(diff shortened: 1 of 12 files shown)


tawnkramer:
I trained a few models with and without crop in the simulator. looked good to me so far. would you like to take a look before merge?

tikurahul:
Will take a look.

tikurahul:
All Keras parts will have to move to derived properties for width and height for the models to work correctly. 
https://github.com/autorope/donkeycar/blob/dev/donkeycar/parts/tensorrt.py#L40 for e.g will be incorrect if cropping is involved.

tawnkramer:
yes, it's unfortunate, but I don't see a better way. Don't think we want to overload the current vars image_w, image_h with the crop folded in. So I made new derived TARGET_W, TARGET_H. I've fixed all references, including the one you point out there. Does this strike you as the right approach or did you see a better way forward?

PR #429, opened 2019-07-12
Fix PartProfiler showing wrong number for KerasLinear

DocGarbanzo:
Fix PartProfiler to take KeyboardInterrupts into account as well as initialisations of parts. It seems like on_part_finished() is not called when a KeyboardInterrupt occurs leaving the time value in the time difference which messes up the timing table max and avg values. Here, we simply chop off first and last entry form the timing arrays. We also add a median to the table to see if the distribution of timings is skewed.

### DIFF
--- a/donkeycar/vehicle.py
+++ b/donkeycar/vehicle.py
@@ -7,6 +7,7 @@ Created on Sun Jun 25 10:44:24 2017
 """
 
 import time
+from statistics import median
 from threading import Thread
 from .memory import Memory
 from prettytable import PrettyTable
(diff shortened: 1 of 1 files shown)


Schmill:
Just wanted to say a big thanks for this commit, once KL reported correct timings I was able to see where my delay was in my control loop and try to do something about it 👍

DocGarbanzo:
Yeah no worries- this thing was bugging me too. However, although I got it to work, I am not sure exactly why the current try/catch of the KeyboardInterrupt seems to be interrupting the code at a different place, it’s just clear the loop doesn’t complete when you hit ctrl-c so I simply removed the last entry.

PR #430, opened 2019-07-13
MakeMovie improvements:

DocGarbanzo:
* Use index to correctly go from first to last frame
* Use index from tub instead of try/catching on continuous frames
* Use picture frame size to determine mid of and length of user / pilot line plots
* Remove unused code/variables
* Some pep-8 formatting


tawnkramer:
in the future, please pull against dev.

DocGarbanzo:
Yes - will do. Thanks.

PR #432, opened 2019-07-15
Schmill pwm

Schmill:
Allow the PWM value for the PCA9685 to be specified, (defaults to the standard).
Includes code to ensure that the given PWM values are scaled appropriately regardless of pwm freq.
Also minor spelling corrections in other files


Schmill:
Do NOT merge, I think mycodebase got message up prior to this PR!

PR #433, opened 2019-07-15
Allow makemovie to be used to plot user input (only)

Schmill:
Allow makemovie to be used to plot user input (only) when no model is specified.
Useful to be able to see quality of user input, and to compare the input timing in relation to the physical car changes (for camera setups where you can see the front wheels)


Schmill:
Do NOT merge, I think mycodebase got message up prior to this PR!

PR #434, opened 2019-07-15
Allow PWM freq. of PCA9685 to be specified

Schmill:
Allow PWM freq. of PCA9685 to be specified (default value remains as before)
Also includes code to scale PWM values appropriately based on the freq. specified.


Schmill:
> Did you find a different frequency helped you?

Still under evaluation, _I think so_, but still testing. Due to the tracks we are training and testing on being relatively small, in order to get any decent speed up I need the steering to react quickly. I think increasing the frequency has helped, but I also believe that the servo that was replaced some time ago may be a little slow.

dlarue:
What voltage do you have the servo getting?  We run pretty fast on a 30' track with laps less than 7 seconds using stock freq set for the PCA9685 PWM.

PR #435, opened 2019-07-15
Update MakeMovie to allow plotting of user data only

Schmill:
Update MakeMovie to allow plotting of user data only (no model) - useful for determining quality of user input.
Also fix the confusing mix of model argument being None vs "None".


tawnkramer:
thanks!

PR #437, opened 2019-07-19
Enable throttle scaling while controlling manual throttle

ebswift:
Previously while driving, if the throttle scaling was changed, the throttle would need to be released, then applied again in order for the new scaling to be used.  This small modification allows scaling to be changed while driving so maximum speed can be slightly increased/decreased on the fly.

Removed some extra spaces in the code and made the layout of the spacing a bit more consistent.

### DIFF
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -35,7 +35,7 @@ class Joystick(object):
             self.num_buttons = 0
             print("no support for fnctl module. joystick not enabled.")
             return False
-            
+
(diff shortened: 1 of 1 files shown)


tawnkramer:
sorry, I'm not understanding. Why would the throttle need to be released and re-applied? Ahh, because we only issue an axis event when the throttle axis changes. It does not need to be released, just changed enough to issue a new axis change event. So if you have the throttle fully pegged and then are trying to find the top end you will have problems. 

got it, thanks!

PR #444, opened 2019-07-31
Remove deprecated message argument from pytest.raises().

DocGarbanzo:
Newer versions of pytest require replacing the `message` parameter in pytest.raises() with a pytest.fail() at the end of the `with` statement:

    def test_should_raise_assertion_on_non_list_inputs_for_add_part():
        vehicle = dk.Vehicle()
        inputs = 'any'
>       with pytest.raises(AssertionError, message="inputs is not a list: %r" % inputs):
E       pytest.PytestDeprecationWarning: The 'message' parameter is deprecated.
E       (did you mean to use `match='some regex'` to check the exception message?)
E       Please see:
E         https://docs.pytest.org/en/4.6-maintenance/deprecations.html#message-parameter-of-pytest-raises

### DIFF
--- a/donkeycar/tests/test_vehicle.py
+++ b/donkeycar/tests/test_vehicle.py
@@ -2,17 +2,20 @@ import pytest
 import donkeycar as dk
 from donkeycar.parts.transform import Lambda
 
+
 def _get_sample_lambda():
     def f():
         return 1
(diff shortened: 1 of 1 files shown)


tawnkramer:
Thanks!!

PR #450, opened 2019-08-24
Dev

DocGarbanzo:
## Fixing slowness of salient mask 
Fixing the slowness of the salient mask in `makemovie`. Improves performance by a factor > 100 (and even more when making longer videos with many frames). Seems that repeated calls to `compute_visualization_mask` create tensor flow graphs without disposing them. This seems to be a known issue. On my MacBook I get a frame rate of 45 fps w/ saliency vs 100 fps w/o saliency which seems reasonable.

### These are the changes:
* replacing the method `visualize_saliency` with the corresponding code from Keras-vis
* splitting that code into the loss and optimizer which are initialised only once in `init_salient`
* running the remainder in the function `compute_visualization_mask` which is runs on every frame

### Remark
Unfortunately the file was checked in w/ Ctrl+LF in git (I'm surprised this is possible), so the diff below is hard to read. Changes are essentially in (original) lines 186 - 193. I also changed the default `start` value for `make movie` to 0.

### DIFF
--- a/donkeycar/management/base.py
+++ b/donkeycar/management/base.py
@@ -226,7 +226,7 @@ class MakeMovieShell(BaseCommand):
         parser.add_argument('--model', default=None, help='the model to use to show control outputs')
         parser.add_argument('--type', default=None, help='the model type to load')
         parser.add_argument('--salient', action="store_true", help='should we overlay salient map showing activations')
-        parser.add_argument('--start', type=int, default=1, help='first frame to process')
+        parser.add_argument('--start', type=int, default=0, help='first frame to process')
         parser.add_argument('--end', type=int, default=-1, help='last frame to process')
         parser.add_argument('--scale', type=int, default=2, help='make image frame output larger by X mult')
         parsed_args = parser.parse_args(args)
--- a/donkeycar/management/makemovie.py
+++ b/donkeycar/management/makemovie.py
@@ -1,19 +1,20 @@
-import math
-
-import numpy as np
 import moviepy.editor as mpy
 from tensorflow.python.keras import activations
+from tensorflow.python.keras import backend as K
 
 try:
     from vis.visualization import visualize_saliency, overlay
     from vis.utils import utils
+    from vis.backprop_modifiers import get
+    from vis.losses import ActivationMaximization
+    from vis.optimizer import Optimizer
 except:
     raise Exception("Please install keras-vis: pip install git+https://github.com/autorope/keras-vis.git")
 
 import donkeycar as dk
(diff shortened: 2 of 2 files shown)


tawnkramer:
Thanks!

sctse999:
Do we need tensorflow 2 to run this? I am getting the following error (running on a colab notebook) on dev branch

Further debugging shows below:
--------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-16-412772757ebf> in <module>()
----> 1 from vis.visualization import visualize_saliency, overlay

5 frames
/usr/local/lib/python3.6/dist-packages/tensorflow/python/util/deprecation_wrapper.py in __getattr__(self, name)
    104     if name.startswith('_dw_'):
    105       raise AttributeError('Accessing local variables before they are created.')
--> 106     attr = getattr(self._dw_wrapped_module, name)
    107     if (self._dw_warning_count < _PER_MODULE_WARNING_LIMIT and
    108         name not in self._dw_deprecated_printed):

AttributeError: module 'tensorflow' has no attribute 'python'

DocGarbanzo:
Hi Jonathan,
I am running this with tensorflow 1.13.1. Does it work on your PC? I have tried google colab only once for training but was a little disappointed with the performance as it was more than 20 times slower than on my small MacBook.

sctse999:
Hmmm .... still getting the same error on colab with tensorflow 1.13.1

DocGarbanzo:
Does it work locally for you?

sctse999:
@DocGarbanzo  just checked. Local works perfect. It's kind of pity that it could not run on colab. I will look into it later

DocGarbanzo:
Ok, good to know. Are you also running python 3.6 locally, as I have 3.7. 

- Dirk

sctse999:
In ```tensorflow_backend.py```, I change line 11 to ```import keras``` and it works in colab. Should I make a pull request?

DocGarbanzo:
It doesn't look right to use `keras` only there and `tf.keras` everywhere else in donkey. I think we don't install `keras` at all. Can you find which code in `Keras-vis` is leading to this deprecation warning, and we could rather fix that?

sctse999:
@DocGarbanzo , you are right. ```import tensorflow.keras``` works as well. Just verified on colab.

DocGarbanzo:
Cool. Do you make a PR? I should check out your colab workbook.

sctse999:
I will make a PR on the donkey car keras-vis repo after I verify this works also ok on local.  I am working on something else now but I will come back to this very soon.

PR #452, opened 2019-08-30
Update get_driving.md

woehrer12:
I have test it and can´t connect to the donkeycar webserver. Then i test it with the hostname and its running.


tawnkramer:
Thanks!

PR #453, opened 2019-08-30
Update Ip-Adress change to Hostname

woehrer12:
When you put in the Browser the ip-Adress with :8887 its not running.
You must put the hostname.local:8887
Test it with Ubuntu 18 and Firefox


tawnkramer:
Thanks!

PR #458, opened 2019-09-24
Small performance improvement and jitter output in ms

DocGarbanzo:
Improve performance by replacing division /255 with multiplication and adding some pep-8 formatting. Also improving jitter output to be in ms.


tawnkramer:
Thanks!

PR #460, opened 2019-10-02
check TensorFlow version before overriding keras session

pperle:
The fix for TensorFlow 1.3.1 introduced in https://github.com/autorope/donkeycar/commit/d1448d8e1cf5148a51853fe282978664db0deacb does not work for TensorFlow 1.4 and higher. Rather than having to remove these lines when trying out a new TensorFlow version, these lines are only executed if TensorFlow 1.3.1 is being used.

### DIFF
--- a/donkeycar/parts/keras.py
+++ b/donkeycar/parts/keras.py
@@ -15,7 +15,7 @@ models to help direct the vehicles motion.
 import os
 import numpy as np
 
-from tensorflow import ConfigProto, Session
+import tensorflow as tf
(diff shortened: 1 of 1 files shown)


tikurahul:
Which version of TF are you using ? TF 2.x is stable but we have not officially announced support for it yet.

dlarue:
It's not in the ARM repos yet so not appropriate to even start talking about changing defaults because it will break the system.

tawnkramer:
Thanks!

PR #463, opened 2019-10-06
Added SHIFT function to controller to allow button combinations

teinturman:
The Shift feature allows to define combinations of buttons in the controller.py file ; this adds more possibilities to control parameters of the car. 

While the shift key is pressed , buttons have a different function. 
usage exemple can be to control the pan/tilt of the camera, increase / decrease loglevel during runtime, adjust camera settings, modify the desirable "location" during runtime etc..
A simple exemple of shift key is provided in the wiiu definitions.

For more readability, the button_up_trigger_map and the button_down_trigger_map have been merged. "_1" means button down and "_0" means button up.

In order to allow the controller part to communicate directly with other parts, the mem object is passed to the controller. this allows to update directly some variables that can be used by other parts.     

i tried to ensure that the new file be compatible with all other existing files without any change.

### DIFF
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -714,6 +714,11 @@ class JoystickController(object):
         # print("On Axis Move:")
         # print(self.axis_trigger_map)
 
+    def set_button_up_trigger(self, button, func):
+         set_button_trigger(self, button, func)
+
+    def set_button_down_trigger(self, button, func):
+         set_button_trigger(self, button, func)
 
     def set_button_trigger(self, button, func):
         '''
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -653,13 +653,16 @@ class JoystickController(object):
         self.estop_state = self.ES_IDLE
         self.chaos_monkey_steering = None
         self.dead_zone = 0.0
+
         self.button_down_trigger_map = {}
+        self.button_up_trigger_map = {}
         self.axis_trigger_map = {}
         self.init_trigger_maps()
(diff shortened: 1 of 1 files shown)


tawnkramer:
The shift is not a bad idea, but I don’t like the change to button up. Not an improvement.

teinturman:
Hi tawnkramer, 
thanks for reviewing my pullrequest so fast. 
The change of button up was more for readability of code. As it does not add new features , i am happy to rollback this particular change and do a pull request with the SHIFT and mem changes only.

teinturman:
I close this pull request, as i will raise a different one with only the changes in the shift and mem.

tawnkramer:
I would like you to create a new sub class for these shift and mem changes. Inherit from the base class JoystickController please. I don't think I need the shift. And I used the button up. Please don't remove those from the base class.

teinturman:
Hi tawnkramer, 
I think i understand the requirement: it must be possible to keep the executed code as it is and have the choice to use shift or not. 
I have not yet succeeded to make it work with a sub class, but working on it.  
I Was thinking the following: 
-could the names of the Buttons be harmonized for the standard controls ? ( 2axis, 4pads,4buttons,l1,r1,r2 and 3 home buttons)? This would solve the fact that in manage.py some controls are created using ‘circle ´ which is only a available in ps controllers .
Readability of buttons assignations  could be maintained using. Comment son the code.
We could Also use existing ps3 names for backward compatibility.

Such change would  also allow to load additional trigger mappings indepensantly of the type of controller. Some users could even define their own config and load it from the manage.py file.
Thanks for your feedback

PR #467, opened 2019-10-17
retry of pwm.set_pwm on first exception due to I2C timing errors

dlarue:
Using the Teensy PCA9685 emulation I ran across a few I/O errors when donkey tried to set a new pwm value. It was explained by the original author of the emulation that the rPi I2C library doesn't "doesn't allow slaves to clock extend" which seems to be something lots of I2C slave devices do. I found that a simple retry of the message was enough to cover the sporadic I/O errors I'd seen.


tawnkramer:
That's interesting that this solves the i2c errors you were seeing.

PR #469, opened 2019-10-20
Docs update

nvtkaszpir:
Update docs, work in progress. Docs taken from master into master.

Update formatting such as:
* missing clickable urls
* wrap in backticks special words (at least most of them)
* fix lists due to missing newlines
* wrap code sections into proper code blocks
* remove some improperly created headers

Rename files to lowercase extensions
Remove +x permissions set on files

Todo - look into legacy.md and extract data from it into another docs.


nvtkaszpir:
notice, travis fails because it is impossible to run tests successfully from master (it is fixed in dev branch, though)

PR #474, opened 2019-10-26
Make PWM actuators threaded parts

DocGarbanzo:
# Performance improvement of PWM actuators
## Motivation
- PWM actuators, i.e. `PWMThrottle` and `PWMSteering` each are taking up 2-3ms in the vehicle loop, that is 5-6ms in aggregate.
- If we want to run the car at 40 Hz than we only have 25ms per loop for all parts. Keras takes up 13-15ms, so optimising the other parts to a minimum will leave more cpu time for the auto pilot.
- In recording drive mode, `TubWriter` seems to be taking up most of the time, around 7ms.
## Implementation
- Change of `PWMSteering` and 'PWMThrottle` into threaded parts. Non-threaded functionality is maintained.
## Impact
- Actuator performance goes down to < 0.1 ms.
### Puzzle
- I have no idea why the README.md shows a change of the file extension from upper case to lower case. It seems to me that it is lower case already now which looks correct looking at the source...

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ community contributions.
 * [Build instructions and Software documentation](http://docs.donkeycar.com)
 * [Slack / Chat](https://donkey-slackin.herokuapp.com/)
 
-![donkeycar](./docs/assets/build_hardware/donkey2.PNG)
+![donkeycar](./docs/assets/build_hardware/donkey2.png)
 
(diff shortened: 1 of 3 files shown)


DocGarbanzo:
I see the PR target is master / will re-route to dev.

PR #475, opened 2019-10-26
Performance improvement of PWMThrottle and PWMSteering 

DocGarbanzo:
# Performance improvement of PWM actuators

## Motivation

- PWM actuators, i.e. PWMThrottle and PWMSteering each are taking up 2-3ms in the vehicle loop, that is 5-6ms in aggregate.
- If we want to run the car at 40 Hz than we only have 25ms per loop for all parts. Keras takes up 13-15ms, so optimising the other parts to a minimum will leave more cpu time for the auto pilot.
- In recording drive mode, TubWriter seems to be taking up most of the time, around 7ms.

## Implementation
- Change of PWMSteering and 'PWMThrottle` into threaded parts. Non-threaded functionality is maintained.
- Minor code changes (pep-8), mostly whitespace

## Impact
- Actuator performance goes down to < 0.1 ms.

### DIFF
--- a/donkeycar/parts/actuator.py
+++ b/donkeycar/parts/actuator.py
@@ -23,7 +23,7 @@ class PCA9685:
         # Initialise the PCA9685 using the default address (0x40).
         if busnum is not None:
             from Adafruit_GPIO import I2C
-            #replace the get_bus function with our own
+            # replace the get_bus function with our own
             def get_bus():
                 return busnum
             I2C.get_default_bus = get_bus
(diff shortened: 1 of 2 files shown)


DocGarbanzo:
I killed PR #474 as it was against master and replaced it w/ this one here.

tawnkramer:
How’s responsiveness when driving?

DocGarbanzo:
Very good. I think the I2C call might be blocking. So it probably helps to move it into a continuous loop into the background.

On 26 Oct 2019, at 21:36, Tawn Kramer <notifications@github.com> wrote:

How’s responsiveness when driving?

tawnkramer:
Thanks!

Issue #471, opened 2019-10-21
Please add hflip & vflip config for picam


x1001000:
That would be very helpful.
Thanks :)

nvtkaszpir:
It's not merged, so the ticket should be kept open ;)

PR #497, opened 2019-11-18
add configs for PiCamera arguments: vflip & hflip

x1001000:


x1001000:
> I believe it would be better to add prefix `CAMERA_` to the constants, so that it would become `CAMERA_HFLIP` and `CAMERA_VFLIP`.
> It would help to specify that this refers to the camera directly.
Yes, you are right!
> Also would be nice to add a line of comment of the purpose of those.
Agree!
> In addition, how about merging 3 pull requests into one?
Sure. But I am new to PR so I don't know what's next.

nvtkaszpir:
fixes #471

PR #502, opened 2019-11-26
removed self arg from map_frange and moved map_frange to utils

Ezward:
map_frange() was defined in actuator.py and called in actuator.py by RPi_GPIO_Servo class.  It was defined with a 'self' argument, as if it was a class method, but it was defined in the module scope.  It's a useful function, so I moved it to utils.py and changed actuator.py to import it.

### DIFF
--- a/donkeycar/parts/actuator.py
+++ b/donkeycar/parts/actuator.py
@@ -7,6 +7,7 @@ are wrapped in a mixer class before being used in the drive loop.
 import time
 
 import donkeycar as dk
+from utils import map_frange
 
         
 class PCA9685:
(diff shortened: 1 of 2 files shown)


tawnkramer:
I don't know about the lint errors, looks like one is not configured correctly. But the test failed because import statement needs fixing:

```
______________________ ERROR collecting test_actuator.py _______________________
ImportError while importing test module '/home/travis/build/autorope/donkeycar/donkeycar/tests/test_actuator.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
donkeycar/tests/test_actuator.py:3: in <module>
    from donkeycar.parts.actuator import PCA9685, PWMSteering, PWMThrottle
donkeycar/parts/actuator.py:10: in <module>
    from utils import map_frange
E   ImportError: No module named 'utils'
```

PR #512, opened 2019-12-18
Thread-safe OLED display part with larger hardware support

deltaflyer:
Dear contributors,

we kindly ask for review of the extension of the OLED part. Following features have been added to the existing OLED-part

* **thread-safety:** upper bound in run time behavior to allow smooth actuator movements, while rendering information on the display
* **more display types**: support for 128x32 and 128x64 displays
* **reduced depencies:** support of Adafruit legacy SSD1306 pythen package removed
* **support for monochrome graphic rendering:** of base64-encoded png pictures
* **smooth integration into i2c-based peripherals:** (e.g. actuators) using fair share bus usage policies
* **compact footprint:** all graphics and resources embedded into the source code as a single file
* graphics are hand drawn and thus also open source

It has been tested with the latest dev branch using a Raspberry PI 3 and an Adafruit 16-Channel 12-bit PWM/Servo HAT. Attached are some pictures. After merging of this PR, I can also contribute documentation for the hardware integration. However, integrating the OLED display as another i2c device is complicated

NOTE: the display update rate is slow by purpose to give actuators on i2c priority. if actuators are controlled differently, the update rate can be incremented in the config.

Kudos to Rahul Ravikumar for his initial implementation and the input into the discussion

Thank you for comments and the review,
Oliver

### DIFF
--- a/donkeycar/parts/oled.py
+++ b/donkeycar/parts/oled.py
@@ -1,145 +1,378 @@
-import Adafruit_SSD1306
-
-from PIL import Image
-from PIL import ImageDraw
-from PIL import ImageFont
+import base64
+from io import BytesIO
+import math
+from PIL import Image, ImageFont, ImageDraw
+from queue import Queue
 import subprocess
-import time
-
-class OLEDDisplay(object):
-    '''
-    Manages drawing of text on the OLED display.
-    '''
-    def __init__(self, bus_number=1):
-        # Placeholder
-        self._EMPTY = ''
-        # Total number of lines of text
-        self._SLOT_COUNT = 4
-        self.bus_number = bus_number
-        self.slots = [self._EMPTY] * self._SLOT_COUNT
-        self.display = None
-
-    def init_display(self):
-        '''
-        Initializes the OLED display.
-        '''
-        if self.display is None:
-            # Use gpio = 1 to prevent platform auto-detection.
-            self.display = Adafruit_SSD1306.SSD1306_128_32(rst=None, i2c_bus=self.bus_number, gpio=1)
-            # Initialize Library
-            self.display.begin()
-            # Clear Display
-            self.display.clear()
-            self.display.display()
-            # Display Metrics
-            self.width = self.display.width
-            self.height = self.display.height
-            # Create Image in 1-bit mode
-            self.image = Image.new('1', (self.width, self.height))
-            # Create a Drawing object to draw into the image
-            self.draw = ImageDraw.Draw(self.image)
-            # Load Fonts
-            self.font = ImageFont.load_default()
-            self.clear_display()
-
-    def clear_display(self):
-        if self.draw is not None:
-            self.draw.rectangle((0, 0, self.width, self.height), outline=0, fill=0)
-
-    def update_slot(self, index, text):
-        if index < len(self.slots):
-            self.slots[index] = text
-
-    def clear_slot(self, index):
-        if index < len(self.slots):
-            self.slots[index] = self._EMPTY
-
-    def update(self):
-        '''Display text'''
-        x = 0
-        top = -2
-        self.clear_display()
-        for i in range(self._SLOT_COUNT):
-            text = self.slots[i]
-            if len(text) > 0:
-                self.draw.text((x, top), text, font=self.font, fill=255)
-                top += 8
-
-        # Update
-        self.display.image(self.image)
-        self.display.display()
 
 
 class OLEDPart(object):
     '''
-    The part that updates status on the oled display.
+    Display part for monochrome i2c OLED displays
+    using SSD1306 display controllers with a dimension of 128x64 or 128x32 pixel.
+    Draws stats information on the OLED display.
     '''
-    def __init__(self, bus_number, auto_record_on_throttle=False):
-        self.bus_number = bus_number
-        self.oled = OLEDDisplay(self.bus_number)
-        self.oled.init_display()
-        self.on = False
+
+    def __init__(self, address=0x3C, i2c=None, process_limit=32, auto_record_on_throttle=False, display_width=128, display_height=64, name_ethernet_interface="eth0", name_wlan_interface="wlan0"):
+        # Display subystem
+        self._EMPTY = ''
+        self._SLOT_COUNT = 4
+        self.slots = [self._EMPTY] * self._SLOT_COUNT
+        self.display = SSD1306Driver(address, i2c, process_limit, display_width, display_height)
+        self.canvas = self.display.get_canvas()
+        self.font = ImageFont.load_default()
+
+        # States
+        self.last_user_mode = None
+        self.last_recoding_state = None
+        self.last_num_records = None
+        self.is_donkeycar_starting_up = True
+        self.update_requested = True
+
+        # Infomation to display
         if auto_record_on_throttle:
             self.recording = 'AUTO'
         else:
-            self.recording = 'NO'
+            self.recording = 'NO (Records = 0)'
         self.num_records = 0
         self.user_mode = None
-        eth0 = OLEDPart.get_ip_address('eth0')
-        wlan0 = OLEDPart.get_ip_address('wlan0')
-        if eth0 is not None:
-            self.eth0 = 'eth0 : %s' % (eth0)
+        eth_interface = OLEDPart.get_ip_address(name_ethernet_interface)
+        wlan_interface = OLEDPart.get_ip_address(name_wlan_interface)
+        if eth_interface is not None:
+            self.eth_interface = 'eth0 : %s' % (eth_interface)
         else:
-            self.eth0 = None
-        if wlan0 is not None:
-            self.wlan0 = 'wlan0 : %s' % (wlan0)
+            self.eth_interface = None
+        if wlan_interface is not None:
+            self.wlan_interface = 'wlan0 : %s' % (wlan_interface)
         else:
-            self.wlan0 = None
+            self.wlan_interface = None
 
-    def run(self):
-        if not self.on:
-            self.on = True
+    def update_slot(self, index, text):
+        '''
+        Set text to a slot
+        '''
+        if index < len(self.slots):
+            self.slots[index] = text
 
-    def run_threaded(self, recording, num_records, user_mode):
-        if num_records is not None and num_records > 0:
-            self.num_records = num_records
+    def clear_slot(self, index):
+        '''
+        Clears slot content
+        '''
+        if index < len(self.slots):
+            self.slots[index] = self._EMPTY
 
-        if recording:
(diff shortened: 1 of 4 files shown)


tikurahul:
I already wrote an `OLED` part some time ago, and have been using it for a while. 

https://github.com/tikurahul/donkey/blob/donkey-v3-dev/donkeycar/parts/oled.py

For reference.

deltaflyer:
Hello @tikurahul 

glad, to get feedback from you. I studied your design. The functionality of your OLED part is very nice and showing the IP is a helpful feature. Since the Adafruit libraries use the Linux-Kernel function  i2c_smbus_xfer(..) that uses mutex-based protection for concurrent i2c access, your multi-threaded design works without locking up the I2C bus. The idea of my implementation is that it provides an upper run time bound for display updates to meet the soft real-time requirements of Donkey car even for bulky display updates. As a result, complex graphic rendering will not slow down I2c-based actuator controllers. What do you think is a good way to progress? Should I integrate your functionality into your existing OLED driver? Or should the upper bound runtime behaviour added to your OLED-driver implemenation?

Thank you for sharing your thoughts,
Oliver

tikurahul:
Thanks for looking at the part. Yes, in general I was trying to minimize the damage caused by a slow part. 

My version does not try to maintain any real-time guarantees but having used it for a couple of months, that has never been an issue. 

I can contribute my part to the main `donkeycar` repo. You can maybe subsequently make changes you want to, to improve upon the part ?

deltaflyer:
Hello @tikurahul 

> I can contribute my part to the main donkeycar repo. You can maybe subsequently make changes you want to, to improve upon the part ? 

I fully agree. Please do it like this. After that I will propose an extension to your OLED part with my real-time preserving i2c backend and the graphics functions. By this, we should powerful and safe OLED support with a straight-forward design.

tawnkramer:
Thanks guys! Excited to try this.

tikurahul:
Sorry, will send out the PR today.

deltaflyer:
Hello @tikurahul thank you for creating the PR. I am looking forward to code on the OLED part on the Christmas break. Best, Oliver

deltaflyer:
@tikurahul thank you for merging your PR, I will continue to work on it now.

deltaflyer:
Hello @tikurahul 
ok, I have the first prototype running on my hardware that uses your front end and my back end code. While doing this, I recognized that our OLED parts are assuming different display dimensions (128x64 vs. 128x32) :desktop_computer: . Now, I am extending the code base to support both resolutions. For owners of the larger 64 pixel height display, I plan to display an icon of the user mode in the lower part. To get proper testing, I ordered a second smaller display, which should arrive by tomorrow evening. It would be very nice to have all major display sizes in donkey car. I'll keep you in the loop. Best, Oliver

tikurahul:
Ah. I just realized that too. Let me know if you need help when testing things.

deltaflyer:
Hello @tikurahul I just pushed my changes and updated the PR description. Can you test the code with your display, if you don't mind? My 128x32 display also arrived, so I will test the code also with both displays.

tikurahul:
Hey @deltaflyer. Thank you for your PR. You have re-ordered changes, and I can no longer rebase as a result.

Can you please squash and reorder your commits so I can rebase them ? (Use `git pull --rebase` from `dev` and apply your changes on top. Ideally there should be 1 commit on top, with all your changes to make it easier to review).

deltaflyer:
Hello @tikurahul I just rebased the commits. Now everything should be fine for review.

donkeycar/parts/oled.py
```diff
@@ -1,145 +1,430 @@
...
+        if x != None:
+            return int(math.ceil(x / 400.0)) * 400
+        return 0
 
     def shutdown(self):
-        self.oled.clear_display()
-        self.on = False
+        self.display.shutdown()
+
+    def get_user_image_as_pil(self):
```
tikurahul: This should be static I would think, given that they are essentially constants.

donkeycar/parts/oled.py
```diff
@@ -1,145 +1,430 @@
...
+        self.cls()
+        if display_height > 32:
+            self._show_donkey_logo()
+        self.process_queues_at_once()
+        print("OLED display initialized.")
+
+    def get_canvas(self):
+        return self.canvas
+
+    def process_queues(self):
```
tikurahul: Out of curiosity, are there any actionable differences in performance when multiple i2c devices are involved ? Usually actuators are on a different bus, and every bus is modeled as a file handle in the kernel. 

Do you have any real benchmarks that measure the difference ? I have been using the `OLEDPart` and I have not detected any issues thus far.

deltaflyer: Hello @tikurahul,
thank you for your question - they made me think a lot about the design. First, I fully agree to you: the kernel models each i2c access as a own device. Thus, OLED drivers are totally fine in running as a own thread, when other thread use i2c at the same time. Hence, I moved the OLED part into an own thread to speed up the main drive loop with the latest commit.

Following your suggestions, I compared the performance of the existing part and the new OLED part on my Raspberry-Pi 3+ using the 128x32 mode from start up to idle with no loaded h5-model.

Original OLED as baseline
```
+--------------------+--------+-------+--------+--------+--------+--------+--------+
|        part        |  max   |  min  |  avg   |  50%   |  90%   |  99%   | 99.9%  |
+--------------------+--------+-------+--------+--------+--------+--------+--------+
|      OLEDPart      | 164.69 | 88.45 | 145.21 | 152.76 | 156.82 | 162.16 | 164.44 |
+--------------------+--------+-------+--------+--------+--------+--------+--------+
```

Implementation in this PR:
```
+--------------------+-------+------+-------+------+-------+-------+-------+
|        part        |  max  | min  |  avg  | 50%  |  90%  |  99%  | 99.9% |
+--------------------+-------+------+-------+------+-------+-------+-------+
|      OLEDPart      | 64.53 | 0.09 | 16.17 | 0.10 | 51.88 | 53.69 | 63.02 |
+--------------------+-------+------+-------+------+-------+-------+-------+
```
On average this the implementation is ~9 times faster. but also the 99% percentile looks promising. This is no suprise, since the low-level stuff is done directly in the part.

Interesting is what happens when the i2c bus gets under heavy load, e.g. when rendering graphics in 128x64 mode in this OLED implementation. If the queue size is set unlimited, the actuators react slow, since the Raspberry-PI hardware cannot dispatch all i2c calls fast enough due to limited system resources. If the process limit size is set to 32 i2c commands per thread call, actuators react smoothly while the display graphic is rendered.

Conclusion:
there is no issue with the existing OLED part. It is doing it's job fast enough and running it in an own thread is totally fine. Thus, I rename the PR. If graphic rendering should be supported on weaker platforms (e.g. Raspi 3+), i2c command queuing and optimized display primitives are needed to have smooth actuator movement. I leave it now open to you and the the other maintainers, if this part is worth the integration. It is quite sophisticated and may not follow the intension of Donkey car that thrives for an simple and straight forward design. Thank your for your time and effort you were spending on this. :-)

donkeycar/parts/oled.py
```diff
@@ -98,11 +98,14 @@ def run(self, recording, num_records, user_mode):
             self.display.cls()
             self.update_slots()
             self.display_text()
-            self.display_user_mode_icon(user_mode)
+            self.display_user_mode_icon()
             self.display.display()
             self.update_requested = False
         self.display.process_queues()
 
+    def update(self):
```
tikurahul: I think you can remove this method ?

donkeycar/parts/oled.py
```diff
@@ -151,10 +154,10 @@ def update_slots(self):
 
     def roundup(self, x):
         '''
-        Rounds a given number up to the next 400
+        Rounds a given number up to the next 100
         '''
         if x != None:
-            return int(math.ceil(x / 400.0)) * 400
+            return int(math.ceil(x / 100.0)) * 100
```
tikurahul: You could just write this as:

```
x // 100 * 100
```

donkeycar/templates/complete.py
```diff
@@ -534,7 +534,7 @@ def run(self, mode, recording):
             cfg.SSD1306_ETH_INTERFACE_NAME,
             cfg.SSD1306_WLAN_INTERFACE_NAME
         )
-        V.add(oled_part, inputs=['recording', 'tub/num_records', 'user/mode'])
+        V.add(oled_part, inputs=['recording', 'tub/num_records', 'user/mode'], threaded=True)
```
tikurahul: I am confused. The original `OLEDPart` was actually threaded, so did we go from a threaded part -> not threaded -> and not back to threaded again ?

donkeycar/parts/oled.py
```diff
@@ -81,7 +81,7 @@ def display_text(self):
                 self.canvas.text((x, top), text, font=self.font, fill=255)
                 top += 8
 
-    def run(self, recording, num_records, user_mode):
+    def run_threaded(self, recording, num_records, user_mode):
```
tikurahul: Look at the old `OLEDPart`. The threaded part needs to run in its own thread, and therefore should define a `run` method.  Then `run_threaded` can update the internal state of the car, and dispatch a call to `update`. I think you should also make `roundUp` optional.

donkeycar/parts/oled.py
deltaflyer: From my understanding of threaded parts, I cannot remove the method. Since the OLED part runs threaded with the latest update, the `update()` method is called from the donkey frame work. Thus from my understanding the part does not provide any output data, the body is empty. But maybe I lack understanding here. Maybe the parts [documentation](https://github.com/autorope/donkeycar/blob/dev/docs/parts/about.md) needs some updates at this point with more clarification on threaded parts

donkeycar/parts/oled.py
deltaflyer: The good old "//", I rework that. :-)

donkeycar/templates/complete.py
deltaflyer: Yes, indeed. The design went from `threaded -> not thread -> thread` over the iterations. I benchmarked every iteration to get the best solution on this:

* original OLED part: threaded -> result: high display update rate, decent i2c actuarors, no graphics, noticable CPU usage on Raspberry Pi 3+
* first iteration: reworked not threaded OLED part -> result: poor display update rate, very responsive i2c actuators, graphics support, moderate CPU usage
* second iteration: reworked threaded OLED part -> result: decent update rate, very responsive i2c actuators, graphics support, moderate CPU usage

donkeycar/parts/oled.py
deltaflyer: Ah, I see the problem here with my understanding. I will fix this in the next iteration until the end of the week

PR #513, opened 2019-12-20
Robo HAT MM1 (Official Support) - Peter v3.1.1

wallarug:
Work to bring an official Robo HAT MM1 implementation to the Donkey Car Community, within the current structure.

Making this PR to see what break and work from there.


wallarug:
**Missing in this commit**

- Documentation Updates:  To be completed separately and then merged in later.

tawnkramer:
Looks good guys. Just a few things:

1) please remove "__Version:__ 3.1.1rm (Robotics Masters)" from README.md
2) looks like rcinput.py is not used. Please remove the file.
3) no need to put cfg_complete.py in /quarantine. Please remove the file.
4) in templates/cfg_complete.py HAVE_ROBOHAT should default to FALSE
5) In templates/complete.py I'm not sure why you limit use of the RoboHATDriver to only when you pass a model path? We can work with that if it was working for you guys.
6) In setup.py, please remove the rm from 3.1.1 and fix the url and author email. We don't put all contributors on email list, just maintainers.

thanks guys

wallarug:
> Looks good guys. Just a few things:
> 
> 1. please remove "**Version:** 3.1.1rm (Robotics Masters)" from README.md
> 2. looks like rcinput.py is not used. Please remove the file.
> 3. no need to put cfg_complete.py in /quarantine. Please remove the file.
> 4. in templates/cfg_complete.py HAVE_ROBOHAT should default to FALSE
> 5. In templates/complete.py I'm not sure why you limit use of the RoboHATDriver to only when you pass a model path? We can work with that if it was working for you guys.
> 6. In setup.py, please remove the rm from 3.1.1 and fix the url and author email. We don't put all contributors on email list, just maintainers.
> 
> thanks guys

Hi @tawnkramer ,

No issues with any of the above.  I can get that sorted out now.

In relation to point 5:  because the Robo HAT MM1 in CircuitPython is controlling the motors directly from analog in - there is no need to send any serial data back to the Robo HAT MM1 to drive the motors - it is already driving them.   So technically there is no 'driver' when you are using an RC controller because it is doing it by itself.

So when you go to drive with a model, you need to send the values to the Robo HAT MM1 so it drives around according to the NN.

Hope this makes some sense.

wallarug:
GDay again @tawnkramer,

So I think that fixes up most of the issues outlined above.  Please let me know if there is anything further.

My apologies for not cleaning this all up.  Those defaults were for events we run with that repository.  We ran an event on the weekend.  It just makes it easier for people.

tawnkramer:
No worries. Looks great now! I will merge it. Can’t wait to give it a go.

PR #515, opened 2019-12-22
Add support for a PiOLED display to show useful information.

tikurahul:
* Connect the PiOLED display over I2C, and configure the bus number.
* Change `myconfig.py` to update the bus number, and the display automatically shows useful information.
* Uses the `Adafruit_SSD1306` library.

### DIFF
--- a/donkeycar/parts/oled.py
+++ b/donkeycar/parts/oled.py
@@ -1,0 +1,145 @@
+import Adafruit_SSD1306
+
+from PIL import Image
+from PIL import ImageDraw
+from PIL import ImageFont
+import subprocess
+import time
+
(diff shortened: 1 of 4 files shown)


tikurahul:
This should unblock https://github.com/autorope/donkeycar/pull/512.

donkeycar/parts/oled.py
```diff
@@ -0,0 +1,145 @@
+import Adafruit_SSD1306
```
Ezward: Do we need to update the virtual environment to include this dependency?

tikurahul: Good catch. I was porting over changes to a different branch. Brought this change over as well.

PR #516, opened 2019-12-22
Add FPV web server

DocGarbanzo:
# Adding a web server to only display the camera view for FPV driving
The LocalWebController displays the camera view but the frame is quite small and doesn't allow for a more immersive driving experience. Also it is geared towards controlling the car from the web client. When driving from a BT joystick or the RC controller, the LocalWebController is not needed. I'm adding a small web server that only displays the camera view. There is occasionally some lagging of the data stream, but the car is quite drivable in FPV view. 
* Adding class WebFpv that sends the camera stream on port 8890 - this is very much a simplified copy of the LocalWebController without the overhead of all the js, css files
* Reducing the async loop waiting time from 0.1s to 0.01s in the VideoAPI request handler to reduce lagging. The LocalWebController also benefits from that improvement
* Adding html file & donkey logo for the new FPV webpage

## Usage
* Add `USE_FPV` in your `myconfig.cfg` if you use the standard car app, or
* Add the part `WebFpv` in your own car app

## Attention: Raspbian Stretch to Buster
When running the WebFpv server and the TubWriter on the RPi 3B+ simultaneously I saw a strong degradation of file writing performance making the TubWriter slow and therefore the WebFpv sluggish. I strongly encourage to upgrade the RPi to Raspbian Buster. Not sure if this is due to Python 3.7 or the OS. At least this completely removes the I/O issues. I can now run the car driving with the AI, recording to the tub and viewing the thing through the WebFpv server at a vehicle frequency of 30Hz w/o exceeding the target loop time (implied 33.3ms).


tawnkramer:
Exciting improvement. Thanks for this! Good performance note as well.

tikurahul:
Just reviewed this change as well. Looks good !

PR #518, opened 2019-12-27
Add 'xboxswapped' joystick device

nvtkaszpir:
This effectively just swaps original XboxOne controller mini-joysticks - orofinally the fleft is for seering sideways and riht for acceleration.

This change allows to define new joystick mapping (which is just swapped the above one), so that throttle is on left joystick, and side steering is on right stick.

This conforms with joystick steering via web interface (when using game controller with web browser such as Chrome, and also follows steering scheme in other games such as Dirt 2 if controller is directly connected to the car, for example via bluetooth).

Other buttons are left intact (especially triggers which follow Forza style).

### DIFF
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -1135,6 +1135,71 @@ class XboxOneJoystickController(JoystickController):
         }
 
 
+class XboxOneSwappedJoystickController(JoystickController):
+    '''
+    A Controller object that maps inputs to actions
+    credit:
+    https://github.com/Ezward/donkeypart_ps3_controller/blob/master/donkeypart_ps3_controller/part.py
+    left joystick - acceleration
+    right joystick - turning
+    '''
+    def __init__(self, *args, **kwargs):
+        super(XboxOneSwappedJoystickController, self).__init__(*args, **kwargs)
+
+
+    def init_js(self):
+        '''
+        attempt to init joystick
+        '''
+        try:
+            self.js = XboxOneJoystick(self.dev_fn)
+            self.js.init()
+        except FileNotFoundError:
+            print(self.dev_fn, "not found.")
+            self.js = None
+        return self.js is not None
+
+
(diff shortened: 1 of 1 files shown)


tikurahul:
Can you consider making this a part of the joystick configuration and make it a part of the original class ?

Alternatively, your new `JoystickController` can extend the current `XboxOneJoyStickController` with the overriden maps.

nvtkaszpir:
You mean to to move 'swap' option to the base class?

tikurahul:
Yes. Make the swapping an option in `cfg_complete.py` and use that to make the swap.

nvtkaszpir:
Well, looking at the code it's not that easy as it sounds.
Each class inherits from JoystickController but each device uses different axes names   - compare Xbox and WiiU - method init_trigger_maps has different axis_trigger_map per device:
Xbox has 'left_stick_horz' while WiiU there is 'LEFT_STICK_X'.


So there are two options:
1. for each device controller add additional 'if' statement for swapped device remap (and then define for example JOYSTICK_SWAP_STICKS=True), but then it would be applicable only for specific devices (so JOYSTICK_SWAP_STICKS would be useless for certain devices)
2. creating new class for specific device with modified mapping (just as in this PR)

tikurahul:
1 sounds messy.
For option 2, you have another choice. Have the swapped controller extend the current `XBoxOneJoyStickController` and override the one method you need to. That way you don't have to duplicate everything.

nvtkaszpir:
Yeah, thanks for the inf, but it still would require to define new controller type.

tawnkramer:
it's just a dict of names and function pointers, so you can also write a generic `swap steering and throttle` that works for all controllers. Just iterate over the dict and find the key that sets the steering, throttle, and then remap.

tawnkramer:
I will roll the new class into dev branch. leave the swapping up to you if you want to do it.

nvtkaszpir:
ok

PR #520, opened 2019-12-28
Add option to swap sticks on joysticks

nvtkaszpir:
Allows to have throttle on left and steering on right stick of the joystick.
Also it allows to swap throttle with steering on RC controller.

This makes #518 obsolete.

### DIFF
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -651,7 +651,8 @@ class JoystickController(object):
                  steering_scale=1.0,
                  throttle_dir=-1.0,
                  dev_fn='/dev/input/js0',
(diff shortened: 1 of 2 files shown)


tawnkramer:
```
def get_key_for_value(a_dict, value):
    for key, val in a_dict.items():
        if val == value:
            return key
    return None

steering_key = get_key_for_value(self.axis_trigger_map, self.set_steering)
throttle_key = get_key_for_value(self.axis_trigger_map, self.set_throttle)
self.axis_trigger_map[throttle_key] = self.set_steering
self.axis_trigger_map[steering_key] = self.set_throttle
```

tawnkramer:
please consider re-implenting with the code example given in previous message.

PR #521, opened 2019-12-29
Extend picamera with more parameters

nvtkaszpir:
I needed to adjust donkey to work with NoIR camera, so I extended picamera class with extra options.

I suggest experimenting with PICAMERA_IMAGE_EFFECT='emboss' or rotation :)

### DIFF
--- a/donkeycar/parts/camera.py
+++ b/donkeycar/parts/camera.py
@@ -11,7 +11,7 @@ class BaseCamera:
         return self.frame
 
 class PiCamera(BaseCamera):
-    def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=False, hflip=False):
+    def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=False, hflip=False, awb_mode='auto', awb_gains=(1.0, 1.0), brightness=50, color_effects=None, contrast=0, exposure_compensation=0, exposure_mode='auto', image_denoise=True, image_effect='none', image_effect_params=None, iso=0, meter_mode='average', rotation=0, saturation=0, sharpness=0, video_denoise=True, video_stabilization=False, zoom=(0.0, 0.0, 1.0, 1.0)):
         from picamera.array import PiRGBArray
         from picamera import PiCamera
         
@@ -22,6 +22,32 @@ class PiCamera(BaseCamera):
         self.camera.framerate = framerate
         self.camera.vflip = vflip
         self.camera.hflip = hflip
+        self.camera.awb_mode = awb_mode
+        self.camera.awb_gains = awb_gains
+
+        self.camera.brightness = brightness
+        self.camera.color_effects = color_effects
+        self.camera.contrast = contrast
+        self.camera.exposure_compensation = exposure_compensation
+        self.camera.exposure_mode = exposure_mode
+
+        self.camera.image_denoise = image_denoise
+        self.camera.image_effect = image_effect
+
+        # notice that some image effects_params CAN NOT be applied
+        if image_effect_params is not None:
+            self.camera.image_effect_params = image_effect_params
+
+        self.camera.exposure_mode = exposure_mode
+        self.camera.iso = iso
+        self.camera.meter_mode = meter_mode
+        self.camera.rotation = rotation
+        self.camera.saturation = saturation
+        self.camera.sharpness = sharpness
+        self.camera.video_denoise = video_denoise
+        self.camera.video_stabilization = video_stabilization
+        self.camera.zoom = zoom
+
         self.rawCapture = PiRGBArray(self.camera, size=resolution)
         self.stream = self.camera.capture_continuous(self.rawCapture,
             format="rgb", use_video_port=True)
--- a/donkeycar/templates/cfg_complete.py
+++ b/donkeycar/templates/cfg_complete.py
@@ -32,6 +32,25 @@ IMAGE_DEPTH = 3         # default RGB=3, make 1 for mono
 CAMERA_FRAMERATE = DRIVE_LOOP_HZ
 CAMERA_VFLIP = False
 CAMERA_HFLIP = False
+# PI Camera specitic options, for more details see https://picamera.readthedocs.io/en/release-1.13/api_camera.html
+PICAMERA_AWB_MODE = 'auto'  # set camera auto white balance mode, if supported
+PICAMERA_AWB_GAINS = (1.0, 1.0)  # set awb gains as tuple, works only if awb mode is set to 'off'
+PICAMERA_BRIGHTNESS = 50  # set camera brightness, from 0 to 100, defaul 50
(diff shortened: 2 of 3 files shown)


nvtkaszpir:
1. There are many additional parameters to control camera, which are very useful when handling car in different light conditions or with different camera model (like NoIR cameras) or in different mounting options (rotation). Frankly speaking I was surprised that it was not implemented earlier - I suspect that most people used normal cameras in very good lighting conditions and the issues were non existent, thus no need to implement improvements for the camera performance. Also using specific camera in specific conditions may render it really hard to use (for example the images from NoIR camera in certain light condition become extremely pink and may present issues for some people).

2. Yes it's exposes quite a lot of parameters, but only for PiCamera. I skipped some parameters but indeed looking at it now on the next day morning the list may be too excessive.

The list of parameters is indeed long but I tried to keep the variables in config prefixed with PICAMERA_ so that it is easily distinguishable from other cameras.
Also that's why there are config.py to set default values, so that user does not have to have all of them in the config.

3. There are already exposed parameters that may break backward compatibility or are applied only to picam (vflip/hflip for example), but on the other hand used pi camera python module is rather well maintained and there may be hardly any other alternative out there.
But then I believe other python modules are more problematic than camera itself (for example bumping  tensorflow to 2.x in the future?)

4.There are already parameters in the configs that are not usable for everyone (just take example of parameters for different cameras). Maybe it is time to think about refactoring the config itself?

tikurahul:
I agree it’s useful to have these options. My disagreement is about config bloat. I was actually not thrilled about some of the other configuration options that you mention before. And I think we might want to undo those changes as well. 

I think making everything a configuration has diminishing returns because at this point the amount of config for the camera actually exceeds the lines of code. I would much rather advanced users forked the repo to add their camera optimizations.

adammconway:
Would breaking this out into a separate config file for camera settings make sense?  I know that can cause other issues down the line, but at least it could be retired as we move to new cameras etc.

tawnkramer:
I like that the camera part has all the options available. I'm not 100% confident those defaults are the ones we want. Have you looked at the driver code to see what the defaults are?

nvtkaszpir:
Yes, I've taken defaults from the python PiCamera library.

donkeycar/parts/camera.py
```diff
@@ -11,7 +11,7 @@ def run_threaded(self):
         return self.frame
 
 class PiCamera(BaseCamera):
-    def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=False, hflip=False):
+    def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=False, hflip=False, awb_mode='auto', awb_gains=(1.0, 1.0), brightness=50, color_effects=None, contrast=0, exposure_compensation=0, exposure_mode='auto', image_denoise=True, image_effect='none', image_effect_params=None, iso=0, meter_mode='average', rotation=0, saturation=0, sharpness=0, video_denoise=True, video_stabilization=False, zoom=(0.0, 0.0, 1.0, 1.0)):
```
DocGarbanzo, line 14: I think all the additional parameters should be passed by dictionary (i.e. **kwargs) here, otherwise the interface is overloaded.

donkeycar/parts/camera.py
```diff
@@ -22,6 +22,32 @@ def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=Fals
         self.camera.framerate = framerate
         self.camera.vflip = vflip
         self.camera.hflip = hflip
+        self.camera.awb_mode = awb_mode
```
DocGarbanzo, line 25: Then here, you would iterate through all the possible parameters of the camera using `get()` and providing a default values. Hence only required overwrites would need to be in the dictionary.

donkeycar/templates/cfg_complete.py
```diff
@@ -32,6 +32,25 @@
 CAMERA_FRAMERATE = DRIVE_LOOP_HZ
 CAMERA_VFLIP = False
 CAMERA_HFLIP = False
+# PI Camera specitic options, for more details see https://picamera.readthedocs.io/en/release-1.13/api_camera.html
```
DocGarbanzo, line 35: And here, instead of passing all camera parameters, you would add a single dictionary as camera parameters. This should be empty, here and then different users can overwrite this to their needs. This should also address @tikurahul comment.

donkeycar/templates/complete.py
```diff
@@ -98,7 +98,7 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type
             inputs = ['angle', 'throttle']
         elif cfg.CAMERA_TYPE == "PICAM":
             from donkeycar.parts.camera import PiCamera
-            cam = PiCamera(image_w=cfg.IMAGE_W, image_h=cfg.IMAGE_H, image_d=cfg.IMAGE_DEPTH, framerate=cfg.CAMERA_FRAMERATE, vflip=cfg.CAMERA_VFLIP, hflip=cfg.CAMERA_HFLIP)
+            cam = PiCamera(image_w=cfg.IMAGE_W, image_h=cfg.IMAGE_H, image_d=cfg.IMAGE_DEPTH, framerate=cfg.CAMERA_FRAMERATE, vflip=cfg.CAMERA_VFLIP, hflip=cfg.CAMERA_HFLIP, awb_mode=cfg.PICAMERA_AWB_MODE, awb_gains=cfg.PICAMERA_AWB_GAINS, brightness=cfg.PICAMERA_BRIGHTNESS, color_effects=cfg.PICAMERA_COLOR_EFFECTS, contrast=cfg.PICAMERA_CONTRAST, exposure_compensation=cfg.PICAMERA_EXPOSURE_COMPENSATION, exposure_mode=cfg.PICAMERA_EXPOSURE_MODE, image_denoise=cfg.PICAMERA_IMAGE_DENOISE, image_effect=cfg.PICAMERA_IMAGE_EFFECT, image_effect_params=cfg.PICAMERA_IMAGE_EFFECT_PARAMS, iso=cfg.PICAMERA_ISO, meter_mode=cfg.PICAMERA_METER_MODE, rotation=cfg.PICAMERA_ROTATION, saturation=cfg.PICAMERA_SATURATION, sharpness=cfg.PICAMERA_SHARPNESS, video_denoise=cfg.PICAMERA_VIDEO_DENOISE, video_stabilization=cfg.PICAMERA_VIDEO_STABILIZATION, zoom=cfg.PICAMERA_ZOOM)
```
DocGarbanzo, line 101: Then here you would just add `cfg.CAMERA_ARGS` - or whatever you want to call it.

nvtkaszpir: Thinking about var naming:
- `CAMERA_ARGS` one could be used by any camera (but then it would be on user side to keep compatible parameters)
- `PICAMERA_ARGS` would expliclty control picamera parameters, based on the module version. But this inroduces more config options bloat.

Moreover:
- should the args contain also image_w, image_h options or just extra args?

DocGarbanzo: I would opt for PICAMERA_ARGS and keep the current `image_w`, etc parameters to maintain backward compatibility.

PR #523, opened 2019-12-29
[WIP] Introduce python code formatter using black

### LINKED ISSUE #522: Automatic python code formatting
After looking around python files I believe the project could benefit from using python code formatters, for example https://github.com/psf/black (I strongly suggest to look into Black Playground page https://black.now.sh/ ).

This could solve multiple issues automatically such as:
- linter warnings
- docstrings formatting
- whitespaces
- extremely long lines
and many others.

Some languages already enforce formatters by default, which tend to make code read much easier.

For now I tried black and it works pretty well, but it needs to be instructed to process files to format according to python 3.5.

At least my editor shows way less colorful icons on the gutter due to solved linter issues ;)

Notice though, that it will introduce massive amount of code changes, so I suggest to make one PR with each commit which would reformat one file and extended CI checks.

What do you think?

nvtkaszpir:
Fixes #522

Yes, this looks massive and scary.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        language_version: python3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ jobs:
   allow_failures:
     - name: markdownlint
     - name: yamllint
-    - name: black
 
   include:
 
--- a/donkeycar/__init__.py
+++ b/donkeycar/__init__.py
@@ -1,11 +1,11 @@
-__version__ = '3.1.1'
+__version__ = "3.1.1"
 
-print('using donkey v{} ...'.format(__version__))
+print("using donkey v{} ...".format(__version__))
 
 import sys
 
 if sys.version_info.major < 3:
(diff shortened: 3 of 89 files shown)


tikurahul:
Thank you for taking the initiative. 

I think we might want to use something standard like `yapf` (https://github.com/google/yapf) and `pylint` in combination.

tawnkramer:
I'm not crazy about these "fixes". Most of them I don't like. The only one I approve of is adding a space after a comma. Do these really bother you guys?

nvtkaszpir:
This PR change looks much better when viewing code directly in the editor and not via github web diff view. Also it takes a moment to get used to it (I was also reluctant before started to use it), and after some time it just works and developer do not have to think about formatting because there is not much you can alter in black as compared to other tools.

Current PR has one main issue with the formatting of the lines with very long inline comments, for example template/config.py indeed looks horrible (round brackets everywhere), so I've decided to add [WIP] prefix to the PR.

YAPF, as stated in their official documentation, may have issues with formatting python code version below 3.6.4+, and it formats to the destination python version it runs under. Black, on the other hand, running under python 3.6+ can format the code according to python 3.5 rules.
Of course this could be somewhat mitigated by enforcing python 3.6.5 in the project itself.

tawnkramer:
thanks for suggestion. Will consider for the future. please run this formatter locally.

PR #524, opened 2019-12-29
add KerasMotion model training using RealSense T265 camera

bo-rc:


bo-rc:
this is created by mistake, closing.

blown302:
@bo-rc , do you have accurate points from the RealSense? I'm working on providing it odom values to improve accuracy. When I was using it without odom I was getting some bad drift unless the conditions were perfect.

PR #525, opened 2019-12-30
Arduino drivetrain support

jithu83:
Certain boards like [LattePanda](https://www.lattepanda.com/products/lattepanda-delta-432.html) have host cpu and arduino combined on a single board. 

In such cases it would be nice to have the  Arduino  generate  PWM signals for controlling
steering/throttle of the car (thus avoiding PCA9685)

[Standard Firmata](https://www.arduino.cc/en/reference/firmata) sketch needs to be loaded on the Arduino to allow it to be managed from the host-cpu runnig python app.

More Context of implementation/working is added in  [docs/parts/actuators.md](https://github.com/jithu83/donkeycar/blob/jj_arduino/docs/parts/actuators.md)


deltaflyer:
Dear @jithu83 I like the idea of using a cheap Arduino as PWM generator for the drive train. I checked your wiring diagrams. The problems with your wiring is that the entire servo power the transferred through the Arduino. In large RC cars the servo can drag up to 2 amps, which lead to a destruction of the Arduino.

To fix this issue, an ESC with a battery elemininator circuit (BEC) is needed (which most ESC provide). This type of ESC provides 5V on the connecto cable to power servos and other RC peripherals.

Below, I provided a fixed version with an BEC-enabled ESC. In this drawing, the Arduino only provides the signal path and the power for the servo is provided by the ESC. The Arduino prodivdes the PWM signals and a reference GND and the ESC provides 5V. Since the Arduino is powered by USB-port, you should connect GND also to the GND of the main computer to get an identical GND level.

ArduinoWiring_new

jithu83:
@deltaflyer appreciate your thoughtful review comments w.r.t servo motor pulling more current from Arduino VCC. I don't mind updating the patch-set to reflect the updated wiring you suggested, however the following doubt is nagging me:

In the stock donkey build using PCA 9685, I believe the power to the servo is fully provided by the raspberryPI / Jetson nano (and not the BEC from ESC) - I am wondering whether similar concerns are applicable there ? (Look for the violet wire from rpi connecting to PCA's vcc [here](https://docs.donkeycar.com/guide/build_hardware/) under _Step 4: Connect Servo Shield to Raspberry Pi._). Image duplicated below:
image

Or is it from the throttle PWM's  V+  in the PCS 9685 case also ? I will await your reply before updating the patchset

adammconway:
The 9685 component is powered by the RPi but the servo is powered by the ESC.  Basically there are 2 voltage rails - 5V from the RPI and 5V from the ESC.

dlarue:
@adammconway  the PCA9586 communications side is 3.3V, not 5V since the rPi I/O is 3.3V and not 5V tolerant. The PWM outputs are on an open voltage rail but in our case it is driven/powered by the ESC BEC which is generally 5-6VDC but some have powered it at higher voltages after isolating the BEC output from the ESC PWM cable.

deltaflyer:
Hello  @adammconway Hello @jithu83 when looking photo of the donkey car doku posted above, we see that the V+ pin is not used and the ESC powers the servo. Looking at the PCB schematics of an Adafruit PCA 6985 servo controller, we see the connections between the RC-peripherals' pins is linking the ESC power to the servo on the 5V rail. The vcc ping with 3,3V rail is comming from the computer and is powering the PCA 6985 controller.

adammconway:
Ah, sorry... you are right, it is 3.3V from the Pi.

jithu83:
Thanks @deltaflyer @adammconway @dlarue  for clarifying the PCA9685 scenario. 
I have updated the  Arduino wiring diagram to use the ESC BEC power for the servo. 
Further I also provided a bit for info about how to load standard firmata sketch on Arduino
This is the updated [documentation](https://github.com/jithu83/donkeycar/blob/jj_arduino/docs/parts/actuators.md). 

Note to maintainer : There are 4 commits in total, the top 2 have been revised (I was a bit confused by the github pull request updating flow)

jithu83:
Looks like the discussion around this approach has concluded in hardware channel on slack. Can we have this merged please @tawnkramer and other maintainers.

deltaflyer:
Hi all, what I really like about this PR is that it enables to run Donkey car with all kind of computers. It removes the needs for I2C or SPI to have working connection to servos and ESC. Your driving computer hardware just needs USB and that's it.

dlarue:
Unfortunately @deltaflyer the USB connector design is not intended for use in vibration rich environments. So if you plan on not crashing your vehicle, you should consider breaking USB out and onto a better connector.  A UART is a good solution and had already been done in the Teensy firmware and the Teensy part in donkey. Unfortunately the UART on rPi takes a few hoops jumped to get control/use of it so not as easy as USB and not as general purpose if expecting to be able to just swap to another board without changes/customization.

deltaflyer:
Hello @dlarue thank you for your feedback and this is a very interesting topic. Fully agree, unmodified USB is not suitable for vibrations and physical stress. The last years a build several drones and indeed most USB connectors without modification are painful sources of errors. Since a few years I either user alternative plugs for vibration intensive setups or hot glue to fix plugs together permanently at first place. For almost all mini-computer boards USB is present as "jumper" pins there is not needs for using an USB cable with plugs at all. Thus, building resilient vehicle hardware is one topic that needs special attention. But poor connector design does not affect software support and thus I cannot see any obstacles for having a pure USB driving stack that is way more cheaper that specialized hats/sombreros.

tawnkramer:
Nice PR! thanks for docs as well.

PR #535, opened 2020-01-30
Update build_hardware.md

dlarue:
For rPi 4B and Jetson Nano processor boards the higher power comes at a price of higher input power and since the voltage remains the same( 5V ) the current must increase. So it's recommended you get a USB battery pack with 3A output capabilities or higher.


tawnkramer:
Thanks!

PR #539, opened 2020-02-08
add precsion on convert-to-uff feature

macfly1202:
Add location of the converter and make path absolute to be sure everything is clear for everyone


tawnkramer:
this is not needed. You can't run `python manage` unless you are in the mycar dir. And people can name the mycar dir anything they want. So makes this less flexible.

PR #540, opened 2020-02-08
update tensorrt conversion (version 2)

macfly1202:
Trail to make more clear documentation, fluent.

### DIFF
--- a/docs/guide/robot_sbc/tensorrt_jetson_nano.md
+++ b/docs/guide/robot_sbc/tensorrt_jetson_nano.md
@@ -66,12 +66,16 @@ After you train the `linear` model you end up with a file with a `.h5` extension
 ```bash
 # You end up with a Linear.h5 in the models folder
 python manage.py train --model=./models/Linear.h5 --tub=./data/tub_1_19-06-29,...
-# Freeze model using freeze_model.py in donkeycar/scripts
-# The frozen model is stored as protocol buffers.
+
+# (optional) copy './models/Linear.h5' from your desktop computer to your Jetson Nano in your working dir (~mycar/models/)
+
+# Freeze model using freeze_model.py in donkeycar/scripts ; the frozen model is stored as protocol buffers.
 # This command also exports some metadata about the model which is saved in ./models/Linear.metadata
-python freeze_model.py --model=./models/Linear.h5 --output=./models/Linear.pb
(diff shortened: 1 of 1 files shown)


docs/guide/robot_sbc/tensorrt_jetson_nano.md
```diff
@@ -66,12 +66,16 @@ After you train the `linear` model you end up with a file with a `.h5` extension
...
+# (optional) copy './models/Linear.h5' from your desktop computer to your Jetson Nano in your working dir (~mycar/models/)
+
+# Freeze model using freeze_model.py in donkeycar/scripts ; the frozen model is stored as protocol buffers.
 # This command also exports some metadata about the model which is saved in ./models/Linear.metadata
-python freeze_model.py --model=./models/Linear.h5 --output=./models/Linear.pb
+python ~/projects/donkeycar/scripts/freeze_model.py --model=~/mycar/models/Linear.h5 --output=~/mycar/models/Linear.pb
+
 # Convert the frozen model to UFF. The command below creates a file ./models/Linear.uff
-convert-to-uff ./models/Linear.pb
+cd /usr/lib/python3.6/dist-packages/uff/bin/
```
nvtkaszpir: I suggest to add a section where to find convert_to_uff.py for given installation - It would make the documentation more generic, or add  a note that this refers to the default donkeycar installation.

Notice that not everyone is on python 3.6 or if someone uses something like pyenv then this will point to another path.

docs/guide/robot_sbc/tensorrt_jetson_nano.md
```diff
@@ -66,12 +66,16 @@ After you train the `linear` model you end up with a file with a `.h5` extension
 ```bash
 # You end up with a Linear.h5 in the models folder
 python manage.py train --model=./models/Linear.h5 --tub=./data/tub_1_19-06-29,...
```
nvtkaszpir: Do you think it would be beneficial to also update that line?

docs/guide/robot_sbc/tensorrt_jetson_nano.md
macfly1202: I think you are right about python version. the most important is to clearly help peole to find convert-to-uff which is a nightmare to find. Also, I prefer to indicate the filepath of the script in the command. Do what you consider the best. Fyi, it took me multiple trial to find this, and I'm not the only one if I refer to google... thanks for review !

tawnkramer:
thanks!

PR #544, opened 2020-02-19
Support for Realsense D435 and D435i depth cameras

Ezward:
- enabled using CAMERA_TYPE = 'D435'
- supports RGB, DEPTH and IMU output; each of which can be individually toggled with configuration REALSENSE_D435_xxxx
- supports addressing a camera with a specific ID in the case of multiple cameras (or in presense of T265)
- IMU output is compatible with MPU6050 and IMU model. (Not tested.)
- depth is output as a separate, aligned image in 16bit grayscale png format.  So pixels in depth image can be queried to find the depth of a pixel in the RGB image.

### DIFF
--- a/donkeycar/parts/datastore.py
+++ b/donkeycar/parts/datastore.py
@@ -236,8 +236,8 @@ class Tub(object):
                 img.save(os.path.join(self.path, name))
                 json_data[key]=name
 
-            elif typ == 'depth_array':
-                # save depth as a 16bit png
+            elif typ == 'gray16_array':
+                # save np.uint16 as a 16bit png
                 img = Image.fromarray(np.uint16(val))
                 name = self.make_file_name(key, ext='.png')
                 img.save(os.path.join(self.path, name))
--- a/donkeycar/parts/datastore.py
+++ b/donkeycar/parts/datastore.py
@@ -236,6 +236,13 @@ class Tub(object):
                 img.save(os.path.join(self.path, name))
                 json_data[key]=name
 
+            elif typ == 'depth_array':
+                # save depth as a 16bit png
+                img = Image.fromarray(np.uint16(val))
+                name = self.make_file_name(key, ext='.png')
+                img.save(os.path.join(self.path, name))
+                json_data[key]=name
+
             else:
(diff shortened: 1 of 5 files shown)


tawnkramer:
thanks for printing that exception on the vehicle loop!

donkeycar/parts/datastore.py
```diff
@@ -236,6 +236,13 @@ def put_record(self, data):
                 img.save(os.path.join(self.path, name))
                 json_data[key]=name
 
+            elif typ == 'depth_array':
```
tikurahul: Nit: call this something like `np_array` so we can support arbitrary array types going forward.

donkeycar/parts/realsense435i.py
```diff
@@ -0,0 +1,301 @@
...
+        self.height = height
+        self.channels = channels
+        self.resize = (width != WIDTH) or (height != height) or (channels != CHANNELS)
+        if self.resize:
+            print("The output images will be resized from {} to {}.  This requires opencv.".format((WIDTH, HEIGHT, CHANNELS), (self.width, self.height, self.channels)))
+
+        # Configure streams
+        self.imu_pipeline = None
+        if self.enable_imu:
+            self.imu_pipeline = rs.pipeline()
```
tikurahul: Do you need a pipeline per feature ?

tikurahul:
Your changes look great in general. I am okay with you fixing them post too (if needed)

donkeycar/parts/realsense435i.py
Ezward: Not a pipeline per feature, but IMU does _have_ to be a separate pipeline.  It doesn't work otherwise.

donkeycar/parts/datastore.py
Ezward: It's not totally generic.  This line `img = Image.fromarray(np.uint16(val))` requires a 16 bit array as input.  Then it always outputs that to a png.  so we could change it be more generic a name, but we want to keep the semantics of the constraint.  So something like 'gray16_array' or 'np16_array'.

Issue #549, opened 2020-03-01
Add support for MPU9250 IMU


AlessioMorale:
Add support for MPU9250 (logically an MPU6500+AK8963).

AlessioMorale:
This is now implemented by #550

PR #550, opened 2020-03-01
Add support for MPU9250

AlessioMorale:
This MR adds support for MPU9250 (#549). 

To configure an mpu9250 just enable 
` HAVE_IMU = True` 
and set 
`IMU_SENSOR = 'mpu9250'`

### DIFF
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -73,5 +73,5 @@ Valid settings are from 0 to 6:
 - `6` 5Hz
 
 ### Notes on MPU9250
-At startup the MPU9250 driver perform a calibration to zero accel and gyro bias. 
+At startup the MPU9250 driver performs calibration to zero accel and gyro bias. Usually the process takes less than 10 seconds, and in that time avoid moving or touching the car.
 Please place the car on the ground before starting Donkey 
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -61,7 +61,7 @@ IMU_DLP_CONFIG = 3
 ```
 `IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using.
 
-`IMU_DLP_CONFIG` allows to change the digital lowpass filter settings for your imu. Lower frequency settings can filter high frequency noise at the expense of increased latency in imu sensor data.
+`IMU_DLP_CONFIG` allows to change the digital lowpass filter settings for your IMU. Lower frequency settings (see below) can filter high frequency noise at the expense of increased latency in IMU sensor data.
 Valid settings are from 0 to 6:
 
 - `0` 250Hz
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -51,7 +51,8 @@ pip install mpu9250-jmdev
 ```
 
 ### Configuration
-Enable the following configurations to your `myconfig.py`
+Enable the following configurations to your `myconfig.py`:
+
 ``` python
 #IMU
 HAVE_IMU = True
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -59,7 +59,7 @@ HAVE_IMU = True
 IMU_SENSOR = 'mpu9250'          # (mpu6050|mpu9250)
 IMU_DLP_CONFIG = 3
 ```
-`IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using
+`IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using.
 
 `IMU_DLP_CONFIG` allows to change the digital lowpass filter settings for your imu. Lower frequency settings can filter high frequency noise at the expense of increased latency in imu sensor data.
 Valid settings are from 0 to 6:
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -57,7 +57,7 @@ Enable the following configurations to your `myconfig.py`:
 #IMU
 HAVE_IMU = True
 IMU_SENSOR = 'mpu9250'          # (mpu6050|mpu9250)
-IMU_DLP_CONFIG = 6
+IMU_DLP_CONFIG = 3
 ```
 `IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using
 
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -6,7 +6,7 @@ IMUs or inertial measurement units are parts that sense the inertial forces on a
 
 This is a cheap, small, and moderately precise imu. Commonly available at [Amazon](https://www.amazon.com/s/ref=nb_sb_noss_2?url=search-alias%3Dindustrial&field-keywords=MPU6050).
 
-MPU9250 offer in addition a integrated magnetometer
+MPU9250 offers additional integrated magnetometer.
 
 * Typically uses the I2C interface and can be chained off the default PWM PCA9685 board. This configuration will also provide power.
 * MPU6050: Outputs acceleration X, Y, Z, Gyroscope X, Y, Z, and temperature.
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -74,4 +74,4 @@ Valid settings are from 0 to 6:
 
 ### Notes on MPU9250
 At startup the MPU9250 driver performs calibration to zero accel and gyro bias. Usually the process takes less than 10 seconds, and in that time avoid moving or touching the car.
-Please place the car on the ground before starting Donkey 
+Please place the car on the ground before starting Donkey.
--- a/docs/parts/imu.md
+++ b/docs/parts/imu.md
@@ -2,13 +2,15 @@
 
 IMUs or inertial measurement units are parts that sense the inertial forces on a robot. They vary depending on sensor, but may commonly include linear and rotational accelleration. They may sometimes include magnetometer to give global compasss facing dir. Frequently temperature is available from these as it affects their sensitivity.
 
-## MPU6050
+## MPU6050/MPU9250
 
 This is a cheap, small, and moderately precise imu. Commonly available at [Amazon](https://www.amazon.com/s/ref=nb_sb_noss_2?url=search-alias%3Dindustrial&field-keywords=MPU6050).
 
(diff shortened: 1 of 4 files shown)


tikurahul:
Thanks for the PR. 

I have been using a MPU 9250 as well, but pretending to be a 6050. I am still able to obtain all the sensor data. What does this add ?

AlessioMorale:
Hi, thanks for reviewing this.
There are basically two reasons: 
The first is that it allows to access magnetometer data (they are already exposed, i'm going to play with them soon to understand if they may provide meaningful info).
The second is that it also provides a `calibration` functionality to zero the bias. (I have committed this change just now)
In the meantime i spotted an issue with accelerometer scale (it was using `g` rather than `m/s^2` I fixed this issue as well in my latest commits)

donkeycar/templates/complete.py
```diff
@@ -263,8 +263,8 @@ def show_record_acount_status():
 
     #IMU
     if cfg.HAVE_IMU:
-        from donkeycar.parts.imu import Mpu6050
-        imu = Mpu6050()
+        from donkeycar.parts.imu import MpuIMU
+        imu = MpuIMU(sensor=cfg.IMU_SENSOR)
```
tikurahul: Nit. Just call it `IMU`.

donkeycar/parts/imu.py
```diff
@@ -12,13 +15,36 @@ class Mpu6050:
...
     '''
 
-    def __init__(self, addr=0x68, poll_delay=0.0166):
-        from mpu6050 import mpu6050
-        self.sensor = mpu6050(addr)
+    def __init__(self, addr=0x68, poll_delay=0.0166, sensor=SENSOR_MPU6050):
+        self.sensortype = sensor
+        if self.sensortype == SENSOR_MPU6050:
+            from mpu6050 import mpu6050
+            self.sensor = mpu6050(addr)
```
tikurahul: If `mpu6050` is a class just change it to `MPU6050`.

That way it's clearer that this is not a invocation, but an instantiation.

tikurahul:
You probably want to add some documentation around how to calibrate this correctly.

AlessioMorale:
> You probably want to add some documentation around how to calibrate this correctly.

Yes, definitely! Tomorrow I should be able to perform a few more tests and write down some doc accordingly.

donkeycar/parts/imu.py
```diff
@@ -30,7 +57,14 @@ def update(self):
...
         try:
-            self.accel, self.gyro, self.temp = self.sensor.get_all_data()
+            if self.sensortype == SENSOR_MPU6050:
+                self.accel, self.gyro, self.temp = self.sensor.get_all_data()
+            else:
+                from mpu9250_jmdev.registers import GRAVITY
+                ret = self.sensor.getAllData()
+                self.accel = { 'x' : ret[1] * GRAVITY, 'y' : ret[2] * GRAVITY, 'z' : ret[3] * GRAVITY }
+                self.gyro = { 'x' : ret[4], 'y' : ret[5], 'z' : ret[6] }
+                self.mag = { 'x' : ret[13], 'y' : ret[14], 'z' : ret[15] }
```
nvtkaszpir: You add magnetometer  data in here but it is unused at all, because in ` donkeycar/templates/complete.py` in line 268 you only use inputs from accelerometer and gyroscope.

Also AFAIR if adding magnetometer data is ok, but in reality it may be confusing for trained model if someone moves to another location (like from EU to USA).
So I think that for driving car the magnetometer data can be dropped (as it is now).

Otherwise it could be exposed if someone would like to use this data for something like geospatial vizualization?

nvtkaszpir:
MPU9250 has integrated Digital Low Pass Filter unit, are you going to integrate that as well?
It allows to smooth readings to get more stable values for the price of increased latency.
https://cloud.githubusercontent.com/assets/18668000/14822142/6e3d82fc-0b9b-11e6-827c-1b7113eefff6.png

Sample code:
```python
DLPF_CFG = 3 # default is 0 max is 6
...
        self.sensor = mpu6050(addr) // we can treat mpu9250 just as mpu6050
        self.sensor.bus.write_byte_data(addr, 0x1A, DLPF_CFG)
        self.accel = { 'x' : 0., 'y' : 0., 'z' : 0. }
        self.gyro = { 'x' : 0., 'y' : 0., 'z' : 0. }
...
```

This is especially visible when car is not moving, and the acceleration/gyro can get reading improved for an order of magnitude.

donkeycar/parts/imu.py
AlessioMorale: I'm going to progress in several steps about the magnetometer. I have previously worked with drones using several MPU family members. For this reason I had to deal a lot with  magnetometers. 
For this reason the first step is collecting data e trying to understand the quality of what i can get from it. I want to understand the effect of surrounding elements as well (so entity of both soft/hard iron distortion without complex imu installations). this will help me understand how usable it is and if I actually need a proper calibration to get something useful from it.

tikurahul: Keep the `magnetometer` data. Even though it's unused for know - the sensor data should be available for future models & approaches.

tikurahul:
You can add the digital LPF as an flag to pass to the constructor of the IMU to enable it. That way the user can pick what is more important. Awesome work. Looking forward to landing this as you make the last couple of minor changes.

AlessioMorale:
Is there any reason why it was not used in 6050? I prefer to configure the 9250 in the same way as 6050 for now (although there are some subtle differences, the 9250 is slightly more sensitive to vibrations but I don't know, compared to drones, how is an RC car from this point of view, time to test), then go back to this once I have more data. 

> You can add the digital LPF as an flag to pass to the constructor of the IMU to enable it. That way the user can pick what is more important. Awesome work. Looking forward to landing this as you make the last couple of minor changes.

docs/parts/imu.md
```diff
@@ -2,13 +2,15 @@
 
 IMUs or inertial measurement units are parts that sense the inertial forces on a robot. They vary depending on sensor, but may commonly include linear and rotational accelleration. They may sometimes include magnetometer to give global compasss facing dir. Frequently temperature is available from these as it affects their sensitivity.
 
-## MPU6050
+## MPU6050/MPU9250
 
 This is a cheap, small, and moderately precise imu. Commonly available at [Amazon](https://www.amazon.com/s/ref=nb_sb_noss_2?url=search-alias%3Dindustrial&field-keywords=MPU6050).
 
+MPU9250 offer in addition a integrated magnetometer
```
nvtkaszpir:
```suggestion
MPU9250 offers additional integrated magnetometer.
```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+For MPU9250: 
+
+Install pip lib for `mpu9250-jmdev`:
+
+```bash
+pip install mpu9250-jmdev
+```
+
+### Configuration
+Enable the following configurations to your `myconfig.py`
```
nvtkaszpir:
```suggestion
Enable the following configurations to your `myconfig.py`:

```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+``` python
+#IMU
+HAVE_IMU = True
+IMU_SENSOR = 'mpu9250'          # (mpu6050|mpu9250)
+IMU_DLP_CONFIG = 6
```
nvtkaszpir:
```suggestion
IMU_DLP_CONFIG = 3
```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+```
+`IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using
```
nvtkaszpir:
```suggestion
`IMU_SENSOR` can be either `mpu6050` or `mpu9250` based on the sensor you are using.
```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+
+`IMU_DLP_CONFIG` allows to change the digital lowpass filter settings for your imu. Lower frequency settings can filter high frequency noise at the expense of increased latency in imu sensor data.
```
nvtkaszpir:
```suggestion
`IMU_DLP_CONFIG` allows to change the digital lowpass filter settings for your IMU. Lower frequency settings (see below) can filter high frequency noise at the expense of increased latency in IMU sensor data.
```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+- `0` 250Hz
+- `1` 184Hz
+- `2` 92Hz
+- `3` 41Hz
+- `4` 20Hz 
+- `5` 10Hz
+- `6` 5Hz
+
+### Notes on MPU9250
+At startup the MPU9250 driver perform a calibration to zero accel and gyro bias.
```
nvtkaszpir:
```suggestion
At startup the MPU9250 driver performs calibration to zero accel and gyro bias. Usually the process takes less than 10 seconds, and in that time avoid moving or touching the car.
```

docs/parts/imu.md
```diff
@@ -32,10 +34,43 @@ cd py-smbus/library
...
+- `1` 184Hz
+- `2` 92Hz
+- `3` 41Hz
+- `4` 20Hz 
+- `5` 10Hz
+- `6` 5Hz
+
+### Notes on MPU9250
+At startup the MPU9250 driver perform a calibration to zero accel and gyro bias. 
+Please place the car on the ground before starting Donkey
```
nvtkaszpir:
```suggestion
Please place the car on the ground before starting Donkey.
```

donkeycar/parts/imu.py
```diff
@@ -47,10 +94,18 @@ def shutdown(self):
 
 if __name__ == "__main__":
     iter = 0
-    p = Mpu6050()
+    import sys
+    sensor_type = SENSOR_MPU6050
```
nvtkaszpir: main function allows testing only for MPU6050, shouldn't there be an extra argument to pass test for MPU9250?

AlessioMorale: It defaults to mpu6050 and DLP=0 but you can pass two additional parameters: sensor type and DLP settings

PR #556, opened 2020-03-14
- update complete.py for 'custom' CONTROLLER_TYPE to load my_joystick.py

Ezward:
Motivation: I found that the default xbox controller does not work on the Nvidia Jetson Nano because the driver maps the joysticks differently.  Using the `donkey createjs` command to create a custom mapping is a great solution, but maintaining patches to manage.py is a problem.  This update makes it much simpler to use the output of the createjs comment.

- user can now choose 'custom' CONTROLLER_TYPE in myconfig.py and it will load the controller from the my_joystick.py file in the mycar folder.
- update cfg_complete.py to show the option
- updated donkey.md docs to remove the code for integrating the my_joystick.py, since that is all done for the user now.


donkeycar/templates/complete.py
```diff
@@ -137,6 +137,17 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type
             from donkeycar.parts.robohat import RoboHATController
             
             ctr = RoboHATController()
+        elif "custom" == cfg.CONTROLLER_TYPE:
+            #
+            # custom controller created with `donkey createjs` command
+            #
+            from my_joystick import MyJoystickController
```
nvtkaszpir: Maybe add some checks if the module import worked, so that people will not get a generic python error but something more obvious with explanation how to fix it?
Some people just don't read docs...

tawnkramer:
Thanks for this! Love the additional docs.

PR #560, opened 2020-03-18
Fix USE_FPV

sctse999:
1. Fallback USE_FPV to use 'cam/image_array' as inf_input is normalized.
2. The video stream should fit the screen by height instead of width

Note: By using 'cam/image_array' , crop (if set) is not shown in the FPV. Just aim to fix the FPV at the moment as it is a bit tricky to incorporate the crop in the current codebase. Welcome any feedback. Thanks.

### DIFF
--- a/donkeycar/parts/web_controller/templates/base_fpv.html
+++ b/donkeycar/parts/web_controller/templates/base_fpv.html
@@ -69,7 +69,7 @@
 
 <div class="gap"></div>
 <div>    
-    <img id='mpeg-image', class='img-responsive' src="/video"/> </img>
+    <img id='mpeg-image' class='img-responsive' src="/video"/> </img>
 </div>
 
 </body>
--- a/donkeycar/parts/web_controller/templates/base_fpv.html
+++ b/donkeycar/parts/web_controller/templates/base_fpv.html
@@ -9,7 +9,6 @@
             display: block;
             margin-left: auto;
             margin-right: auto;
-            width: 100%;
             height: 100%;
             border: 2px solid rgb(226, 110, 57);
             border-radius: 8px;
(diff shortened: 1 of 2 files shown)


donkeycar/parts/web_controller/templates/base_fpv.html
```diff
@@ -45,23 +44,33 @@
...
-    </div>
+    <img class="logo" src="/static/donkeycar-logo-sideways.png" alt="Donkeycar Logo">
 </div>
 
 <body style="background-color:rgb(48, 49, 51);">
 
 <div class="gap"></div>
-<img id='mpeg-image', class='img-responsive' src="/video"/> </img>
+<div>    
+    <img id='mpeg-image', class='img-responsive' src="/video"/> </img>
```
nvtkaszpir: I believe that comma in here is not needed.

sctse999:
This fix does not support cropping. It just fallback to use the original image captured by the camera. 

Before this commit, the FPV use ```inf_input``` as the input which is a normalized image that is not usable as a mjpeg.  I revert this by using the original image, which is the ```cam/image_array```.

If we want to support cropping, we might need to decouple the normalize and cropping operation in ```normalize_and_crop``` in [utils.py](https://github.com/autorope/donkeycar/blob/77b1dc76c03657e887428a99929c14f78bbb865f/donkeycar/utils.py#L133)

donkeycar/templates/complete.py
```diff
@@ -301,7 +301,7 @@ def run(self, img_arr):
 
     # Use the FPV preview, which will show the cropped image output, or the full frame.
     if cfg.USE_FPV:
-        V.add(WebFpv(), inputs=[inf_input], threaded=True)
+        V.add(WebFpv(), inputs=['cam/image_array'], threaded=True)
```
DocGarbanzo: I thought that was what I had in my original PR, maybe it got overwritten later.

donkeycar/parts/web_controller/templates/base_fpv.html
```diff
@@ -45,23 +44,33 @@
...
+    <img class="logo" src="/static/donkeycar-logo-sideways.png" alt="Donkeycar Logo">
 </div>
 
 <body style="background-color:rgb(48, 49, 51);">
 
 <div class="gap"></div>
-<img id='mpeg-image', class='img-responsive' src="/video"/> </img>
+<div>    
+    <img id='mpeg-image' class='img-responsive' src="/video"/> </img>
+</div>
```
DocGarbanzo: I checked your changes and they work well. Really better to fit by height than by width!

DocGarbanzo: @tawnkramer - I think you can merge this straight away.

PR #563, opened 2020-03-21
Add websocket to drive api

sctse999:
1. Add websocket drive api
2. Switch web controller to use the websocket api


tawnkramer:
this looks great! how does it perform?

PR #564, opened 2020-03-23
Add inverted pwm drivetrain

santerioksanen:
Instead of utilizing the servo board as in the instructions to drive servos, I built a custom board using MOSFETs to increase PWM voltage from 3.3 V provided by Raspberry to 5 V understood by the servo and motor controller. The circuit inverts the PWM signal, for which I added the inverted parameter to parts/actuator.py.

The circuit is simple: https://www.openimpulse.com/blog/2012/06/unidirectional-voltage-level-translators/, though had to use 1 k resistors instead of 10 k due to resistance in the control bus of both motor controller and the servo.


tawnkramer:
thanks!

PR #574, opened 2020-04-18
tubplot option fix

naisy:
I think,  tubplot option in management/base.py is mistaking.
--tub is only one. I changed help message.
--limit is integer. I add type=int.


tawnkramer:
Thanks!

PR #578, opened 2020-04-25
Allow data augmentation on tub outside of training

DocGarbanzo:
# Stand-alone data augmentation on tubs
The data augmentation can be enabled in training by passing the `--aug` flag. However, this setting slows down training by a bit, in particular if the tub contains > 100k records. This change contains the following:
* Adding a new member function to the `Tub` through which augmentation can be performed on the  images in place
* Adding a new donkey command `donkey tubaugment` to call the above function with the option `--inplace`. If omitted a new tub is created instead of the manipulation of the existing images.
* There is a small refactoring in `augment.py` to decouple the image transformation from numpy arrays to PIL and back. Also the color scaling has been increased from[0.0, 1.0] to [0.0, 2.0].
* Adding the new command `donkey tubaugment` to the docs
* Small bits of pep-8 reformatting

On my MacBook the image augmentation takes around 3ms, running this as part of the training w/ 100k records adds about 5min / epoch process time. On colab adding the `--aug` flag makes the training prohibitively slow on such a data set. I had to re-start the colab session kernel several times to get it to complete. Augmenting the tub data exogenously also allows to train on both sets of data.

### DIFF
--- a/docs/utility/donkey.md
+++ b/docs/utility/donkey.md
@@ -102,6 +102,20 @@ donkey tubcheck <tub_path> [--fix]
 * It will print the records that throw an exception while reading
 * The optional `--fix` will delete records that have problems
 
+## Augment Tub
+
+This command allows you to perform the data augmentation on a tub or set of tubs directly. The augmentation is also available in training via the `--aug` flag. Preprocessing the tub can speed up the training as the augmentation can take some time. Also you can train with the unmodified tub and the augmented tub joined together. 
+
+Usage:
+
+```bash
+donkey tubaugment <tub_path> [--inplace]
+```
+
+* Run on the host computer or the robot
+* The optional `--inplace` will replace the original tub images when provided. Otherwise `tub_XY_YY-MM-DD` will be copied to a new tub `tub_XX_aug_YY-MM-DD` and the original data remains unchanged
+
+
 ## Histogram
 
 This command will show a pop-up window showing the histogram of record values in a given tub.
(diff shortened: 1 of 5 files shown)


tawnkramer:
Very nice. Thanks for this! Is it ready for merge?

tikurahul:
FWIW, this won’t be necessarily one we switch to the new store.

DocGarbanzo:
@tawnkramer - yeah this can be merged.
@tikurahul - why do we not need this using the new tub? I would have thought we will have to port it to the new tub (I'm happy to do this and do a PR on your branch if it's up-to-date). The idea is to allow you augmenting your tub before going into training.

tikurahul:
@DocGarbanzo we don't need this feature, because I already planned for this feature from the get go. This is already done :)

An example is v1 -> v2 tub migration:
https://github.com/autorope/donkeycar/pull/532/commits/edbc3e6a88ea6452e55869a91fd9d8728c792368#diff-34122bc1699752be2209d5b740691feb

tikurahul:
I also added the ability to `augment` sequences just by wrapping the sequence itself. Here is an example: https://github.com/autorope/donkeycar/pull/532/commits/edbc3e6a88ea6452e55869a91fd9d8728c792368#diff-03fffa923184e50a01f4400403d966ea

DocGarbanzo:
@tikurahul - yeah I see your Augmentation class, this is great, seems to allow much more flexibility. I can't see though how this is applied in your tub conversion script, but it'll be easy enough to do. When is the new tub planned to be merged?

PR #579, opened 2020-04-26
Restrict support to python 3.6 and 3.7

DocGarbanzo:
# Stop supporting Python 3.5
Raspbian Buster is using Python 3.7. Previous OS Stretch was on Python 3.5 which now seems outdated. Raspberry Pi4 are coming with Buster but RPi 3 users should upgrade their OS. For compatibility to older Python installations on the PC version 3.6 is kept.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 ---
 language: python
-
+# this is required for python 3.7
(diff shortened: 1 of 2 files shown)


tawnkramer:
Thanks!

PR #580, opened 2020-04-28
Introduce workaround for some of the exceptions around installation on Stretch

soswow:
That's what I experienced during installation on Stretch with Python 3.5


tawnkramer:
Believe you want to be on Buster now. Python 3.6+

tawnkramer:
thanks for this contribution. But we are leaving behind support for 3.5 now. Keeping this would be confusing.

soswow:
No worries. I suggest stating this requirement (for Buster and Python >=3.6) very clear in docs and I will be happy =) As well as those who came with Stretch version like I did =)

PR #582, opened 2020-05-07
Fix tubhist figure storage location

jasonleung101:
Fixed the tubhist figure storage location from the root directory to target tub directory.


tawnkramer:
I don't like putting it in the tub dir. That's not where non-tub data goes. I understand that perhaps you don't want it to default to the current dir. Perhaps we can add an option to specify output path on command line?

jasonleung101:
Thanks for the suggestion. I added a new parameter called "out" for a user to specify the output path on the command line. If a user did not enter the specify the output path, the output will be set by default just like before. Otherwise, the histogram will store in the specified output path. Thank you.

jasonleung101:
Sorry for the missing argument.

tawnkramer:
looks good, thanks!

PR #587, opened 2020-05-27
Updated Robo HAT MM1 instructions

wallarug:
Added Hackster IO Links
Changed the link for the code.py to internal Donkey Car maintained version.


wallarug:
Can this also go through to master so it gets published immediately?

PR #591, opened 2020-06-05
Fix image in FPS profiling script

DocGarbanzo:
Fix image which was simply black after casting [0,1) values to int.


tawnkramer:
thanks

PR #592, opened 2020-06-07
Add real-time calibration 

sctse999:
Allow user to change config value such as STEERING_LEFT_PWM or MM1_STEERING_MID without restarting the car or running donkey calibrate command. I have added the following:

1. A much simplified verison of manage.py called "calibrate.py" to wire up the drive train to the web controller without changing manage.py. It will be created when user run createapp 

2. A /calibrate web page to perform the calibration

3. A websocket endpoint /wsCalibrate to support calibration message from mobile app and web page 

Currently the real-time calibration support PCA9685 and MM1 only. 

screencapture-yingwa-local-8887-calibrate-2020-06-07-16_17_06


tawnkramer:
Very cool!

PR #593, opened 2020-06-08
@peterpantechland changes for mapping to be correct.

wallarug:
As discussed with Peter.  There were strange values generated by the mapping function.  

Moving around the MID and top values should improve / fix this.


sctse999:
I need to look into the changes. Will get back to you.

wallarug:
kk ;)

sctse999:
I tested this but it looks like it is not compatible with web controller. It is flipping the web controller's input.

sctse999:
@tawnkramer we can just close this PR without merging. Thanks.

Issue #583, opened 2020-05-09
Save gym data to tub files


Bleyddyn:
I wanted to be able save data returned by the DonkeyCar simulator into tub files. Rather than having to add a new entry for each item (reward, done, cte, etc) I added support for saving dictionaries to the tub format. I also added support in the drive code and the DonkeyGymEnv class.

My fork has diverged quite a lot so I haven't created a pull request but I can if there's interest.

My changes are in these two commits

https://github.com/Bleyddyn/donkeycar/commit/0964d0fd5b9f30cc893017d28f735bdc0df5724e
 https://github.com/Bleyddyn/donkeycar/commit/552493fd5d41dd45902d0808434fed7908ee229e

tawnkramer:
the default setup from donkey will write out data returned from the sim to tub files.

Bleyddyn:
I think I've checked everywhere in master that might have that code but I can't find it. Any suggestions where to look?

tawnkramer:
If you run the Donkeycar framework with the DONKEY_GYM = True in
myconfig.py then it will use the Tub manager to save data.

Bleyddyn:
No-one else seems to be interested in this so it's kind of a moot point, but I ran this find in both master and dev of autorope/donkeycar:

$ find . -name '*.py' -exec grep -H '\.info' {} \; | grep -v "logging.info" | grep -v "logger.info"
./donkeycar/parts/dgym.py:        self.info = { 'pos' : (0., 0., 0.)}
./donkeycar/parts/dgym.py:            self.frame, _, _, self.info = self.env.step(self.action)

So DonkeyGymEnv is collecting the info dictionary returned by the the simulator, but there's no code that I can find that's doing anything with it, like saving it to a Tub file.

joostaafjes:
This would be nice!! +1

wallarug:
Hey @Bleyddyn  do you still have issues with this in the latest Donkey V4.0 dev branch?   Happy to pick this one up.

Bleyddyn:
@wallarug Sorry I haven't checked out v4.0 yet. If there's any interest you're welcome to check out the commits I linked.

wallarug:
Hey @Bleyddyn ,

I've taken a good look at the changes you have proposed.   I've spent a lot of time the past few weeks working on exactly the same parts to get the 'extra' information back out to Donkey Car.

In particular, I wanted to speed variable out of that `info` array.

https://github.com/autorope/donkeycar/pull/673

[...]

PR #601, opened 2020-07-05
Sim info dict

Bleyddyn:
Add support for saving dict's to Tub files, provided all entries are json compatible. Also add an option and support to drive and gym code to save the simulator's info dictionary (reward and done flags, cte, hit flag, position, etc.). 

See issue #583 .

### DIFF
--- a/donkeycar/parts/datastore.py
+++ b/donkeycar/parts/datastore.py
@@ -222,7 +222,7 @@ class Tub(object):
                 # in case val is a numpy.float32, which json doesn't like
                 json_data[key] = float(val)
 
-            elif typ in ['str', 'float', 'int', 'boolean', 'vector']:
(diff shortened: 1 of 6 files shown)


blown302:
I feel that this part is such a hub we either make is more generic accepting a output config or use the power of oop and override the run method.

Bleyddyn:
@blown302 Which part do you mean? If you mean the drive code, I’ve converted it to a class with multiple methods that can be overridden, in my fork.

blown302:
No the gym part

tawnkramer:
Looks like a useful way to get data into the tubs. I like it.

Bleyddyn:
> No the gym part

That seems like it would be a relatively easy change.

PR #602, opened 2020-07-13
Make model type optional in makemovie and check model type if salient is true

sctse999:
1. Make model type optional in makemovie. If model type is not provided, use config file
2. If salient is true, check whether model type is linear or categorical. If not, exit and warn user.


tawnkramer:
Love it! Thanks

PR #606, opened 2020-07-20
New parts - Stop sign detector

sctse999:
I am adding a stop sign detector part here. This part is enabled by a switch ```STOP_SIGN_DETECTOR``` and leveraged a Google Coral USB Accelerator to do the detection. It will automatically download a pre-trained model hosted by Google if it has not been downloaded yet. It will also overlay a bounding box to the camera feed.

The accuracy is not great but that's what it can do for now.


tawnkramer:
Exciting! Thanks

PR #637, opened 2020-09-25
Tensorflow 2.2 upgrade and adding support for Categorical model

DocGarbanzo:
# Partial contribution to the donkey 4.0.0 release:

## Tensorflow upgrade
- Upgraded to tf 2.2.0 and made corresponding changes in setup, yaml files, etc
- Added automatic tflite conversion if model name ends on `.tflite`
- Unified the different yaml files for Mac, Windows, Ubuntu as these should be all the same

## Categorical model
- Added support for the Categorical model in training where we need one-hot encoded vectors in y in the Y, y generator output

## Misc
- Some pep-8 formatting (line breaks, white-space), removal of unnecessary module inclusions etc, re-arranging of comments, removing unused constructor arguments, etc.
- Small bug fixes in old tub performance script


install/envs/pc.yml
```diff
@@ -0,0 +1,25 @@
...
+- progress
+- moviepy
+- paho-mqtt
+- PrettyTable
+
+- pip:
+  - tensorflow==2.2.0
+  - git+https://github.com/autorope/keras-vis.git
+  - simple-pid
+  - opencv-python-headless
```
DocGarbanzo, line 24: This is a pip dependency from `imgaug` but not a Conda one (strange). Hence I explicitly added it here. I prefer to have `pip install -e .[pc]` not to pip install anything, rather all packages should be defined already here. In general Conda has better package dependency control. Without `opencv-python-headless' the example augmentation script runs, but it is very slow.

donkeycar/templates/train.py
```diff
@@ -67,6 +68,7 @@ def __getitem__(self, index):
         throttles = []
 
         is_inferred = type(self.keras_model) is KerasInferred
+        is_categorical = type(self.keras_model) is KerasCategorical
```
DocGarbanzo, line 71: Actually I don't like this `if/else` stuff here, I would prefer to move this data processing that is specific to the `KerasPilot` into a virtual method there. Then each model could decide how to massage the data by itself, that would be much more OO. @tikurahul - let me know if you agree.

install/envs/pc.yml
tikurahul: Good to know. I don't use `conda` myself., so I am not sure about these YAML files. Thanks for fixing this.

donkeycar/templates/train.py
tikurahul: Yes, in general the  model is coupled with the input format. So that makes a lot of sense to me. 

However, lets try and figure out a good API first - because as we get to the more exotic models, I have a feeling that we might have to do a lot more.

donkeycar/parts/keras.py
```diff
@@ -143,20 +130,20 @@ def run(self, img_arr):
...
-    The KerasLinear pilot uses one neuron to output a continous value via the 
+    """
+    The KerasLinear pilot uses one neuron to output a continous value via the
     Keras Dense layer with linear activation. One each for steering and throttle.
     The output is not bounded.
-    '''
-    def __init__(self, num_outputs=2, input_shape=(120, 160, 3), roi_crop=(0, 0), *args, **kwargs):
-        super(KerasLinear, self).__init__(*args, **kwargs)
+    """
+    def __init__(self, num_outputs=2, input_shape=(120, 160, 3), roi_crop=(0, 0)):
```
tikurahul, line 145: Lets get rid of `roi_crop` because that won't compose well with more complicated transforms made possible by the new datastore. 

We should also document this new workflow.

donkeycar/utils.py
```diff
@@ -407,27 +407,42 @@ def get_model_by_type(model_type, cfg):
     if model_type == "tflite_linear":
         kl = TFLitePilot()
     elif model_type == "localizer" or cfg.TRAIN_LOCALIZER:
-        kl = KerasLocalizer(num_locations=cfg.NUM_LOCATIONS, input_shape=input_shape)
```
tikurahul, line 410: Lets support the basic `linear` and `categorical` models out of the box in `get_model_by_type`.

Other models (RNN et. al.) require us to do more with the input data pipeline. We can do it as a stretch goal. Given we are not ready for this yet, I think we can leave those models in `keras.py` but not support them in `get_model_by_type` for now.

donkeycar/templates/train.py
DocGarbanzo: Sure, we can keep this for now, and I'll make a PR with a design proposal.

donkeycar/parts/keras.py
DocGarbanzo: I'll remove the `roi_crop` and update the PR.

donkeycar/utils.py
DocGarbanzo: Makes sense, let me update this.

DocGarbanzo:
Accidentally closed the PR. Have now added the missing bits:
* Only support core models in get_model_by_type
* Factoring out common code in keras.py
* Removed roi_crop from model interface

donkeycar/templates/train.py
```diff
@@ -177,6 +186,10 @@ def main():
     tubs = args['--tubs']
     model = args['--model']
     model_type = args['--type']
+    model_name, model_ext = os.path.splitext(model)
+    is_tflite = model_ext == '.tflite'
+    if is_tflite:
+        model = model_name + '.h5'
```
tikurahul, line 192: Prefer string interpolation here. So something like `f'{model_name}.h5'`. Works on Python 3.6 +

tikurahul (approved):

donkeycar/templates/train.py
DocGarbanzo: I'll have a look at this - I need to merge in some other small changes for RPi setup in a follow up PR.

PR #639, opened 2020-09-30
Alternative yaml based car configuration

DocGarbanzo:
# Add support for yaml based car app
Minimal implementation of an alternative to the template based car app, manage.py. This has the following changes:
* Adding a `Part` base class for donkey parts
* Adding a `PartFactory` so parts can be created from dictionaries
* Adding a `Builder` which reads a yaml file and creates a `Vehicle`, the yaml file contains:
  - some settings like car frequency, etc
  - a list of parts with:
    - the part's constructor arguments
    - the part's inputs and outputs
    - if the part is threaded
* Added an Open CV camera
* Adding an example car yaml file containing the open cv camera and the new tub
* Adding a new command `donkey drive --car_file` to allow creating and driving the car off the file

## Missing bits
This needs to be done in order to create a valid alternative for the real-world car:
* Making all other donkey parts `Part`
* Adding more documentation to the code
* Updating the docs
* Translating the current car app (at least in  a smaller version) into a standard car template file
* Moving the tub name from the `TubWriter` part recipe, this should be auto-generated as before
* Pull other classes out of the current car app so they can be made `Part` and be added into the vehicle's data flow


DocGarbanzo:
Withdrawn after discussions with @tikurahul.

PR #642, opened 2020-10-04
RPi config, basic car template and minor other fixes.

DocGarbanzo:
# Dev-4.0 release items
* Creating a minimal car template, by extending the current web template to allow web or js control and renaming it accordingly ('basic')
* Changing the default template to 'basic'
* Reverting back changes on categorical model to produce the exact same model as before - there is a small tweak in the 4th can layer to use strides=2 making the model significantly better and smaller. We want to keep this behaviour for now.
* Changing image handling in the car app to work on uint8 images, so `KerasPilot.run()` now expects a `uint8`, normalises the image and delegates to virtual method inference() which is implemented in the derived classes. Changed code accordingly where `KerasPilot.run()` is called. Derived `TFLitePilot` from `KerasPilot` 
* Fixing yaml files to include Conda-forge
* Fixed `makemovie` to also work with `tflite` models


docs/guide/robot_sbc/setup_raspberry_pi.md
```diff
@@ -219,8 +219,8 @@ git clone https://github.com/autorope/donkeycar
 cd donkeycar
 git checkout master
 pip install -e .[pi]
-pip install tensorflow==1.13.1
 pip install numpy --upgrade
+pip install https://github.com/lhelontra/tensorflow-on-arm/releases/download/v2.2.0/tensorflow-2.2.0-cp37-none-linux_armv7l.whl
```
tikurahul, line 223: We should not be using this. There are precompiled wheels on Pi Wheels that should be picked up. Google also makes precompiled binaries available. 

Switch to something like `tensorflow==2.3.0`.

donkeycar/parts/image.py
```diff
@@ -104,3 +108,11 @@ def run(self, img_arr):
 
     def shutdown(self):
         pass
+
+
+class ImgPrecondition:
```
tikurahul, line 113: I want to get rid of this and fold this into either the camera or a new threaded part.

donkeycar/templates/basic.py
```diff
@@ -0,0 +1,137 @@
...
+    run_threaded method depending on the constructor flag `threaded`. All
+    parts are updated one after another at the framerate given in
+    cfg.DRIVE_LOOP_HZ assuming each part finishes processing in a timely
+    manner. Parts may have named outputs and inputs. The framework handles
+    passing named outputs to parts requesting the same named input.
+    """
+
+    car = dk.vehicle.Vehicle()
+    # add camera
+    cam = PiCamera(image_w=cfg.IMAGE_W, image_h=cfg.IMAGE_H,
```
tikurahul, line 60: Should we checking for `cfg.camera` to see which one to use?

tikurahul (changes requested):
I am wondering if we should keep `ImgPrecondition` part in the basic setup.

docs/guide/robot_sbc/setup_raspberry_pi.md
DocGarbanzo: Let me check this, I couldn't find armv7 on google for tf 2.2.0, but I'll try 2.2.3

donkeycar/parts/image.py
DocGarbanzo: I can make this part threaded, that's easy, but it's very fast, so won't make a big difference.

I don't think it should go into the camera which returns uint8 native images. These also go into the tub, etc, so there we don't want to convert back again. The place where normalised floats are required are in the pilot. I can put this normalisation/float conversion into a pilot base class and have that virtually dispatch `run` into derived parts.

donkeycar/templates/basic.py
DocGarbanzo: Yes, let me add that.

docs/guide/robot_sbc/setup_raspberry_pi.md
tikurahul: https://www.piwheels.org/project/tensorflow  

Shows `2.3.1` as the most recent version of TF.

DocGarbanzo: Through pypi we only get:
```bash
pip install tensorflow==
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
ERROR: Could not find a version that satisfies the requirement tensorflow== (from versions: 0.11.0, 1.12.0, 1.13.1, 1.14.0)
```
On the TF page (https://www.tensorflow.org/install/pip) google only have a binary for tf==2.3.1 on py3.5 for armv7 which doesn't install on python 3.7: 
```bash
pip install https://storage.googleapis.com/tensorflow/raspberrypi/tensorflow-2.3.0-cp35-none-linux_armv7l.whl
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
ERROR: tensorflow-2.3.0-cp35-none-linux_armv7l.whl is not a supported wheel on this platform.
```
The other option would be to compile it ourselves from https://www.tensorflow.org/install/source_rpi#python-3.7, I can give it a shot.

donkeycar/parts/image.py
tikurahul: I meant the `normalize_and _crop` won't actually be doing much given we are getting rid of `ROI_CROP_*`.  The only thing that its doing is converting an uin8 image to a normalized image.

DocGarbanzo: Yes correct, and we need both representations, the uint8 for the tub and the normalised float for the pilot. I think this should go into the pilot, that makes the most sense to me.

docs/guide/robot_sbc/setup_raspberry_pi.md
tikurahul: That's because the `tf` version is still set to `1.x` in `setup.py`.

DocGarbanzo: There is no `tf` in `setup.py`.

tikurahul: Sorry, I sent you a new change: https://github.com/autorope/donkeycar/pull/643

tikurahul (approved):

tikurahul (changes requested):
Please rebase your changes.

DocGarbanzo:
Shall I rebase this against `dev` instead of `dev-4.x`?

tikurahul:
I think you will have to restage your commits because you seem to be bringing in all the commits I squashed. 

I would `reset --soft` and stash your changes. Reset `HEAD` to `dev`  (`--hard`). 

Then stage your changes and submit it as a single commit.

DocGarbanzo:
I have re-based to dev now. I think the list of commits is misleading (not sure why). If you look at the diffs then it's all as expected. We should squash all commits into one when merging (https://cloudfour.com/thinks/squashing-your-pull-requests/).

tikurahul:
This is still incorrect. You are rewriting git history. Lets discuss offline, and I can help you with this.

PR #643, opened 2020-10-04
Minor fixes to setup.py and remove ROI_CROP completely.

tikurahul:


setup.py
```diff
@@ -68,34 +67,31 @@ def package_files(directory, strip_leading):
               'pytest-cov',
               'codecov'
           ],
+          'ci': ['codecov'],
```
DocGarbanzo, line 70: This in already in 'pi' and for the PC installation part of the Conda env, so we don't need this here any longer.

setup.py
```diff
@@ -68,34 +67,31 @@ def package_files(directory, strip_leading):
...
+          'tf': ['tensorflow>=2.2.0'],
```
DocGarbanzo, line 71: `tf` is already in the conda envs, so will be installed on the PC already. We need this bit rather in the `[pi]` section, but there it won't be working as there is no `tf>1.15` in PyPI.

tikurahul: PyPy is only one of places that we look for. PiWheels has `2.3.1` in the package repo. This is the correct thing to do to support Pip and regular `setup.py` methods to install the package. Not everyone uses `conda` / `miniconda`.

setup.py
tikurahul: I don't want to rely on `conda` too much. I want to support traditional `pip` installs as well.

setup.py
DocGarbanzo: Actually if you run
```
pip install tensorflow>=2.2.0
```
on RPi you get version 1.14.0 installed.

tikurahul: Hmm.. That is super strange.. Try `tensorflow=2.2.0` ?

DocGarbanzo: I think we should be installing via conda on the PC, this is much easier and avoids any problems with incompatible packages, etc. This was already in the docs. On Pi/Nano, etc, where there is no conda, we need to use PyPi. Let me check PiWheels...

tikurahul: `conda` is too magical. It really messes with virtual environments when you are dealing with cuDNN and other shared libraries. It _must_ be optional.

tikurahul: Are you running `pip install tensorflow>=2.2.0` on a PC on a Pi or someplace else ? Are you sure its looking for piwheels ?

DocGarbanzo: Actually I already did this, PiWheels is part of the standard place to look for, I already pasted that in the other PR:
```
(env4) pi@donkeypi:~/projects/donkeycar $ pip install tensorflow==
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
ERROR: Could not find a version that satisfies the requirement tensorflow== (from versions: 0.11.0, 1.12.0, 1.13.1, 1.14.0)
```

tikurahul: You can also try `pip install tensorflow == 2.3.1` on a Pi.

setup.py
DocGarbanzo: Yes, this will work too, as in setup.py we more or less duplicate what we have in the conda yaml files. So if you build your env with conda and then run `pip install -e .` it will basically do nothing b/c everything is there and b/c dependencies are set up correctly, you also don't need to upgrade `numpy` etc. 
However, if you only install with `pip` using setup, your installation will also work, but I thought this is only really relevant for PI / Nano where we don't have conda. But it should work on every other box, too, I agree. For this we would need `tf` in here too, the way you added it. Hopefully it doesn't mess up the Conda env, I can check that.

DocGarbanzo: Yeah shouldn't be a problem. Sorry put this whole shebang one line too high...

setup.py
DocGarbanzo: This doesn't work as the only packages on armv7 in pypi and pinwheels are `0.11.0, 1.12.0, 1.13.1, 1.14.0`.

DocGarbanzo: Your change here is fine. It won't resolve the install on PI, but it won't harm anything if you have started w/ a Conda env and run it on top, so looks safe, and makes sense if you want to run w/o conda, good point.

DocGarbanzo (approved):
Looks good, thanks for explaining.

PR #647, opened 2020-10-07
Cleanup MacOS install

cfox570:
Added channel conda-forge so Conda could install  - paho-mqtt, prettytable, imgaug, moviepy.

Updated mac.yml to match ubuntu.yml as much as possible.

Changed tensorflow command to install latest version.

Removed redundant install setting in setup.py (pc: imaug, progress)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,8 +62,6 @@ setup(name='donkeycar',
           ],
           'pc': [
               'matplotlib',
-              'imgaug',
-              'progress',
           ],
           'dev': [
               'pytest',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -62,8 +62,6 @@ def package_files(directory, strip_leading):
           ],
           'pc': [
               'matplotlib',
-              'imgaug',
```
tikurahul, line 65: Why did you remove these from `setup.py` ? 

I think these dependencies are still relevant.

tikurahul:
This is no longer necessary because we landed a fix.

PR #648, opened 2020-10-08
Improvements to the car app and handling of KerasPilot parts:

DocGarbanzo:
# Minimal car app and image data flow improvements for dk 4.0
* Created a simpler webserver or joystick car app by modifying and renaming the basic_web into basic template and switched this on as default. Also renamed the target from 'manage.py' to 'drive.py' because there is only driving and no training included
* Simplified the handling of uint8 and float32 numpy image arrays. KerasPilot.run() now expects uint8 data, transforms into float32 and delegates to its children. The corresponding rescaling step is gone from the car app
* smaller updates/fixes to environment and config files
* added support for tflite pilots in makemovie and moved teilte pilot under KerasPilot


donkeycar/parts/keras.py
```diff
@@ -55,7 +58,19 @@ def set_optimizer(self, optimizer_type, rate, decay):
             self.model.optimizer = keras.optimizers.RMSprop(lr=rate, decay=decay)
         else:
             raise Exception("unknown optimizer type: %s" % optimizer_type)
-    
+
+    def get_input_shape(self):
+        assert self.model is not None, "Need to load model first"
+        return self.model.inputs[0].shape
+
+    def run(self, img_arr, other_arr=None):
```
tikurahul, line 66: Nit: add documentation for `other_arr`. Its not obvious what it is actually used for.

donkeycar/templates/cfg_complete.py
```diff
@@ -94,7 +94,7 @@
 #The DEFAULT_MODEL_TYPE will choose which model will be created at training time. This chooses
 #between different neural network designs. You can override this setting by passing the command
 #line parameter --type to the python manage.py train and drive commands.
-DEFAULT_MODEL_TYPE = 'linear'   # (linear|categorical|tflite_liner|tensorrt_linear)
+DEFAULT_MODEL_TYPE = 'linear'   #(linear|categorical|rnn|imu|behavior|3d|localizer|latent)
 BATCH_SIZE = 128                #how many records to use when doing one pass of gradient decent. Use a smaller number if your gpu is running out of memory.
 TRAIN_TEST_SPLIT = 0.8          #what percent of records to use for training. the remaining used for validation.
```
tikurahul, line 99: We don't have the any of the models `rnn|imu|behavior|3d|localizer|latent`

Set  it to `linear|categorical|tflite_liner|tensorrt_linear`

tikurahul (changes requested):
Looks great. 

Minor comment, and we can land this.

donkeycar/parts/keras.py
DocGarbanzo: Yep - makes sense, added full doc string.

donkeycar/templates/cfg_complete.py
DocGarbanzo: Yep, merge error on my side.

tikurahul:
I am squashing and merging this, but when you make changes on a CL you should `git commit --amend` to not create a new commit. 

We should strive for a single commit that has all the changes. No more rewriting of history or spurious commits. Lets keep the history clean.

DocGarbanzo:
Yep agreed, just figured out how to do this.

PR #649, opened 2020-10-10
Add auto-generated tub path to basic car template

DocGarbanzo:
Add auto-generated tub path to basic car template, otherwise the vari…able cfg.DATA_PATH will be used as is for the tub path.


tikurahul (commented):
I don't know if we actually need this. I personally do not appreciate a new tub being created every time we call `drive`. Lets make this configurable ?

tikurahul (approved):
LGTM

PR #651, opened 2020-10-11
Add testing for training:

DocGarbanzo:
* added example tub data as tar.gz in tests with 1000 records
* check validation data size in train
* created new test_train.py
* added support for read-only tub, required to create r/o tub in tmp dir
* changed loss in categorical model to equally weight throttle and steering
* changed mod on scripts to u+x


tikurahul (approved):
Awesome :clap:

PR #659, opened 2020-10-17
Refactoring `KerasLatent` into auto-encoder-controller and other tub improvements

DocGarbanzo:
## Refactoring of `KerasLatent` model
* decomposing the latent model into encoder, decoder and controller - the first two parts are the auto-encoder, the controller takes the latent vector as input and outputs steering and throttle
* the encoder cnn now matches the linear model so the models are comparable
* added support to build the latent model from a trained encoder, that can be loaded from a trained latent model
* integration of latent model into training

## Additional features for new tub
* deleting records by values - useful to suppress certain records in training
* undeleting records - useful to temporarily suppress records and then set them back
* added test of new features

## Additional features for training
* allow suppressing of shuffling - mostly useful to keep tests more repeatable
* train script now shows train time again
* changed train test to relative convergence

## Additional features in donkey
* donkey tubplot now accepts an optional start parameter too

## Additional features in conversion script
* add flag to put legacy tub path into new tub, so records can be identified w/o having to use multiple new tubs

### DIFF
--- a/donkeycar/management/base.py
+++ b/donkeycar/management/base.py
@@ -1,10 +1,7 @@
 
 import argparse
-import os
 import shutil
-import socket
 import stat
-import sys
 from socket import *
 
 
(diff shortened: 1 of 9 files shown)


DocGarbanzo:
Needs rebasing into newest standard.

PR #660, opened 2020-10-17
Be explicit about newline characters

tikurahul:


DocGarbanzo:
I think there is os.linesep instead of \n or \n\r if we want to be system agnostic to new lines.

tikurahul:
`os.linesep` would mean that it could be problematic to collect data on Linux and train on Windows. This way irrespective of the platform we are explicit about the newline character and we get portability.

DocGarbanzo:
Ok you just hard-replaced the file endings now with windows format. Usually if you transfer text files to windows you would just run `unix2dos` on them. Now under linux/mac I believe the text files will look odd in a standard text editor (emacs, nano, less, etc) because they will have `^R` in them, or not?

tikurahul:
No. I use the same for both. It will look correct on editors everywhere because editors already support different types of line endings.

Issue #652, opened 2020-10-17
myconfig - track name updates error


wallarug:
```
ERROR:gym_donkeycar.core.client:Exception:Scene name mountain_track not in scene list ['generated_road', 'warehouse', 'sparkfun_avc', 'generated_track', 'roboracingleague_1']
```

TODO:  Remove from myconfig as a recommended option.

wallarug:
Fixed in #669  by using a simulator.py template now.  Simulator should be removed from all other configs to avoid incompatibility.

PR #666, opened 2020-10-19
USYD Brakes

wallarug:
**New functionality for Simulator**

- Brake Communication between sim and donkey vehicle


wallarug:
This commit contains the following changes.

- Added a `brake` variable to a number of functions in a number of files.
  - donkeycar/parts/dgym.py `run_threaded`
  - donkeycar/parts/launch.py `run`
  - donkeycar/parts/web_controller/web.py `run return value`

- Reflect these function changes in the following templates:
  - complete.py
  - basic.py

All other changes to tubs and datastore should not be of any notable change - they are coming from changes made by rahul that were merged in during the brake changes.

wallarug:
Also has fix for #652 by changing the track in the comment to something that works.

tikurahul:
@wallarug you are reordering commits here. 

Please isolate your changes and apply them on top of commits and not interleave them with the existing git history. Please use `git pull --rebase` instead of a regular `pull` because pull is a shorthand for a fetch and merge. Ideally we should strive for never having to merge and only have fast forwards.

tikurahul:
Also please squash your changes into a single commit by using `git rebase -i`. That makes your changes easier to isolate if we need to.

PR #669, opened 2020-10-20
USYD Brakes - Simulator

wallarug:
added in brake functionality for simulator only

**Updates**
- changed `dgym.py` file
- added new simulator.py template and configuration to keep other ones untouched.
- retains backwards compatibility with complete.py and other templates that have the simulator.
- doesn't interfere with training / data collection.

**Background**

This is to allow custom object and sign detection to stop the car effectively in the simulator (see _StopSignDetector_).   At the moment, the only way to control the car is via throttle.  To stop the car in the simulator (with it's low friction) you have to send very high reverse throttle values.  This leads to the car bouncing / rolling and not really being able to come to a complete stop.  With this update, brakes can be deployed and allow the algorithm to stop the car.

### DIFF
--- a/donkeycar/parts/dgym.py
+++ b/donkeycar/parts/dgym.py
@@ -24,7 +24,7 @@ class DonkeyGymEnv(object):
         conf['guid'] = 0
         self.env = gym.make(env_name, conf=conf)
         self.frame = self.env.reset()
-        self.action = [0.0, 0.0]
+        self.action = [0.0, 0.0, 0.0]
         self.running = True
         self.info = { 'pos' : (0., 0., 0.)}
         self.delay = float(delay)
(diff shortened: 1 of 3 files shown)


wallarug:
model_loss1 test failed - I don't know why?  That's the only one.

sctse999:
I see that there is a separated config.py template for simulator. Suggest to make it to minimal since it is a dedicated config for simulator. Same goes for the simulator template (e.g. remove the LED part).

I have no idea why the test fails too.

wallarug:
Hey @sctse999  I left it in there because they are working on some fancy LED work in the simulator as well.

I also was not sure how much more I could take out of the simulator template.  I think the LED bits are the only ones that 'extra'.

wallarug:
@sctse999  could you message me on Discord or make the suggestion on how to improve it?   

I would like to see this put in ASAP as I have 50 students dependant on this PR.

PR #670, opened 2020-10-20
MakeMovie Utility Update - option for turning off line drawing

wallarug:
add in --user flag for MakeMovie so that you can select if you want to draw the line on the video or not.  default: true

**Details**
Command works exactly the same except now you can add a --user flag to advise the donkey makemovie command that you don't want the user input shown.  The reason for this is that if you draw the green line for the user input from collected data, you cannot use the video for OpenCV or object classification because it throws out the algorithms (a random green line goes across everything).  So the solution is to still compile the video - but leave off the green line.

**Usage**

```bash
donkey makemovie --tub=data/latest --scale=1 --user False --out latest.mp4
```
... will output a video without the green line.

```bash
donkey makemovie --tub=data/latest --scale=1 --out latest.mp4
```
... will output a video with the green line.

### DIFF
--- a/donkeycar/management/base.py
+++ b/donkeycar/management/base.py
@@ -252,7 +252,7 @@ class MakeMovieShell(BaseCommand):
         parser.add_argument('--start', type=int, default=0, help='first frame to process')
         parser.add_argument('--end', type=int, default=-1, help='last frame to process')
         parser.add_argument('--scale', type=int, default=2, help='make image frame output larger by X mult')
-        parser.add_argument('--user', default=True, action='store_false', help='show user input on the video')
+        parser.add_argument('--draw-user-input', default=True, action='store_false', help='show user input on the video')
         parsed_args = parser.parse_args(args)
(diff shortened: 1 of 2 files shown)


donkeycar/management/base.py
```diff
@@ -252,6 +252,7 @@ def parse_args(self, args):
         parser.add_argument('--start', type=int, default=0, help='first frame to process')
         parser.add_argument('--end', type=int, default=-1, help='last frame to process')
         parser.add_argument('--scale', type=int, default=2, help='make image frame output larger by X mult')
+        parser.add_argument('--user', default=True, action='store_false', help='show user input on the video')
```
tikurahul, line 255: Call this option `--draw-user-input`. `--user` is a bit too succinct.

tikurahul (approved):

donkeycar/management/base.py
wallarug: Done!

PR #675, opened 2020-10-27
Allow overriding WEB_CONTROL_PORT from env variable

sctse999:
To test this:
```
export WEB_CONTROL_PORT=12345
python manage.py drive
```
Visit `http://localhost:12345/drive` instead of `http://localhost:8887/drive`
Web Controller should start listening on port 12345. Port 8887 should not work after you specify the port in the env variable.

### DIFF
--- a/donkeycar/templates/cfg_complete.py
+++ b/donkeycar/templates/cfg_complete.py
@@ -137,7 +137,7 @@ FREEZE_LAYERS = False               #default False will allow all layers to be m
 NUM_LAST_LAYERS_TO_TRAIN = 7        #when freezing layers, how many layers from the last should be allowed to train?
 
 #WEB CONTROL
-WEB_CONTROL_PORT = 8887             # which port to listen on when making a web controller
+WEB_CONTROL_PORT = int(os.getenv("WEB_CONTROL_PORT", 8887))  # which port to listen on when making a web controller
 WEB_INIT_MODE = "user"              # which control mode to start in. one of user|local_angle|local. Setting local will start in ai mode.
 
 #JOYSTICK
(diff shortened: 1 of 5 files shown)


donkeycar/templates/cfg_complete.py
```diff
@@ -137,7 +137,7 @@
 NUM_LAST_LAYERS_TO_TRAIN = 7        #when freezing layers, how many layers from the last should be allowed to train?
 
 #WEB CONTROL
-WEB_CONTROL_PORT = 8887             # which port to listen on when making a web controller
+WEB_CONTROL_PORT = int(os.getenv("WEB_CONTROL_PORT", 8887))  # which port to listen on when making a web controller
```
DocGarbanzo, line 140: Can you add that into the `cfg_basic.py` too?

donkeycar/tests/test_web_controller.py
```diff
@@ -1,19 +1,33 @@
...
     json_result = json.dumps(result)
     d = json.loads(json_result)
+    
+    assert server.port == 8887
+    
     assert d is not None
     assert int(d[0]) == 0
 
 
+def test_web_control_user_defined_port():
```
DocGarbanzo, line 26: Great that you added a test!

DocGarbanzo (commented):
Looks good, only one minor comment.

donkeycar/templates/cfg_complete.py
sctse999: Hey this is master and there is no `cfg_basic.py`. I will port this to `cfg_basic.py` in dev in another PR.

DocGarbanzo: You're right. Looks good.

DocGarbanzo (approved):
LGTM

PR #677, opened 2020-11-01
Bump version to 4.0.1

DocGarbanzo:


DocGarbanzo:
Travis only fails for one train test which we are taking out of travis anyway (the test if fine but the convergence non-deterministic)

PR #688, opened 2020-11-28
Add mqtt telemetry

cloud-rocket:

PR #689, opened 2020-11-29
Add performance monitor

cloud-rocket:
Performance monitor allows monitoring CPU / Memory and real Donkey execution frequency. I noticed that my own execution loop is considerably lower than 20Hz on Jetson Nano (~8Hz) and decided to measure it. On PC  simulator everything runs at 20hz as required.

This pull request includes performance monitor together with MQTT telemetry from another pull request (https://github.com/autorope/donkeycar/pull/688).

### DIFF
--- a/donkeycar/parts/perfmon.py
+++ b/donkeycar/parts/perfmon.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Performance monitor for analyzing real-time CPU/mem/execution frequency
+
+author: @miro (Meir Tseitlin) 2020
+
+Note:
+"""
+import time
+import psutil
+
+
+class PerfMonitor:
+
+    def __init__(self, cfg):
+
+        self.STATS_BUFFER_SIZE = 10
+        self._calc_buffer = [cfg.DRIVE_LOOP_HZ for i in range(self.STATS_BUFFER_SIZE)]
+        self._runs_counter = 0
+        self._last_calc_time = time.time()
+        self._on = True
+        self._update_metrics()
+        print("Performance monitor activated.")
+
+    def _update_metrics(self):
+        self._mem_percent = psutil.virtual_memory().percent
+        self._cpu_percent = psutil.cpu_percent()
+
+    def update(self):
+        while self._on:
+            self._update_metrics()
+            time.sleep(2)
+
+    def shutdown(self):
+        # indicate that the thread should be stopped
+        self._on = False
+        print('Stopping Perf Monitor')
+        time.sleep(.2)
+
+    def run_threaded(self):
+
+        # Calc real frequency
+        curr_time = time.time()
+        if curr_time - self._last_calc_time > 1:
+            self._calc_buffer[int(curr_time) % self.STATS_BUFFER_SIZE] = self._runs_counter
+            self._runs_counter = 0
+            self._last_calc_time = curr_time
+
+        self._runs_counter += 1
+
+        vehicle_frequency = float(sum(self._calc_buffer)) / self.STATS_BUFFER_SIZE
+
+        return self._cpu_percent, self._mem_percent, vehicle_frequency
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -580,6 +580,14 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type
         inputs += ['pilot/angle', 'pilot/throttle']
         types += ['float', 'float']
 
+    if cfg.HAVE_PERFMON:
+        from donkeycar.parts.perfmon import PerfMonitor
+        mon = PerfMonitor(cfg)
+        perfmon_outputs = ['perf/cpu', 'perf/mem', 'perf/freq']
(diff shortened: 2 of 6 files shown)


donkeycar/parts/perfmon.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, cfg):
+
+        self.STATS_BUFFER_SIZE = 10
+        self._calc_buffer = [cfg.DRIVE_LOOP_HZ for i in range(self.STATS_BUFFER_SIZE)]
+        self._runs_counter = 0
+        self._last_calc_time = time.time()
+        print("Performance monitor activated.")
+
+    def run(self):
```
tikurahul, line 24: Let's make this a threaded part. Look at the camera or the OLED part for more information.

tikurahul (changes requested):

donkeycar/parts/perfmon.py
cloud-rocket: Ok. Will do...

cloud-rocket: Moved heavy functions into thread

DocGarbanzo:
This change seems to contain your change for the MQTT telemetry too. I suggest we tackle that MQTT change first and then you rebase on the updated `dev` branch with that second change here.

cloud-rocket:
@DocGarbanzo , @tikurahul 
rebased with recent changes, can pls review again?

Tnx

setup.py
```diff
@@ -48,6 +48,7 @@ def package_files(directory, strip_leading):
           "simple_pid",
           'progress',
           'typing_extensions',
+          'psutil'
```
DocGarbanzo, line 51: Because this only runs on the car, it might go into the `pi` section.

DocGarbanzo:
Sorry @cloud-rocket - we were a bit tied up w/ the 4.1 release. It looks good to me only one minor comment. @tikurahul ?

tikurahul:
@cloud-rocket I don't think you rebase correctly. This change includes some of the changes that I authored again which tells me we are rewriting git history. (Pipeline, datastore etc)

Please create a PR that only includes _your_ new changes.

cloud-rocket:
@tikurahul ,

Sorry for a delay, everything is cleaned and rebased to the last dev version. Please review

cloud-rocket:
> This change seems to contain your change for the MQTT telemetry too. I suggest we tackle that MQTT change first and then you rebase on the updated `dev` branch with that second change here.

Can pls review again?
Tnx

donkeycar/templates/complete.py
```diff
@@ -580,6 +586,10 @@ def run(self, mode, recording):
         inputs += ['pilot/angle', 'pilot/throttle']
         types += ['float', 'float']
 
+    if cfg.HAVE_PERFMON:
```
tikurahul, line 589: Why not do this on the if-check on L#307 ?

tikurahul (approved):
One minor change, otherwise things look great.

donkeycar/templates/complete.py
cloud-rocket: Because I am extending the `inputs` and `types` arrays which are initialized after L#562, those arrays are used fby TubHandler and MqttTelemetry. I can maybe move the initialization of Perfmon from lines 307 to here also if you think it'll make a difference.....

DocGarbanzo: I agree, it would be cleaner to have this in a single block here.

setup.py
DocGarbanzo: Can you move this into the `'pi'` section?

DocGarbanzo:
@cloud-rocket - do you want to fix this last small bit and then we can merge it?

cloud-rocket:
Yes, will do in the next few days

cloud-rocket:
@DocGarbanzo, @tikurahul, I fixed the `donkeycar/templates/complete.py` and also reconfigured MqttTelemetry to be threaded by default

DocGarbanzo:
@cloud-rocket - this looks all fine to me, except I made a comment in `setup.py` - I don't think we need this package on the host, but only on pi and nano.

setup.py
cloud-rocket: I don't understand why should it be limited to `pi` or `nano`? What if I want. to verify performance running in simulation?
Sorry for a delayed response.

cloud-rocket:
> @cloud-rocket - this looks all fine to me, except I made a comment in `setup.py` - I don't think we need this package on the host, but only on pi and nano.

Sorry for a delay - see my comment inline. Tnx

setup.py
DocGarbanzo: Fair point - then we should add it to the conda.yml files too.

cloud-rocket: @DocGarbanzo - should I add it to `envs/mac.yml` and `envs/ubuntu.yml` ?

DocGarbanzo: Yeah, I think the second one got renamed to 'pc.yml' in the mean time.

cloud-rocket: fixed

DocGarbanzo (approved):
Thank you that looks fine.

PR #691, opened 2020-11-30
Switch to using memory-mapped files when reading.

tikurahul:
* This makes reading a 100x faster.

### DIFF
--- a/donkeycar/parts/datastore_v2.py
+++ b/donkeycar/parts/datastore_v2.py
@@ -1,4 +1,5 @@
 import json
+import mmap
 import os
 import time
 from pathlib import Path
@@ -14,11 +15,14 @@ class Seekable(object):
     This reader maintains an index of line lengths, so seeking a line is a O(1) operation.
     '''
 
-    def __init__(self, file, method='a+', line_lengths=list()):
-        self.method = method
+    def __init__(self, file, read_only=False, line_lengths=list()):
         self.line_lengths = list()
         self.cumulative_lengths = list()
+        self.method = 'r' if read_only else 'a+'
         self.file = open(file, self.method, newline=NEWLINE)
+        if self.method == 'r':
+            # If file is read only improve performance by memory mappping the file.
+            self.file = mmap.mmap(self.file.fileno(), length=0, access=mmap.ACCESS_READ)
         self.total_length = 0
         if len(line_lengths) <= 0:
             self._read_contents()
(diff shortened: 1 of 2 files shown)


donkeycar/parts/datastore_v2.py
```diff
@@ -46,6 +50,9 @@ def __enter__(self):
         return self
 
     def writeline(self, contents):
+        if self.method == 'r':
+            raise Exception(f'Seekable {self.file} is read-only.')
```
DocGarbanzo, line 54: Only minor comment - I try to avoid throwing base `Exceptions` but rather use more specific ones, here probably `RuntimeError`

donkeycar/parts/datastore_v2.py
```diff
@@ -14,11 +15,14 @@ class Seekable(object):
     This reader maintains an index of line lengths, so seeking a line is a O(1) operation.
     '''
 
-    def __init__(self, file, method='a+', line_lengths=list()):
-        self.method = method
+    def __init__(self, file, read_only=False, line_lengths=list()):
         self.line_lengths = list()
         self.cumulative_lengths = list()
+        self.method = 'r' if read_only else 'a+'
         self.file = open(file, self.method, newline=NEWLINE)
```
DocGarbanzo, line 22: We should probably have a `with` here...

donkeycar/parts/datastore_v2.py
```diff
@@ -14,11 +15,14 @@ class Seekable(object):
...
+        if self.method == 'r':
+            # If file is read only improve performance by memory mappping the file.
+            self.file = mmap.mmap(self.file.fileno(), length=0, access=mmap.ACCESS_READ)
```
DocGarbanzo, line 25: ... and here, too.

DocGarbanzo (commented):
Looks good to me only minor comments.

donkeycar/parts/datastore_v2.py
tikurahul: We cannot use a context manager here, given the  explicit `close()` semantics.

donkeycar/parts/datastore_v2.py
tikurahul: Same as above.

donkeycar/parts/datastore_v2.py
tikurahul: Done.

donkeycar/parts/datastore_v2.py
DocGarbanzo: Ok - I missed that.

PR #693, opened 2020-12-03
A configurable training pipeline.

tikurahul:
* We start off with a list of shallow records, because it makes
  things easier with a train_test_split.
* The rest of the pipeline is lazy.
* The `pipeline` module is typed.

### DIFF
--- a/donkeycar/pipeline/sequence.py
+++ b/donkeycar/pipeline/sequence.py
@@ -1,0 +1,217 @@
+from typing import Any, Callable, Generic, Iterable, Iterator, List, Sized, Tuple, TypeVar
+
+import math
+import numpy as np
+from donkeycar.pipeline.types import TubRecord
+
+# Note: Be careful when re-using `TypeVar`s.
+# If you are not-consistent mypy gets easily confused.
+
+R = TypeVar('R', covariant=True)
+X = TypeVar('X', covariant=True)
+Y = TypeVar('Y', covariant=True)
+XOut = TypeVar('XOut', covariant=True)
+YOut = TypeVar('YOut', covariant=True)
+
+
+class SizedIterator(Generic[X], Iterator[X], Sized):
+    def __init__(self) -> None:
+        # This is a protocol type without explicitly using a `Protocol`
+        # Using `Protocol` requires Python 3.7
+        pass
+
+    def batched(self, batch_size: int) -> Iterator[List[X]]:
+        # Produces a batch of results
+        # Ideally we should be able to return a `SizedIterator` but `mypy`
+        # won't resolve this kind of recursive type
+        raise NotImplemented
+
+
+class TubSeqIterator(SizedIterator[TubRecord]):
+    def __init__(self, records: List[TubRecord]) -> None:
+        self.records = records or list()
+        self.current_index = 0
+
+    def __len__(self):
+        return len(self.records)
+
+    def __iter__(self) -> SizedIterator[TubRecord]:
+        return self
+
+    def __next__(self):
+        if self.current_index >= len(self.records):
+            raise StopIteration('No more records')
+
+        record = self.records[self.current_index]
+        self.current_index += 1
+        return record
+
+    next = __next__
+
+
+class TfmIterator(Generic[R, XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
+    def __init__(self,
+                 iterator: Iterable[R],
+                 x_transform: Callable[[R], XOut],
+                 y_transform: Callable[[R], YOut]) -> None:
+
+        self.iterator = BaseTfmIterator_(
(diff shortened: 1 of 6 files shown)


donkeycar/pipeline/types.py
```diff
@@ -0,0 +1,61 @@
...
+from typing_extensions import TypedDict
+
+TubRecordDict = TypedDict(
+    'TubRecordDict',
+    {
+        'cam/image_array': str,
+        'user/angle': float,
+        'user/throttle': float,
+        'user/mode': str,
+        'imu/acl_x': Optional[float],
```
DocGarbanzo, line 15: We should use two 3-D arrays here, one for acceleration one for gyro.

DocGarbanzo (commented):
Looks good to me.

donkeycar/pipeline/types.py
tikurahul: I am keeping the representation of the underlying dictionary here the same as what is actually stored.

tikurahul: I am keeping it the same as what is actually stored.

tikurahul:
This works as expected. I think we are pretty close to an eventual answer. 
Linear h5_pred

I realized that both `tf.data` and the old `Sequence` based API are actually have a slightly orthogonal API.
All our pipelines are `Iterator`s. Where as both `tf.Data` and `Sequence` expect a list of shallow rows to be an indexable `List` like type. Notice they dispatch a call to `__get_item__(self, index)`.

I think our models need to return a pipeline of `numpy.ndarray` producers. So imagine a type:

```python
class ToNumpy(object):
    def __init__(self) -> None:
         # Does nothing
         pass
     
    def to_nd_array(self):
         raise NotImplemented
```

Once we have this type, if we had a `SizedIterator[Tuple[ToNumpy, ToNumpy]]` then we can easily convert it to a `tf.Data` or a `Sequence` type. We are essentially one step away from a super nice API.

donkeycar/pipeline/sequence.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class BatchedIterator(SizedIterator[List[R]]):
+    def __init__(self, pipeline: SizedIterator[R], batch_size: int) -> None:
+        self.pipeline = pipeline
+        self.batch_size = batch_size
+        self.consumed = 0
+
+    def __len__(self):
+        return math.ceil(len(self.pipeline) / self.batch_size)
```
DocGarbanzo, line 138: Good to use `ceil` because before we were neglecting the the last non-completely filled batch.

donkeycar/pipeline/types.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def image(self, cached=True, normalize=False) -> np.ndarray:
+        if not self._image:
+            image_path = self.underlying['cam/image_array']
+            full_path = os.path.join(self.base_path, 'images', image_path)
+            _image = load_image_arr(full_path, cfg=self.config)
+            if normalize:
+                _image = normalize_image(_image)
+            if cached:
+                self._image = _image
```
DocGarbanzo, line 43: I think here we need to save the uint8 image in order to save memory, as the float array is w/ double precision, i.e. 8 times the size - and do the `normalize_image` on the fly.

setup.py
```diff
@@ -46,7 +46,8 @@ def package_files(directory, strip_leading):
           'PrettyTable',
           'paho-mqtt',
           "simple_pid",
-          'progress'
+          'progress',
+          'typing_extensions',
```
DocGarbanzo, line 50: We need to update the yaml files as well.

DocGarbanzo (approved):
Minor comments.

DocGarbanzo:
> This works as expected. I think we are pretty close to an eventual answer.
> Linear h5_pred
> 
> I realized that both `tf.data` and the old `Sequence` based API are actually have a slightly orthogonal API.
> All our pipelines are `Iterator`s. Where as both `tf.Data` and `Sequence` expect a list of shallow rows to be an indexable `List` like type. Notice they dispatch a call to `__get_item__(self, index)`.
> 
> I think our models need to return a pipeline of `numpy.ndarray` producers. So imagine a type:
> 
> ```python
> class ToNumpy(object):
>     def __init__(self) -> None:
>          # Does nothing
>          pass
>      
>     def to_nd_array(self):
>          raise NotImplemented
> ```
> 
> Once we have this type, if we had a `SizedIterator[Tuple[ToNumpy, ToNumpy]]` then we can easily convert it to a `tf.Data` or a `Sequence` type. We are essentially one step away from a super nice API.

Yeah `tf.Sequence` is just a sequence requiring `__get_item__(self, index)` - but  you can create `tf.Data` also from a generator (or iterator, probably too), as it has a method `from_generator()`.

PR #696, opened 2020-12-12
change the public mqtt server name to one that's listening

tawnkramer:
The previous server wasn't listening anymore. This test runs fine with the new server name.


donkeycar/tests/test_telemetry.py
```diff
@@ -8,7 +8,6 @@
 import pytest
 
 
-@pytest.mark.skip(reason="the connection always times out")
```
cloud-rocket, line 11: Why to skip if it's running with the new broker?

DocGarbanzo (approved):
Thanks Tawn!

PR #701, opened 2020-12-14
Change training pipeline from tf.Sequence to tf.data

DocGarbanzo:
# Improve new training pipeline to be iterator based

Improve pipeline use: move from building list of pipelines of single transforms to building a single pipeline with a list of transforms (actually just looping through function to go from TubRecord -> image -> augment -> normalise -> x and TubRecord -> y).

### Fixed TfmIterators and TfmIterables (for discussion)
* Iterables are the containers and are sized - these are the user objects
* Iterators are protocol objects to allow iteration, they have no logic and are local to the Iterables
* build/map_pipeline both return sized Iterables
* removed all batch logic, this is not required
* still commented but left code that uses generator based pipeline as this is simpler code

### Using new small temporary pipeline generator(for discussion, currently commented out)
* this keeps the structure simple
* added a test to check consistency of the pipeline and it's transformations
* remove empty (after moved) augmentation file
* removed augmentation from old tub (as it's not needed and we removed the old augmentation)

### New pipeline changes:
* moved augmentation into own class that is used above and can be used a  threaded or non-threaded part

### New training interface
* moved train functionality out of template and added `donkey train --tub ... --model ...`, train.py just a simple dummy script for backward compatibility

### Misc
* small fix in fast_stretch


donkeycar/pipeline/sequence.py
```diff
@@ -1,7 +1,7 @@
-from typing import Any, Callable, Generic, Iterable, Iterator, List, Sized, Tuple, TypeVar
+from typing import Any, Callable, Generic, Iterable, Iterator, List, Sized, \
```
tikurahul, line 1: Is it possible to leave the `sequence` as before ? The reason I ask is the _only_ important change here is to return a new `iterator` instance when a `__iter__(self)` is called. There are a lot more changes here. 

Even though some of your changes are good, I want to be a bit more thoughtful about them. If possible, lets leave the old shallow `list()` approach that allows training, and make fewer changes here.

donkeycar/pipeline/types.py
```diff
@@ -32,36 +32,41 @@ def __init__(self, config: Any, base_path: str, underlying: TubRecordDict) -> No
         self.underlying = underlying
         self._image: Optional[Any] = None
 
-    def image(self, cached=True, normalize=False) -> np.ndarray:
-        if not self._image:
+    def image(self, cached=True) -> np.ndarray:
+        if self._image is None:
```
tikurahul, line 36: Is this actually different from `if not self._image`?

tikurahul:
This looks great. The only concern i have is you subtly changed the `sequence` implementation and I think we might want to make a lot fewer changes there. This is why i was hoping that you would check-in the pipeline with most of the sequence as-is with a shallow `list()` and then we could take a look at look at the next step.

donkeycar/pipeline/types.py
DocGarbanzo: Yeah it is - if the image is an np.ndarray, then this tries to compare a single bool on the LHS against 'the truth value of an array' on  and falls over.

donkeycar/parts/fast_stretch.py
```diff
@@ -4,10 +4,10 @@
 import time
 
 Mx = 128  # Natural mean
-C = 0.007  # Base line fraction
-Ts = 0.02  # Tunable amplitude
-Tr = 0.7  # Threshold
-T = -0.3  # Gamma boost
+C = 0.7  # Base line fraction
```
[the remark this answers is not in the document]
DocGarbanzo: Will reset those constants, please have a look at the other change in the file which is a bug fix

donkeycar/parts/augment.py
```diff
@@ -1,117 +1,39 @@
...
+            if self.img_arr is not None:
+                self.aug_img_arr = self.augmenter.augment(self.img_arr)
+                self.img_arr = None
+            else:
+                # waiting for 500 mu s and then trying again
+                time.sleep(0.0005)
+
+    def run_threaded(self, img_arr: np.ndarray) -> np.ndarray:
+        self.img_arr = img_arr
+        time.sleep(self.delay)
```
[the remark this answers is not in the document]
DocGarbanzo: Ok, will remove.

donkeycar/pipeline/sequence.py
DocGarbanzo: I have intentionally changed this because the list is iterable and will always return a new iterator which is standard python, whereas an iterator only returns self. This means we want to work w/ iterables, the iterator is just a small pointer. If we want to work with iterators then I need to implement `__iter__` to not return self for the iterator which I can do, but this is not very pythonic I think. 
Have a look at the whole file and not only at the change as it looks much more understandable now, because you only work with the shallow lists and never with the iterators directly.  
I will do what you suggest alternatively and push that, so you can compare.

donkeycar/parts/keras.py
```diff
@@ -59,11 +65,11 @@ def set_optimizer(self, optimizer_type, rate, decay):
         else:
             raise Exception("unknown optimizer type: %s" % optimizer_type)
 
-    def get_input_shape(self):
+    def get_input_shape(self) -> tf.TensorShape:
         assert self.model is not None, "Need to load model first"
         return self.model.inputs[0].shape
 
-    def run(self, img_arr, other_arr=None):
+    def run(self, img_arr: np.ndarray, other_arr: np.ndarray = None) -> Any:
```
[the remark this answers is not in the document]
DocGarbanzo: Definitely this is what we want in the future. But that is a re-design of the parts interface, I put up some toy ideas of how to do this here: https://gist.github.com/DocGarbanzo/f80c0ce9a8984d9f5f8e857ffd1cb244. That gist is still thread based, so need to think about `async` there. But a parts re-design is a discussion for 4.2 Mar 21 earliest, I think.

donkeycar/parts/keras.py
```diff
@@ -77,7 +83,7 @@ def run(self, img_arr, other_arr=None):
         return self.inference(norm_arr, other_arr)
 
     @abstractmethod
-    def inference(self, img_arr, other_arr):
+    def inference(self, img_arr: np.ndarray, other_arr: np.ndarray) -> Any:
```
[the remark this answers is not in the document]
DocGarbanzo: See above...

donkeycar/parts/keras.py
```diff
@@ -167,6 +196,26 @@ def inference(self, img_arr, other_arr):
...
+        return {'angle_out': angle, 'throttle_out': throttle}
+
+    def output_types(self):
+        types = (tf.float64, {'angle_out': tf.float64,
+                              'throttle_out': tf.float64})
+        return types
+
+    def output_shapes(self):
+        # need to cut off None from [None, 120, 160, 3] tensor shape
+        img_shape = tf.TensorShape(self.get_input_shape().as_list()[1:])
```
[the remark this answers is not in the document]
DocGarbanzo: Yeah there is :-)

donkeycar/pipeline/sequence.py
DocGarbanzo: Done that, but pls have a look, the Iterators now have an Iterable-like `__iter__` which is a bit strange...

donkeycar/pipeline/sequence.py
```diff
@@ -1,7 +1,6 @@
...
+    Tuple, TypeVar
```
tikurahul, line 2: You should be able to completely undo changes to this file now. I fixed the `sequence`s so they behave per the iterator protocol. (https://github.com/autorope/donkeycar/pull/704)

donkeycar/parts/fast_stretch.py
```diff
@@ -49,13 +49,13 @@ def fast_stretch(image, debug=False):
     targetFl = Sl * size
     targetFh = Sh * size
 
-    count = histogram[Xl]
-    while count < targetFl:
+    count = 0
```
tikurahul, line 52: What you did here is incorrect. Please undo the changes to this file. I can explain why this works the way it does.

tikurahul (commented):
I think you can mostly undo the changes to `sequence.py` now. That and please undo the changes to `fast_stretch.py`.

tikurahul:
Please pull the latest changes to the pipeline from the `dev` branch. That should help you out.

donkeycar/pipeline/training.py
```diff
@@ -0,0 +1,112 @@
...
+                                                y_transform=get_y)
+
+        return pipeline
+
+    def make_tf_data(self):
+        dataset = tf.data.Dataset.from_generator(
+            generator=lambda: self.pipeline,
+            output_types=self.model.output_types(),
+            output_shapes=self.model.output_shapes())
+        return dataset.repeat().batch(self.batch_size)
```
tikurahul, line 57: Why `repeat()`  with no args ?

You might want to specify a `count` of the times it can be repeated. It should be set to `1` (otherwise we might end up overfitting).  The docs (https://www.tensorflow.org/api_docs/python/tf/data/Dataset#repeat) say:

```

(Optional.) A tf.int64 scalar tf.Tensor, representing the number of times the dataset should be repeated. The default behavior (if count is None or -1) is for the dataset be repeated indefinitely.
--

That's not what we want right ?

File: donkeycar/pipeline/training.py [hunk 11]
```diff
@@ -0,0 +1,112 @@
...
+        self.sequence = TubSequence(records)
+        self.batch_size = self.config.BATCH_SIZE
+        self.is_train = is_train
+        self.augmentation = ImageAugmentation(config)
+        self.pipeline = self._make_pipeline()
+
+    def __len__(self):
+        return math.ceil(len(self.pipeline) / self.batch_size)
+
+    def _make_pipeline(self):
```
tikurahul, line 33: Nit: Rename method to `create_pipeline`

donkeycar/pipeline/training.py
```diff
@@ -0,0 +1,112 @@
...
+        def get_y(record):
+            return self.model.y_transform(record)
+
+        # 2. Build pipeline using the transformations
+        pipeline = self.sequence.build_pipeline(x_transform=get_x,
+                                                y_transform=get_y)
+
+        return pipeline
+
+    def make_tf_data(self):
```
tikurahul, line 52: Nit: Rename method to `create_dataset`

donkeycar/pipeline/types.py
tikurahul: :facepalm:  

Thanks for the explanation.

donkeycar/pipeline/types.py
```diff
@@ -32,36 +32,41 @@ def __init__(self, config: Any, base_path: str, underlying: TubRecordDict) -> No
...
+        self.tub_paths = tub_paths
+        self.shuffle = shuffle
+        self.tubs: List[Tub] = [Tub(tub_path, read_only=True)
+                                for tub_path in self.tub_paths]
         self.records: List[TubRecord] = list()
 
-    def load_records(self) -> List[TubRecord]:
-        self.records.clear()
+    def train_test_split(self) -> Tuple[List[TubRecord], List[TubRecord]]:
+        print(f'Loading tubs from paths {self.tub_paths}')
```
tikurahul, line 64: You should call `self.records.clear()` as the first step just in case someone calls this method repeatedly.

tikurahul (changes requested):
This is awesome. :clap: 

I think the only major thing for you to look at is the way you construct the `tf.data.Dataset`. Specify a `repeat` count of `1`, and I think we should be okay there.

donkeycar/pipeline/training.py
DocGarbanzo: If we set this to `1` then we can run through the data only once, i.e. only one epoch. After the first epoch we are getting: 
```
WARNING:tensorflow:Your input ran out of data; interrupting training. Make sure that your dataset or generator can generate at least `steps_per_epoch * epochs` batches (in this case, 780 batches). You may need to use the repeat() function when building your dataset.
```

donkeycar/pipeline/training.py
DocGarbanzo: done.

donkeycar/pipeline/training.py
DocGarbanzo: done

donkeycar/pipeline/types.py
DocGarbanzo: Yes, good point. Done.

tikurahul (approved):

PR #705, opened 2020-12-18
Improve sequence classes

DocGarbanzo:
* Fixed TubRecord in pipeline test and updated training w/ new class
* Adding alternative pipeline generator
* Add type hints, mypy passing

### DIFF
--- a/donkeycar/pipeline/sequence.py
+++ b/donkeycar/pipeline/sequence.py
@@ -1,154 +1,90 @@
-
-from typing import (Any, Callable, Generic, Iterable, Iterator, List, Sized, Tuple, TypeVar)
+from typing import Any, Callable, Generic, Iterable, Iterator, List, Sized, \
+    Tuple, TypeVar, Union
 
 from donkeycar.pipeline.types import TubRecord
 
 # Note: Be careful when re-using `TypeVar`s.
 # If you are not-consistent mypy gets easily confused.
 
-R = TypeVar('R', covariant=True)
 X = TypeVar('X', covariant=True)
 Y = TypeVar('Y', covariant=True)
 XOut = TypeVar('XOut', covariant=True)
 YOut = TypeVar('YOut', covariant=True)
 
 
-class SizedIterator(Generic[X], Iterator[X], Sized):
+# This is a protocol type without explicitly using a `Protocol`
+# Using `Protocol` requires Python 3.7
+class SizedIterable(Generic[X], Iterable[X], Sized):
     def __init__(self) -> None:
-        # This is a protocol type without explicitly using a `Protocol`
-        # Using `Protocol` requires Python 3.7
+        pass
+
+    def __len__(self) -> int:
+        pass
+
+    def __iter__(self) -> Iterator[X]:
         pass
 
 
-class TubSeqIterator(SizedIterator[TubRecord]):
-    def __init__(self, records: List[TubRecord]) -> None:
-        self.records = records or list()
-        self.current_index = 0
-
-    def __len__(self):
-        return len(self.records)
-
-    def __iter__(self) -> SizedIterator[TubRecord]:
-        return TubSeqIterator(self.records)
-
-    def __next__(self):
-        if self.current_index >= len(self.records):
-            raise StopIteration('No more records')
-
-        record = self.records[self.current_index]
-        self.current_index += 1
-        return record
-
-    next = __next__
-
-
-class TfmIterator(Generic[R, XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
+class _BaseTfmIterator(Generic[X, Y, XOut, YOut], Iterator[Tuple[XOut, YOut]]):
+    """
+    A basic transforming iterator. Do no use this class directly.
+    """
     def __init__(self,
-                 iterable: Iterable[R],
-                 x_transform: Callable[[R], XOut],
-                 y_transform: Callable[[R], YOut]) -> None:
+                 iterable: 'TfmShallowList') -> None:
+        self.iterable = iterable
+        self.iterator: Iterator[Union[TubRecord, Tuple[X, Y]]] \
+            = iter(self.iterable.iterable)
 
+    def __iter__(self) -> '_BaseTfmIterator[X, Y, XOut, YOut]':
+        return self
+
+    def __next__(self) -> Tuple[XOut, YOut]:
+        r: Union[TubRecord, Tuple[X, Y]] = next(self.iterator)
+        return self.iterable.transform_item(r)
+
+
+class TfmShallowList(Generic[XOut, YOut], SizedIterable[Tuple[XOut, YOut]]):
+    def __init__(self,
+                 iterable: Union[List[TubRecord],
+                                 'TfmShallowList[XOut, YOut]'],
+                 x_transform: Callable[[Any], XOut],
+                 y_transform: Callable[[Any], YOut]) -> None:
         self.iterable = iterable
         self.x_transform = x_transform
         self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
 
-    def __len__(self):
-        return len(self.iterator)
+    def __len__(self) -> int:
+        return len(self.iterable)
 
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+    def __iter__(self) -> _BaseTfmIterator[X, Y, XOut, YOut]:
+        return _BaseTfmIterator(self)
 
-    def __next__(self):
-        return next(self.iterator)
-
-
-class TfmTupleIterator(Generic[X, Y, XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
-    def __init__(self,
-                 iterable: Iterable[Tuple[X, Y]],
-                 x_transform: Callable[[X], XOut],
-                 y_transform: Callable[[Y], YOut]) -> None:
-
-        self.iterable = iterable
-        self.x_transform = x_transform
-        self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
-
-    def __len__(self):
-        return len(self.iterator)
-
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
-
-    def __next__(self):
-        return next(self.iterator)
-
-
-class BaseTfmIterator_(Generic[XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
-    '''
-    A basic transforming iterator.
-    Do no use this class directly.
-    '''
-
-    def __init__(self,
-                 # Losing a bit of type safety here for a common implementation
-                 iterable: Iterable[Any],
-                 x_transform: Callable[[R], XOut],
-                 y_transform: Callable[[R], YOut]) -> None:
-
-        self.iterable = iterable
-        self.x_transform = x_transform
-        self.y_transform = y_transform
-        self.iterator = iter(self.iterable)
-
-    def __len__(self):
-        return len(self.iterator)
-
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(self.iterable, self.x_transform, self.y_transform)
-
-    def __next__(self):
-        record = next(self.iterator)
-        if (isinstance(record, tuple) and len(record) == 2):
-            x, y = record
+    def transform_item(self, r: Union[TubRecord, Tuple[X, Y]]) \
+            -> Tuple[XOut, YOut]:
+        if isinstance(r, tuple) and len(r) == 2:
+            x, y = r
             return self.x_transform(x), self.y_transform(y)
         else:
-            return self.x_transform(record), self.y_transform(record)
+            return self.x_transform(r), self.y_transform(r)
 
 
-class TubSequence(Iterable[TubRecord]):
-    def __init__(self, records: List[TubRecord]) -> None:
-        self.records = records
+class Pipeline(Generic[X, Y, XOut, YOut], SizedIterable[Tuple[XOut, YOut]]):
+    def __init__(self,
+                 iterable: Union['Pipeline', List[TubRecord]],
+                 x_transform: Callable[[Any], XOut],
+                 y_transform: Callable[[Any], YOut]) -> None:
+        self.iterable = iterable
+        self.x_transform = x_transform
+        self.y_transform = y_transform
 
-    def __iter__(self) -> SizedIterator[TubRecord]:
-        return TubSeqIterator(self.records)
+    def __iter__(self) -> Iterator[Tuple[XOut, YOut]]:
+        for record in self.iterable:
+            if isinstance(record, tuple):
+                x, y = record
+                yield self.x_transform(x), self.y_transform(y)
+            else:
+                yield self.x_transform(record), self.y_transform(record)
 
-    def __len__(self):
-        return len(self.records)
+    def __len__(self) -> int:
+        return len(self.iterable)
 
-    def build_pipeline(self,
-                       x_transform: Callable[[TubRecord], X],
-                       y_transform: Callable[[TubRecord], Y]) -> SizedIterator[Tuple[X, Y]]:
-        return TfmIterator(self, x_transform=x_transform, y_transform=y_transform)
-
-    @classmethod
-    def map_pipeline(
-            cls,
-            x_transform: Callable[[X], XOut],
-            y_transform: Callable[[Y], YOut],
-            pipeline: SizedIterator[Tuple[X, Y]]) -> SizedIterator[Tuple[XOut, YOut]]:
-        return TfmTupleIterator(pipeline, x_transform=x_transform, y_transform=y_transform)
-
-    @classmethod
-    def map_pipeline_factory(
-            cls,
-            x_transform: Callable[[X], XOut],
-            y_transform: Callable[[Y], YOut],
-            factory: Callable[[], SizedIterator[Tuple[X, Y]]]) -> SizedIterator[Tuple[XOut, YOut]]:
-
-        pipeline = factory()
-        return cls.map_pipeline(pipeline=pipeline, x_transform=x_transform, y_transform=y_transform)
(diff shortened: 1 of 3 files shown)


donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,113 @@
...
 X = TypeVar('X', covariant=True)
 Y = TypeVar('Y', covariant=True)
 XOut = TypeVar('XOut', covariant=True)
 YOut = TypeVar('YOut', covariant=True)
 
 
-class SizedIterator(Generic[X], Iterator[X], Sized):
+# This is a protocol type without explicitly using a `Protocol`
+# Using `Protocol` requires Python 3.7
+class SizedIterable(Generic[X], Iterable[X], Sized):
```
DocGarbanzo, line 17: Base class of the user objects, which are sized iterables - here `TubSequence` and `TfmShallowList`.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,113 @@
...
 
-        self.iterable = iterable
-        self.x_transform = x_transform
-        self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
 
-    def __len__(self):
-        return len(self.iterator)
+class TubSeqIterator(Iterator[TubRecord]):
```
DocGarbanzo, line 28: Iterator objects are meant to be only used as auxiliary object to iterate through a specific iterable. I.e. each iterable user object `TubSequence`, `TfmShallowList` has its own iterator. The iterator always has a reference to the iterable, but otherwise as little logic and data as possible (localisation, data hiding).

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,113 @@
...
+    def __init__(self, records: List[TubRecord]) -> None:
+        self.records = records or list()
+        self.iter = iter(self.records)
 
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+    def __iter__(self) -> 'TubSeqIterator':
+        return self
```
DocGarbanzo, line 34: Per standard, the iterator always returns `self` under the iterator protocol.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+class SizedIterable(Generic[X], Iterable[X], Sized):
     def __init__(self) -> None:
-        # This is a protocol type without explicitly using a `Protocol`
-        # Using `Protocol` requires Python 3.7
         pass
 
+    def __len__(self) -> int:
+        pass
 
-class TubSeqIterator(SizedIterator[TubRecord]):
```
DocGarbanzo, line 23: `TubSequence` and `TubSeqIterator` are obsolete. They just wrap around the underlying `list(TubRecord)` without functionality so can be removed.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-        self.current_index += 1
-        return record
-
-    next = __next__
+    def __iter__(self) -> Iterator[X]:
+        pass
 
 
-class TfmIterator(Generic[R, XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
+class _BaseTfmIterator(Generic[X, Y, XOut, YOut], Iterator[Tuple[XOut, YOut]]):
```
DocGarbanzo, line 28: The iterator is specific to its iterable (renamed `TfmShallowList`). It has a little logic as possible, it just serves to implement the iterator protocol.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+    """
+    A basic transforming iterator. Do no use this class directly.
+    """
     def __init__(self,
-                 iterable: Iterable[R],
-                 x_transform: Callable[[R], XOut],
-                 y_transform: Callable[[R], YOut]) -> None:
-
+                 iterable: 'TfmShallowList') -> None:
         self.iterable = iterable
```
DocGarbanzo, line 34: It needs a reference to the iterable.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-        self.x_transform = x_transform
-        self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+        self.iterator: Iterator[Union[TubRecord, Tuple[X, Y]]] \
+            = iter(self.iterable.iterable)
```
DocGarbanzo, line 36: And create an iterator to the underlying.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
 
-    def __len__(self):
```
DocGarbanzo, line 57: This is not a user class, `len(...)` can be called on the `TfmShallowList`.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-        return len(self.iterator)
+    def __iter__(self) -> '_BaseTfmIterator[X, Y, XOut, YOut]':
```
DocGarbanzo, line 38: Per protocol, an iterator will return `self` here.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+        return self
 
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+    def __next__(self) -> Tuple[XOut, YOut]:
+        r: Union[TubRecord, Tuple[X, Y]] = next(self.iterator)
+        return self.iterable.transform_item(r)
```
DocGarbanzo, line 43: The logic how to transform is all in the iterable (i.e. the container class) - so here is only a callback.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
 
-    def __next__(self):
-        return next(self.iterator)
 
-
-class TfmTupleIterator(Generic[X, Y, XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
```
DocGarbanzo, line 68: No need to have two different iterators.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+class TfmShallowList(Generic[XOut, YOut], SizedIterable[Tuple[XOut, YOut]]):
```
DocGarbanzo, line 46: Renamed into 'TfmShallowList` - this is the user / container class which wraps around the list of records and stores / applies the transformation logic. It is sized.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
     def __init__(self,
-                 iterable: Iterable[Tuple[X, Y]],
-                 x_transform: Callable[[X], XOut],
-                 y_transform: Callable[[Y], YOut]) -> None:
-
+                 iterable: Union[List[TubRecord],
+                                 'TfmShallowList[XOut, YOut]'],
```
DocGarbanzo, line 49: Can be initialised with two types: either `list(TubRecord)` or itself to allow chaining.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+                 x_transform: Callable[[Any], XOut],
+                 y_transform: Callable[[Any], YOut]) -> None:
```
DocGarbanzo, line 51: `Callables` and `Union` don't mix, therefore we have to choose `Any` here. It will be either a `TubRecord` in both callable or an `X` and `Y` respectively if initialised with another `TfmShallowList`.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
         self.iterable = iterable
         self.x_transform = x_transform
         self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
```
DocGarbanzo, line 77: The iterator of this iterable is only generated when called by `__iter__()`.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
-
-    def __len__(self):
-        return len(self.iterator)
 
-    def __iter__(self) -> SizedIterator[Tuple[XOut, YOut]]:
-        return BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+    def __len__(self) -> int:
+        return len(self.iterable)
```
DocGarbanzo, line 57: Sized protocol

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
 
-    def __next__(self):
-        return next(self.iterator)
+    def __iter__(self) -> _BaseTfmIterator[X, Y, XOut, YOut]:
```
DocGarbanzo, line 59: Iterable protocol, return a new Iterator object to allow independent and multiple iterations through this list.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+        return _BaseTfmIterator(self)
 
+    def transform_item(self, r: Union[TubRecord, Tuple[X, Y]]) \
+            -> Tuple[XOut, YOut]:
+        if isinstance(r, tuple) and len(r) == 2:
```
DocGarbanzo, line 64: Here is were all the magic happens, the other objects and code are only protocol implementations.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+        else:
+            return self.x_transform(r), self.y_transform(r)
 
-class BaseTfmIterator_(Generic[XOut, YOut],  SizedIterator[Tuple[XOut, YOut]]):
-    '''
-    A basic transforming iterator.
-    Do no use this class directly.
-    '''
 
+class Pipeline(Generic[X, Y, XOut, YOut], SizedIterable[Tuple[XOut, YOut]]):
```
DocGarbanzo, line 71: This is a parallel implemention of the `TfmShallowList` with its `_BaseTfmIterator`. It provides the exact same functionality as the iterable / iterator protocol, by using the generator protocol.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-    def __init__(self, records: List[TubRecord]) -> None:
-        self.records = records
-
-    def __iter__(self) -> SizedIterator[TubRecord]:
-        return TubSeqIterator(self.records)
-
-    def __len__(self):
-        return len(self.records)
-
-    def build_pipeline(self,
```
DocGarbanzo, line 133: These helper functions are not needed, `TfmShallowList` or `Pipeline` and just be constructed as user objects.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
+    def __iter__(self) -> Iterator[Tuple[XOut, YOut]]:
+        for record in self.iterable:
+            if isinstance(record, tuple):
+                x, y = record
+                yield self.x_transform(x), self.y_transform(y)
+            else:
+                yield self.x_transform(record), self.y_transform(record)
 
-    @classmethod
-    def map_pipeline_factory(
```
DocGarbanzo, line 147: Obsolete.

donkeycar/pipeline/training.py
```diff
@@ -21,7 +19,7 @@ class BatchSequence(object):
     def __init__(self, model, config, records: List[TubRecord], is_train: bool):
         self.model = model
         self.config = config
-        self.sequence = TubSequence(records)
```
DocGarbanzo, line 24: `TubSequence` not needed.

donkeycar/pipeline/training.py
```diff
@@ -44,9 +42,9 @@ def get_y(record):
             return self.model.y_transform(record)
 
         # 2. Build pipeline using the transformations
-        pipeline = self.sequence.build_pipeline(x_transform=get_x,
-                                                y_transform=get_y)
-
+        pipeline = TfmShallowList(self.sequence,
```
DocGarbanzo, line 45: Create the object directly.

donkeycar/tests/test_pipeline.py
```diff
@@ -25,7 +25,7 @@ def random_record() -> TubRecord:
         'imu/gyr_y': None,
         'imu/gyr_z': None
     }
-    return TubRecord('/base', None, underlying=underlying)
+    return TubRecord(config=None, base_path='/base', underlying=underlying)
```
DocGarbanzo, line 28: Small fix, should have seen this before : - )

donkeycar/tests/test_pipeline.py
```diff
@@ -34,8 +34,7 @@ def random_record() -> TubRecord:
 class TestPipeline(unittest.TestCase):
 
     def setUp(self):
-        records = random_records(size=size)
-        self.sequence = TubSequence(records=records)
```
DocGarbanzo, line 38: Obsolete

donkeycar/tests/test_pipeline.py
```diff
@@ -101,40 +99,67 @@ def test_more_map_operations(self):
...
+        self.assertEqual(l1, l2)
+        for e, t in zip(extract, transformed):
+            ex, ey = e
+            tx, ty = t
+            self.assertAlmostEqual(2 * ex, tx)
+            self.assertAlmostEqual(3 * ey, ty)
 
-        x1, y1 = transformed_list[index]
-        x2, y2 = transformed_list_2[index]
+    def test_iterator_consistency_pipeline(self):
```
DocGarbanzo, line 131: Same test as above, just using alternative `Pipeline` class.

donkeycar/tests/test_pipeline.py
```diff
@@ -101,40 +99,67 @@ def test_more_map_operations(self):
...
+        l1 = list(transformed)
+        l2 = list(transformed)
+        self.assertEqual(l1, l2)
+        for e, t in zip(extract, transformed):
+            ex, ey = e
+            tx, ty = t
+            self.assertAlmostEqual(2 * ex, tx)
+            self.assertAlmostEqual(3 * ey, ty)
 
     def test_iterator_consistency(self):
```
DocGarbanzo, line 160: Same test as above but w/t `TfmShallowList`.

DocGarbanzo:
You might want to look at the file `sequence.py` directly instead of the diff, because it's much more readable. Effectively implemented the same functionality twice in 2/3 of the code, either one sequence (iterable) with one iterator, or alternatively one sequence with one generator.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-        self.y_transform = y_transform
-        self.iterator = BaseTfmIterator_(
-            iterable=self.iterable, x_transform=self.x_transform, y_transform=self.y_transform)
+        self.iterator: Iterator[Union[TubRecord, Tuple[X, Y]]] \
+            = iter(self.iterable.iterable)
 
-    def __len__(self):
-        return len(self.iterator)
+    def __iter__(self) -> '_BaseTfmIterator[X, Y, XOut, YOut]':
+        return self
```
tikurahul, line 39: Its not returning a new instance of the `Iterator` here though.

donkeycar/pipeline/sequence.py
```diff
@@ -1,154 +1,90 @@
...
-            cls,
-            x_transform: Callable[[X], XOut],
-            y_transform: Callable[[Y], YOut],
-            pipeline: SizedIterator[Tuple[X, Y]]) -> SizedIterator[Tuple[XOut, YOut]]:
-        return TfmTupleIterator(pipeline, x_transform=x_transform, y_transform=y_transform)
+    def __iter__(self) -> Iterator[Tuple[XOut, YOut]]:
+        for record in self.iterable:
+            if isinstance(record, tuple):
+                x, y = record
+                yield self.x_transform(x), self.y_transform(y)
```
tikurahul, line 84: I am not a big fan of the generator here.

tikurahul (commented):
I see what the intention of the changes are. 

We can simplify the existing implementation - but defining a `TfmdShalllowList` type does not make a lot of sense to me. Its essentially just a `List[TubRecord] + a couple of helpers like before. 

If anything it actually makes it a bit more confusing. Nothing in the implementation of the class enforces that this is actually a shallow iterator. You also inadvertently broke the iterator protocol in the `_BaseTfmdIterator` by returning `self` directly. 

If the real intention to just use `Iterable`s over `Iterator`s ? it does not buy us a lot save for a little bit of brevity. If anything, this implementation might actually be more useful further down the line. 

I am tempted to make a few minor changes here instead, rather than having to introduce a generator.

donkeycar/pipeline/sequence.py
DocGarbanzo: Correct. Iterators are expected to return `self` here.

donkeycar/pipeline/sequence.py
DocGarbanzo: I'm just showing an alternative way to implement the Iterable/Iterator protocol using generators, we don't need it, I can delete it.

DocGarbanzo:
> I see what the intention of the changes are.
> 
> We can simplify the existing implementation - but defining a `TfmdShalllowList` type does not make a lot of sense to me. Its essentially just a `List[TubRecord]` + a couple of helpers like before.
> 
> If anything it actually makes it a bit more confusing. Nothing in the implementation of the class enforces that this is actually a shallow iterator. You also inadvertently broke the iterator protocol in the `_BaseTfmdIterator` by returning `self` directly.
> 
> If the real intention to just use `Iterable`s over `Iterator`s ? it does not buy us a lot save for a little bit of brevity. If anything, the original implementation might actually be more useful further down the line.
> 
> I am tempted to make a few minor changes to the original implementation instead, rather than having to introduce a generator to the mix.

You are right. The implementation is  following the "standard" as far as I can see - please have a look here: https://python-patterns.guide/gang-of-four/iterator/. `Iterable` (which is any kind of shallow or deep list) returns its `Iterator` and the `Iterator` returns `self` under `__iter()`.  Sticking to that concept makes everything very clear and tidy in my view, and it's the expected behaviour for iterables/iterators of that kind (file iterables are different). What we currently have is working, but it's not following that design pattern. 
Shallowness is explicitly part of python's `Iterator/Generator` design, `__next__` will only be called if it is needed like in this example:
```python
# Creating a generator, but same concept for Iterables/Iterators, so you can swap `range(10)` for our shallow list 
# nothing gets evaluated in this line
g = (k * 2 for k in range(10))
# only here __next__ is getting called on g which will call __next__ on range:
lg = list(g)
# or here:
for i in g:
    print(i)
```

PR #706, opened 2020-12-21
Added the ability to train PyTorch models

EricWiener:
I needed to edit the Mac environment YAML file. I will also likely need to modify the Windows and Ubuntu environments as well but would appreciate some feedback on the Mac environment first.

I tried to make as few changes as possible to the existing code to preserve backward compatibility. Please let me know what changes are needed.

I tested this on a MacBook Pro 2019 + Jetson Nano.


donkeycar/parts/pytorch/torch_data.py
```diff
@@ -0,0 +1,138 @@
...
+    transform = transforms.Compose([
+        transforms.Resize(input_size),
+        transforms.ToTensor(),
+        transforms.Normalize(mean=mean, std=std)
+    ])
+
+    return transform
+
+
+class TorchTubDataset(Dataset):
```
tikurahul, line 42: You might want to consider using: https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset

That way you can plug in the `Iterator` returned by the pipeline and not have to deal with the indexing part. Your default transform can be plugged in to the pipeline directly.

donkeycar/pipeline/types.py
```diff
@@ -3,7 +3,7 @@
 
 import numpy as np
 from donkeycar.parts.tub_v2 import Tub
-from donkeycar.utils import load_image_arr, normalize_image, train_test_split
+from donkeycar.utils import load_image_arr, load_pil_image, normalize_image, train_test_split
```
tikurahul, line 6: I think this might be unnecessary (unless there is a specific optimization you are going after).  We could just add a `raw` option to `load_image` to load the raw PIL image if that's what you were going for. 

Otherwise, `Image.from_array(...)` can be used to construct a `PIL` `Image` type from a `numpy` array. That might be what you want to do in the pipeline itself. Also add a note on channel ordering for those who might be unfamiliar with the incoming shapes.

For the reverse, you could do a `np.as_array(pil_image)`.

donkeycar/templates/train.py
```diff
@@ -1,18 +1,17 @@
 #!/usr/bin/env python3
 """
-Scripts to train a keras model using tensorflow.
```
tikurahul, line 3: This has been refactored recently. This should be a lot easier now.

tikurahul (changes requested):
This is looking pretty awesome. I am very excited about the potential for PyTorch Lightning models. I am so glad that these are the kinds of workflows the new pipeline makes super easy.

donkeycar/pipeline/types.py
EricWiener: I was trying to avoid an unnecessary conversion from `PIL` -> `numpy` -> `PIL` -> `tensor`. I wasn't sure modifying `load_image_arr` to also return a raw image made sense, since the function name specifies loading an array.

Could I change `load_image_arr` to be named `load_image` and then have an option whether to return a `numpy` array or a `PIL` image?

donkeycar/templates/train.py
EricWiener: Sorry - not sure I understand what changes to make here.

donkeycar/parts/pytorch/torch_data.py
EricWiener: Just pushed again and changed the Dataset to IterableDataset.

donkeycar/parts/pytorch/torch_data.py
```diff
@@ -0,0 +1,149 @@
...
+    transform = transforms.Compose([
+        transforms.Resize(input_size),
+        transforms.ToTensor(),
+        transforms.Normalize(mean=mean, std=std)
+    ])
+
+    return transform
+
+
+class TorchTubDataset(IterableDataset):
```
tikurahul, line 43: :clap: This is so awesome

donkeycar/pipeline/types.py
tikurahul: Yeah.. lets make that change. Lets have the following:

`load_pil_image` returns a `Image`. 
`load_image` converts a PIL image -> numpy.ndarray using `np.as_array(load_pil_image(...))`. 

That way you don't have to do back and forth.

donkeycar/templates/train.py
tikurahul: I meant this got moved to `manage.py`. @DocGarbanzo can help you figure out how to move them over.

install/envs/mac.yml
```diff
@@ -3,9 +3,10 @@ name: donkey
 channels:
   - defaults
   - conda-forge
+  - pytorch
 
 dependencies:
-  - python=3.7
+  - python=3.8
```
tikurahul, line 9: Do we need to move to 3.8?

EricWiener: I had some issues installing with 3.7, but I'll retry and see if those are still present

donkeycar/pipeline/types.py
EricWiener: Sounds good. Just pushed again with these changes.

install/envs/mac.yml
EricWiener: 3.7 works. I updated the enviroment configs

install/envs/mac.yml
```diff
@@ -25,6 +26,14 @@ dependencies:
   - PrettyTable
   - pyfiglet
   - mypy
+  - pytorch=1.7.1
+  - torchvision
+  - torchaudio
+  - docopt
```
DocGarbanzo, line 32: this is included above

install/envs/mac.yml
```diff
@@ -25,6 +26,14 @@ dependencies:
...
+  - pytorch-lightning
+  - imgaug
```
DocGarbanzo, line 34: this is included above

install/envs/mac.yml
```diff
@@ -25,6 +26,14 @@ dependencies:
...
+  - matplotlib
```
DocGarbanzo, line 35: This is included above.

install/envs/mac.yml
```diff
@@ -25,6 +26,14 @@ dependencies:
...
+  - numpy
```
DocGarbanzo, line 36: I think we forgot this as it always comes as a dependency of pandas or tf, but better to put it here!

install/envs/ubuntu.yml
```diff
@@ -25,9 +26,16 @@ dependencies:
...
   - pyfiglet
   - mypy
+  - pytorch=1.7.1
+  - torchvision
+  - torchaudio
+  - docopt
+  - pytorch-lightning
+  - imgaug
+  - matplotlib
+  - numpy
```
DocGarbanzo, line 36: pls see above.

DocGarbanzo (changes requested):
This looks mostly fine, thank you @EricWiener.  I was hoping we wouldn't need to duplicate the existing pipeline at all as the interaction between the data and the model would be implemented in the model itself. This will be easy in the pipeline and a bit more tricky on the training interface, but I can have a go at that, no show stopper. Otherwise only minor refactorings and questions. 
Can you please also add a test that builds the model and just inferences it and one that does the training, check `test_train.py` - here we only test a couple of steps for convergence.

donkeycar/templates/train.py
EricWiener: Just pushed with a commit that moves this into base.py and reverts train.py

docs/guide/robot_sbc/setup_jetson_nano.md
```diff
@@ -16,10 +16,50 @@ Visit the official [Nvidia Jetson Nano Getting Started Guide](https://developer.
...
-sudo apt-get install build-essential python3 python3-dev python3-pip python3-pandas python3-h5py libhdf5-serial-dev hdf5-tools nano ntp
+sudo -H pip3 install -U pip testresources setuptools
+sudo -H pip3 install -U futures==3.1.1 protobuf==3.12.2 pybind11==2.5.0
+sudo -H pip3 install -U cython==0.29.21
+sudo -H pip3 install -U numpy==1.19.0
+sudo -H pip3 install -U future==0.18.2 mock==4.0.2 h5py==2.10.0 keras_preprocessing==1.1.2 keras_applications==1.0.8 gast==0.3.3
+sudo -H pip3 install -U grpcio==1.30.0 absl-py==0.9.0 py-cpuinfo==7.0.0 psutil==5.7.2 portpicker==1.3.1 six requests==2.24.0 astor==0.8.1 termcolor==1.1.0 wrapt==1.12.1 google-pasta==0.2.0
+sudo -H pip3 install -U scipy==1.4.1
+sudo -H pip3 install -U pandas==1.0.5
+sudo -H pip3 install -U gdown
```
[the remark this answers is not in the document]
EricWiener: Just pushed with this update.

donkeycar/parts/pytorch/ResNet18.py
```diff
@@ -0,0 +1,106 @@
...
+    model.fc = nn.Linear(512, 2)
+
+    for param in model.fc.parameters():
+        param.requires_grad = True
+
+    return model
+
+
+class ResNet18(pl.LightningModule):
+  def __init__(self, input_shape=(128, 3, 224, 224), output_size=(2,)):
```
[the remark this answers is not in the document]
EricWiener: Just formatted with pep-8 and pushed.

donkeycar/parts/pytorch/torch_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    from donkeycar.parts.pytorch.ResNet18 import ResNet18
+
+    if model_type is None:
+        model_type = cfg.DEFAULT_MODEL_TYPE
+    print("\"get_model_by_type\" model Type is: {}".format(model_type))
+
+    input_shape = (cfg.BATCH_SIZE, cfg.IMAGE_DEPTH, cfg.IMAGE_H, cfg.IMAGE_W)
+    
+    if model_type == "linear":
+        model = ResNet18(input_shape=input_shape)
```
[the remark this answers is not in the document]
EricWiener: Nope. This is just used for printing a weight summary:
```
# Used by PyTorch Lightning to print an example model summary
self.example_input_array = torch.rand(input_shape)
```

docs/guide/host_pc/setup_mac.md
```diff
@@ -37,6 +37,7 @@ conda env create -f install/envs/mac.yml
 conda activate donkey
 pip install -e .[pc]
 ```
+Note: if you are using ZSH (you'll know if you are), you won't be able to run `pip install -e .[pc]`. If you followed the setup using `conda`, then you can just run `pip install -e .` and you'll be fine. If you didn't use `conda`, you should manually install the packages in `setup.py` under the `pc` flag.
```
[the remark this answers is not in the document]
EricWiener: You're right. I updated the docs to mention that you need to escape the brackets

donkeycar/management/base.py
```diff
@@ -375,7 +375,7 @@ def plot_predictions(self, cfg, tub_paths, model_path, limit, model_type):
 
         for record in records:
             img_filename = os.path.join(base_path, Tub.images(), record['cam/image_array'])
-            img = load_image_arr(img_filename, cfg)
+            img = load_image(img_filename, cfg)
             user_angle = float(record["user/angle"])
             user_throttle = float(record["user/throttle"])
             pilot_angle, pilot_throttle = model.run(img)
```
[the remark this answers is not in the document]
EricWiener: Sorry - not sure I understand this.

donkeycar/parts/pytorch/torch_data.py
```diff
@@ -0,0 +1,149 @@
...
+            required """
+
+        def y_transform(record: TubRecord):
+            angle: float = record.underlying['user/angle']
+            throttle: float = record.underlying['user/throttle']
+            return torch.tensor([angle, throttle])
+
+        def x_transform(record: TubRecord):
+            # Loads the result of Image.open()
+            img_arr = record.image(cached=True, as_nparray=False)
```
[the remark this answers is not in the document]
EricWiener: Should I make `x_transform` and `y_transform` instance methods? Or should I leave as is?

donkeycar/parts/pytorch/torch_train.py
```diff
@@ -0,0 +1,58 @@
...
+
+    is_torch_model = model_ext == '.ckpt'
+    if is_torch_model:
+        model = f'{model_name}.ckpt'
+
+    if not model_type:
+        model_type = cfg.DEFAULT_MODEL_TYPE
+
+    tubs = tub_paths.split(',')
+    tub_paths = [os.path.expanduser(tub) for tub in tubs]
```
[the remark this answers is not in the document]
EricWiener: I think it should still work?
```
>>> x = 'test'
>>> print(x.split(','))
['test']
```

install/envs/ubuntu.yml
EricWiener: Just pushed with fixes

install/envs/mac.yml
EricWiener: Just pushed with fixes

install/envs/mac.yml
EricWiener: Just pushed with fixes

install/envs/mac.yml
EricWiener: Just pushed with fixes

install/envs/mac.yml
EricWiener: Just pushed with fixes

donkeycar/parts/pytorch/torch_data.py
EricWiener: @DocGarbanzo anything you want me to do here?

donkeycar/management/base.py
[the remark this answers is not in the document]
EricWiener: Actually, I think I realized the issue. Shouldn't be a problem because of one of the recent commits

EricWiener:
> This looks mostly fine, thank you @EricWiener. I was hoping we wouldn't need to duplicate the existing pipeline at all as the interaction between the data and the model would be implemented in the model itself. This will be easy in the pipeline and a bit more tricky on the training interface, but I can have a go at that, no show stopper. Otherwise only minor refactorings and questions.
> Can you please also add a test that builds the model and just inferences it and one that does the training, check `test_train.py` - here we only test a couple of steps for convergence.

Working on this now.

donkeycar/parts/pytorch/torch_data.py
tikurahul: Its a good idea to leave this class as-is because we can account for things like `multiprocessing` as we get closer to the ML frameworks. Its good for our `Iterator`s to be more "general purpose" and leave all the specificity to the ML frameworks.

donkeycar/parts/pytorch/torch_data.py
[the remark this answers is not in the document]
tikurahul: You should leave this as-is. Look at the comment above. Having a class here, gives us more flexibility to take advantage of what the ML framework as to offer (aka multiprocessing).

donkeycar/parts/pytorch/torch_train.py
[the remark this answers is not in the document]
tikurahul: I think the intent it unclear though. Do you mind adding an `isinstance(...)` check here so we can remember why this works in the future? :smile:

donkeycar/parts/pytorch/torch_data.py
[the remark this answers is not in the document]
DocGarbanzo: Yes, definitely the training with all bells and whistles needs to live in the model which it does on the TF side. I'm fine with this, I can show you what I mean in a small refactoring PR afterwards.

donkeycar/parts/pytorch/torch_train.py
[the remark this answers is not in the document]
DocGarbanzo: You're right, @EricWiener - this works fine, all good from my side.

docs/guide/host_pc/setup_mac.md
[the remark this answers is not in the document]
DocGarbanzo: Awesome, thanks. That'll be a big mystery resolver for many folks.

donkeycar/management/base.py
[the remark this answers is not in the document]
DocGarbanzo: Thanks looks good.

donkeycar/parts/pytorch/torch_utils.py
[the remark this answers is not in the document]
DocGarbanzo: So is the first dimension of the `input_shape` required at all then? Or can we leave it out? Optimally the batch size would only be used in training.

donkeycar/tests/test_torch.py
```diff
@@ -0,0 +1,112 @@
...
+                         progress_bar_refresh_rate=30, max_epochs=30)
+    trainer.fit(model, data_module)
+    final_loss = model.loss_history
+    assert final_loss < 0.10
+
+    # Check the batch data makes sense
+    for batch in data_module.train_dataloader():
+        x, y = batch
+        assert(x.shape == (config.IMAGE_DEPTH, config.IMAGE_H, config.IMAGE_W))
+        assert(.shape == (2,))
```
DocGarbanzo, line 107: Missing `y` - if you look at Travis, you see it falls over there.

install/nano/requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+psutil==5.7.2
+portpicker==1.3.1
+six
+requests==2.24.0
+astor==0.8.1
+termcolor==1.1.0
+wrapt==1.12.1
+google-pasta==0.2.0
+scipy==1.4.1
+pandas==1.0.5
```
DocGarbanzo, line 27: I think on the other platforms we don't specify the versions except where we know it is need (e.g. h5py), definitely in the conda and pip installations. @tikurahul - what do you think?

DocGarbanzo (commented):
Thanks @EricWiener - lots of work over Xmas. Great new functionality. Only small minor questions /issues from my side. In general, have a look at the CI - we normally want this to pass.

donkeycar/tests/test_torch.py
EricWiener: Thanks. Will fix this.

donkeycar/parts/pytorch/torch_data.py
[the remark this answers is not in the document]
EricWiener: @DocGarbanzo the point of the [PyTorch Lightning Data Module](https://pytorch-lightning.readthedocs.io/en/latest/datamodules.html) is to encapsulate all the logic regarding data preparation:

```
A datamodule encapsulates the five steps involved in data processing in PyTorch:
1. Download / tokenize / process.
2. Clean and (maybe) save to disk.
3. Load inside Dataset.
4. Apply transforms (rotate, tokenize, etc…).
5. Wrap inside a DataLoader.
```

This makes the DataModule much more re-usable. Your call, but moving the transforms into the model might lose the benefits of having the DataModule. Additionally, a lot of the models will probably be applying similar augmentations, so it might be helpful for these to be coupled with the data vs. with the model.

EricWiener:
@DocGarbanzo the first dimension in the input shape is required (it won't let me reply to your comment above). If you run without a batch size you get:
```
RuntimeError: Expected 4-dimensional input for 4-dimensional weight [64, 3, 7, 7], but got 3-dimensional input of size [3, 120, 160] instead
```

I could just use `1` for the batch size, but I thought it would be a better estimate if I used what the batch size actually was.

install/nano/requirements.txt
tikurahul: Yeah. Let's not pin the versions.

EricWiener: Versions are no longer pinned

donkeycar/parts/pytorch/torch_utils.py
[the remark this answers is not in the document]
tikurahul: PyTorch requires that the batch size be specified. They are more explicit than Keras. Personally I prefer that to `None`.

setup.py
```diff
@@ -75,6 +75,11 @@ def package_files(directory, strip_leading):
           ],
           'ci': ['codecov'],
           'tf': ['tensorflow>=2.2.0'],
+          'torch': [
```
tikurahul, line 78: Formatting seems off?

tikurahul (approved):
There is a minor formatting issue in `requirements.txt` (am reviewing from my phone so apologies if there is not). If you fix that you can merge this.

setup.py
EricWiener: Just fixed. Sorry about that. I'll need to get VSCode to format on save

tikurahul (approved):

PR #707, opened 2020-12-21
Add support for models with multiple inputs and provide developer guide for model building

DocGarbanzo:
* Add support for multi-dimensional model input by using dictionaries for tf data inputs
* Symmetric behaviour for `X, Y` data input model  - converting from tub record into numpy arrays and then into dictionaries.
* Simplify model interface by implementing output_types() directly in the base class using output_shapes() dictionary.
* Adding developer guide for own model development.
* Updated donkey command documentation with `train` command.
* Improve asserts and type hints in keras.py
* Added missing __init__.py in parts module.
* Add cool ascii text for donkey init and update yml and setup files including mypy.
* Remove model training test from Travis and change the test to relative convergence. This avoids random fall overs in CI.
* Added test of tf.data as used in the training pipeline through re-implementation of data transformation from tub records to tf expected dictionaries, for all currently supported models.

### DIFF
--- a/donkeycar/pipeline/training.py
+++ b/donkeycar/pipeline/training.py
@@ -1,15 +1,16 @@
 import math
 import os
-from pathlib import Path
-from typing import List
+from typing import List, Dict, Union
 
+from donkeycar.config import Config
+from donkeycar.parts.keras import KerasPilot
 from donkeycar.parts.tflite import keras_model_to_tflite
-from donkeycar.pipeline.sequence import TubRecord
-from donkeycar.pipeline.sequence import TubSequence
+from donkeycar.pipeline.sequence import TubRecord, TubSequence, TfmIterator
 from donkeycar.pipeline.types import TubDataset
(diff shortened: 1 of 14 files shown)


donkeycar/pipeline/training.py
```diff
@@ -27,37 +32,52 @@ def __init__(self, model, config, records: List[TubRecord], is_train: bool):
...
+            """ Extracting x from record for training"""
+            # this transforms the record into x for training the model to x,y
             x0 = self.model.x_transform(record)
-            x1 = self.augmentation.augment(x0) if self.is_train else x0
-            x2 = normalize_image(x1)
-            return x2
-
-        def get_y(record):
-            return self.model.y_transform(record)
+            # for multiple input tensors the return value here is a tuple
```
tikurahul, line 46: This is beautiful :smiley:

tikurahul (approved):
This is awesome. I am so happy we are at this point where you don't have to touch `train.py` anymore.

PR #708, opened 2020-12-22
Minor changes for 4.1 in tub conversion script and developer doc

DocGarbanzo:
* Update conversion script to translate discontinuous data.
* In developer guide add disclaimer for version and correct intra-page links.

### DIFF
--- a/scripts/convert_to_tub_v2.py
+++ b/scripts/convert_to_tub_v2.py
@@ -35,28 +35,32 @@ def convert_to_tub_v2(paths, output_path):
 
         record_paths = legacy_tub.gather_records()
         bar = IncrementalBar('Converting', max=len(record_paths))
-        prev_rec_num = None
+        previous_index = None
         for record_path in record_paths:
             try:
                 contents = Path(record_path).read_text()
                 record = json.loads(contents)
                 image_path = record['cam/image_array']
-                rec_num = int(image_path.split('_')[0])
+                current_index = int(image_path.split('_')[0])
                 image_path = os.path.join(legacy_tub.path, image_path)
                 image_data = Image.open(image_path)
                 record['cam/image_array'] = image_data
                 # first record or they are continuous, just append
-                if not prev_rec_num or rec_num == prev_rec_num + 1:
+                if not previous_index or current_index == previous_index + 1:
                     output_tub.write_record(record)
-                    prev_rec_num = rec_num
+                    previous_index = current_index
                 # otherwise fill the gap with dummy records
                 else:
-                    prev_rec_num += 1
-                    while prev_rec_num < rec_num:
+                    # Skipping over previous record here because it has
+                    # already been written.
+                    previous_index += 1
+                    # Adding empty record nodes, and marking them deleted
+                    # until the next valid record.
+                    while previous_index < current_index:
                         idx = output_tub.manifest.current_index
-                        output_tub.write_record({})
+                        output_tub.write_record()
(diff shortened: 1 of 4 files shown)


scripts/convert_to_tub_v2.py
```diff
@@ -22,29 +22,44 @@
...
     for legacy_tub in legacy_tubs:
         if not output_tub:
-            output_tub = Tub(output_path, legacy_tub.inputs, legacy_tub.types, list(legacy_tub.meta.items()))
+            output_tub = Tub(output_path, legacy_tub.inputs,
+                             legacy_tub.types, list(legacy_tub.meta.items()))
 
         record_paths = legacy_tub.gather_records()
         bar = IncrementalBar('Converting', max=len(record_paths))
-
+        prev_rec_num = None
```
tikurahul, line 38: Nit: Call this `previous_index`.

scripts/convert_to_tub_v2.py
```diff
@@ -22,29 +22,44 @@
...
         for record_path in record_paths:
             try:
                 contents = Path(record_path).read_text()
                 record = json.loads(contents)
                 image_path = record['cam/image_array']
+                rec_num = int(image_path.split('_')[0])
```
tikurahul, line 44: Nit: Call this `current_index`

scripts/convert_to_tub_v2.py
```diff
@@ -22,29 +22,44 @@
...
                 image_data = Image.open(image_path)
                 record['cam/image_array'] = image_data
-                output_tub.write_record(record)
+                # first record or they are continuous, just append
+                if not prev_rec_num or rec_num == prev_rec_num + 1:
+                    output_tub.write_record(record)
+                    prev_rec_num = rec_num
+                # otherwise fill the gap with dummy records
+                else:
+                    prev_rec_num += 1
```
tikurahul, line 54: Add a comment here that says something like:
```python
# Skipping over previous record here because it has already been written.
# Adding empty record nodes, and marking them deleted until the next valid record.
```

scripts/convert_to_tub_v2.py
```diff
@@ -22,29 +22,44 @@
...
+                    while prev_rec_num < rec_num:
+                        idx = output_tub.manifest.current_index
+                        output_tub.write_record({})
```
tikurahul, line 57: Define a `dict` `EMPTY`  which looks something like:

```python
{'__empty__': true}
```
Write that for every deleted record.  That way the intention here is very clear when reading though the file.

tikurahul (changes requested):

scripts/convert_to_tub_v2.py
DocGarbanzo: Ok.

scripts/convert_to_tub_v2.py
DocGarbanzo: Ok.

scripts/convert_to_tub_v2.py
DocGarbanzo: Ok.

scripts/convert_to_tub_v2.py
DocGarbanzo: Good point. And I think this logic should be part of the tub itself, not the user of the function. So we rather define a specific empty record in the tub, that will be generated in the same consistent way by whoever needs to add empty records.

tikurahul (approved):

PR #709, opened 2020-12-22
* Update doc with developer section for creating own models

DocGarbanzo:
* Adding new page with detailed info how to create own models
* Update donkey command section with `donkey train` command.

### DIFF
--- a/docs/dev_guide/model.md
+++ b/docs/dev_guide/model.md
@@ -1,0 +1,324 @@
+# How to build your own model
+
+---
+ **Note:** _This requires version >= 4.1.X_
+
+---
+
+* [Overview](model.md#overview)
+* [Constructor](model.md#constructor)
+* [Training Interface](model.md#training-interface)
+* [Parts Interface](model.md#parts-interface)
+* [Example](model.md#example)
+
+## Overview
+
+You might want to write your own model:
+
+* If you find the models that ship with donkey not sufficient, and you want to
+  experiment with your own model infrastructure
+* If you want to add more input data to the model because your car has more
+  sensors
+
+## Constructor
+
+Models are located in `donkeycar/parts/keras.py`. Your own model needs to
+inherit from `KerasPilot` and initialize your model:
+
+```python
+class KerasSensors(KerasPilot):
+    def __init__(self, input_shape=(120, 160, 3), num_sensors=2):
+        super().__init__()
+        self.num_sensors = num_sensors
+        self.model = self.create_model(input_shape)
+```
+Here, you implement the [keras model](https://www.tensorflow.org/guide/keras/sequential_model)
+in the member function `create_model()`. The model needs to have labelled input
+and output tensors. These are required for the training to work.
+
+
+## Training interface
+
+What is required for your model to work, are the following functions:
+
+```python
+def compile(self):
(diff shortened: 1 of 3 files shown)


docs/dev_guide/model.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+In this function you define how to extract the input data from your
+recorded data. This data is usually called `X` in the ML frame work . We have
+shown the implementation in the base class which works for all models that have
+only the image as input. 
+
+The function returns a single data item if the model has only one input. You 
+need to return a tuple if your model uses more input data.
+**Note:** _If your model has more inputs, the tuple needs to have the image in
```
tikurahul, line 70: Add 2 extra lines to emphasize the `Note`.  (Creates a new paragraph)

docs/dev_guide/model.md
```diff
@@ -0,0 +1,310 @@
...
+    return {'img_in': x}
+```
+Here we require a translation of how the `X` value that you extracted above will
+be fed into `tf.data`. Note, `tf.data` expects a dictionary if the model has
+more than one input variable, so we have chosen to use dictionaries also in the
+one-argument case for consistency. Above we have shown the implementation in the
+base class which works for all models that have only the image as input. You
+don't have to overwrite neither `x_transform` nor
+`x_translate` if your model only uses the image as input data.
+**Note:** _the keys of the dictionary must match the name of the **input**
```
tikurahul, line 94: Same as above.

docs/dev_guide/model.md
```diff
@@ -0,0 +1,310 @@
...
+    img_shape = self.get_input_shape()[1:]
+    shapes = ({'img_in': tf.TensorShape(img_shape)},
+              {'angle_out': tf.TensorShape([15]),
+               'throttle_out': tf.TensorShape([20])})
+    return shapes
+```
+This function returns a tuple of _two_ dictionaries that tells tensorflow which
+shapes are used in the model. We have shown the example of the 
+`KerasCategorical` model here.
+**Note 1:** _The keys of the two dictionaries must match the name of the
```
tikurahul, line 123: Nit: Call this **Note** and not **Note 1**. 
Add a paragraph like the ones above.

docs/dev_guide/model.md
```diff
@@ -0,0 +1,310 @@
...
+    outputs = self.model.predict(img_arr)
+    steering = outputs[0]
+    throttle = outputs[1]
+    return steering[0][0], throttle[0][0]
+```
+Here we are showing the implementation of the linear model. Please note that 
+the input tensor shape always contains the batch dimension in the first 
+place, hence the shape of the input image is adjusted from 
+`(120, 160, 3) -> (1, 120, 160, 3)`.
+**Note:** _If you are passing another array in the`other_arr` variable, you will
```
tikurahul, line 149: Same here.

tikurahul (approved):

docs/dev_guide/model.md
DocGarbanzo: yup

docs/dev_guide/model.md
DocGarbanzo: yup

docs/dev_guide/model.md
DocGarbanzo: I have 2 notes there...

PR #711, opened 2020-12-24
Minor fixes for Python3.5 and RPi4 compatibility

showsep93:
-print(f"wsCalibrate {message}") replaced with print("wsCalibrate ",message) in donkeycar/donkeycar/parts/web_controller/web.py  so that it works with Python 3.5 (in case using Ubuntu 16.04).
-FindCar donkey function now works with RPi model 4 too, whose MAC addresses are different.


tikurahul:
We no longer plan on supporting 3.5. There is a lot of code in Donkey 4.x which will not work on 3.5 anymore.

PR #712, opened 2020-12-27
fixing FindCar for RPi4

showsep93:
FindCar donkey function will now work with Raspberries model 4 too, whose MAC addresses are different.


DocGarbanzo:
@showsep93 - This is great! - Thanks for fixing this, I see this is working. Could you please do the following thing:
1) rebase your branch onto `dev` and to the PR against `dev`
2) squash your commits into one and re-push (force push) - that should leave you with a single commit
Let me know if you have any problems with that.

donkeycar/management/base.py
```diff
@@ -173,11 +173,12 @@ def run(self, args):
         
         print("Finding your car's IP address...")
         cmd = "sudo nmap -sP " + ip + "/24 | awk '/^Nmap/{ip=$NF}/B8:27:EB/{print ip}'"
+        cmdRPi4 = "sudo nmap -sP " + ip + "/24 | awk '/^Nmap/{ip=$NF}/DC:A6:32/{print ip}'"
         print("Your car's ip address is:" )
         os.system(cmd)
```
DocGarbanzo, line 178: I see the old command is hanging and not returning on RPi4, whereas the new one works. Can you add a flag that only one gets run and enable the new command by default ?

DocGarbanzo (changes requested):

PR #715, opened 2020-12-29
replace ubuntu.yml with the mac.yml file.

wallarug:
Only change was that TensorFlow was being installed differently.


DocGarbanzo:
On Linux, TensorFlow >= 2.2.0 is available as conda package, whereas on Mac it is not, hence we use this here. I had a look, but it seems one cannot easily create a common conda yaml file that provides different logic based on the  platform. Hence for now we have to stick with (at least) two different versions.

PR #717, opened 2020-12-30
Update ubuntu.yml on master

sctse999:
Fix dependency issue again

For some reason h5py can't pin to 2.10.0 so falling back to make it the same as mac.yml


DocGarbanzo:
@sctse999 - you are not on latest master. Do we really need to pin h5py on linux? Can you do the PR against dev?

sctse999:
Closing this as we don't have `ubuntu.yml` now

PR #718, opened 2021-01-02
Improvements in tub and training to manage new datastore and new tub GUI

DocGarbanzo:
# Tub and training improvements and addition of tub management UI
## Tub improvements:
* Fixed manifest iterator to avoid to deep recursive function calls which segfaults on deleting too many records
* Created automatic addition of '_session_id' field in the Tub. This id resembles the naming convention we used to have for old-style tubs before: `YY-MM-DD_N`, where `N` is a running integer.
* The '_session_id' is automatically added when using the conversion script from old tubs to new, using the date of the conversion and the order of the input tubs
* Added test for the new tub/tubwriter behaviour

## Tub GUI development:
* Support starting the GUI from 'donkey ui' command
* Relevant settings are stored in `~/.donkeyrc` file which is editable, so users can add support for their own data.
tub_ui

## Training improvements:
* You can now add a filter in form of a function to `my_config.py` to only train for a subset of records. The function has to be given as the parameter `TRAIN_FILTER`. Here is an example.
```python
def filter_record(record: TubRecord):
   """
   Return True if you want the `record` to be included for training.
   """
   include = record['user/throttle'] > 0.4
   return include

TRAIN_FILTER = filter_record
```
* Improve training performance by using tf prefetch
* Default AI framework to 'tensorflow' in training

## Others:
* Update docs
* Update version number to 4.2.0
* Reduced run time of test_train using only 5 epochs and adjusting thresholds accordingly
* Reduced run time of torch tests using 3 epochs
* Minor updates in docstrings and switching to f-strings
* Fixed duplicated function in keras.py


donkeycar/parts/datastore_v2.py
```diff
@@ -233,7 +243,7 @@ def __init__(self, base_path, inputs=[], types=[], metadata=[],
         has_catalogs = False
 
         if self.manifest_path.exists():
-            self.seekeable = Seekable(self.manifest_path, read_only=self.read_only)
+            self.seekeable = Seekable(self.manifest_path, read_only=False)
```
tikurahul, line 246: Is the intention here to support deletes even during read only mode ?
Please add a comment stating that.

donkeycar/parts/datastore_v2.py
```diff
@@ -268,12 +282,19 @@ def delete_record(self, record_index):
         self.deleted_indexes.add(record_index)
         self._update_catalog_metadata(update=True)
 
+    def un_delete_record(self, record_index):
```
tikurahul, line 285: Rename method to `restore_record(...)`.

donkeycar/pipeline/training.py
```diff
@@ -76,8 +76,8 @@ def create_tf_data(self) -> tf.data.Dataset:
         return dataset.repeat().batch(self.batch_size)
 
 
-def train(cfg: Config, tub_paths: str, model: str, model_type: str) \
-        -> tf.keras.callbacks.History:
+def train(cfg: Config, tub_paths: str, model: str, model_type: str,
+          select: Optional[str] = None) -> tf.keras.callbacks.History:
```
tikurahul, line 80: This should not be a string at all. It should be a function or a lambda.

donkeycar/pipeline/types.py
```diff
@@ -70,22 +71,50 @@ class TubDataset(object):
     '''
 
     def __init__(self, config: Config, tub_paths: List[str],
-                 shuffle: bool = True) -> None:
+                 shuffle: bool = True, select: Optional[str] = None) -> None:
```
tikurahul, line 74: Same here, This should not be a string, and we should not have to parse it at all.

donkeycar/pipeline/types.py
```diff
@@ -70,22 +71,50 @@ class TubDataset(object):
...
+        print(msg)
         self.records.clear()
         for tub in self.tubs:
             for underlying in tub:
-                record = TubRecord(self.config, tub.base_path,
-                                   underlying=underlying)
-                self.records.append(record)
+                record = TubRecord(self.config, tub.base_path, underlying)
+                # eval string uses 'record' as variable so this parses
+                if not self.select or eval(self.select):
```
tikurahul, line 97: We should not be using `eval` for anything. Its extremely unsafe.

donkeycar/pipeline/types.py
```diff
@@ -70,22 +71,50 @@ class TubDataset(object):
...
+                    self.records.append(record)
 
         return train_test_split(self.records, shuffle=self.shuffle,
                                 test_size=(1. - self.config.TRAIN_TEST_SPLIT))
+
+
+def create_filter_string(filter_text: str, inputs: List[str],
```
tikurahul, line 104: This should be completely unnecessary.

donkeycar/parts/tub_v2.py
```diff
@@ -105,21 +110,44 @@ class TubWriter(object):
     """
     def __init__(self, base_path, inputs=[], types=[], metadata=[],
                  max_catalog_len=1000):
-        self.tub = Tub(base_path, inputs, types, metadata, max_catalog_len)
+        self.tub = Tub(base_path, inputs + ['session_id'], types + ['str'],
```
tikurahul, line 113: Lets take the inputs and types and store them in a `dict` in the beginning. That way we only need to manipulate the dict going forward.

donkeycar/parts/tub_v2.py
```diff
@@ -105,21 +110,44 @@ class TubWriter(object):
...
         return self.tub.manifest.current_index
 
     def __iter__(self):
         return self.tub.__iter__()
 
     def close(self):
-        self.tub.manifest.close()
+        self.tub.close()
+
+    def create_session_id(self):
```
tikurahul, line 136: I don't like this idea at all.

We should be able to derive the name of the last session from the manifest metadata and we should not have to read a tub record for that. So add that to the manifest metadata.

donkeycar/management/tub_gui.py
```diff
@@ -0,0 +1,598 @@
...
+import math
+import os
+from collections import namedtuple
+import tkinter as tk
+from tkinter import ttk, filedialog
+import time
+from threading import Thread, active_count
+from PIL import ImageTk, Image
+import pandas as pd
+import yaml
```
tikurahul, line 11: Why YAML :disappointed:  ?

Maybe try using `json` consistently ?

donkeycar/management/tub_gui.py
```diff
@@ -0,0 +1,598 @@
+#!/usr/bin/env python3
```
tikurahul, line 1: General comment: Reviewing UI code is quite hard. 

Especially given I see a few `row`s of UI being built with the implementation that looks similar enough to refactor. Also the calls to `decompose` some information stored in strings, I am not sure why its being done that way. 

I am okay with leaving it the way it is - but it might be pretty hard to maintain this going forward. OOC : Any reason why you did not choose to extend the web UI ?

tikurahul (changes requested):
This is awesome work :clap: :clap: 

I have a couple of comments about the changes to the `session` and the `select` expression. But those should be easy to get to. 

I am okay with checking in the UI bits. But ideally i would like to get things consistent. We already have 1 implementation to view and delete records (the Web UI). Its okay to use this, as long as we agree that we maintain 1 thing and  not 2 going forward.

docs/utility/donkey.md
```diff
@@ -68,9 +68,23 @@ donkey tubclean <folder containing tubs>
...
 ```bash
-donkey train --tub=<tub_path> [--config=<config.py>] [--model=<model path>] [--model_type=(linear|categorical|inferred)] 
+donkey train --tub=<tub_path> [--config=<config.py>] [--model=<model path>] 
+[--model_type=(linear|categorical|inferred)] [--filter=<expr>]
 ```
+* Uses the data from the `--tub` datastore
+* Uses the config file from the `--config` path (optionally)
+* Saves the model into `--model`
+* Uses the model type `--type`
+* Supports a logic expression to filter records, for example:
```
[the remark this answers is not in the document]
DocGarbanzo: Yes, agreed. We can support an entry in the `myconfig.py`file, which makes it easier to avoid misspellings, etc, and it doesn't clutter the interface.

donkeycar/management/base.py
```diff
@@ -432,24 +432,34 @@ def parse_args(self, args):
         parser.add_argument('--config', default='./config.py', help='location of config file to use. default: ./config.py')
         parser.add_argument('--framework', choices=['tensorflow', 'pytorch', None], required=False, help='the AI framework to use (tensorflow|pytorch). Defaults to config.DEFAULT_AI_FRAMEWORK')
         parser.add_argument('--checkpoint', type=str, help='location of checkpoint to resume training from')
+        parser.add_argument('--filter', default=None,
```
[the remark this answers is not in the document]
DocGarbanzo: Ok, see above.

donkeycar/parts/datastore_v2.py
DocGarbanzo: Yes, but we can rethink this. I see very little performance improvements by setting this, so let me remove that - actually I thought I took it out :-)

donkeycar/parts/datastore_v2.py
DocGarbanzo: Ok.

donkeycar/pipeline/training.py
DocGarbanzo: Yeah, see solution above.

donkeycar/pipeline/types.py
DocGarbanzo: Not sure I understand what you mean... the logic is, that you type 'user/throttle > 0.1' and the function converts this into `record.underlying['user/throttle'] > 0.1` which is then applied to the object `record`.

donkeycar/management/tub_gui.py
DocGarbanzo: The web UI was very hard to change. In my fork I had added some sub-functionality, like displaying more data  (`user/angle, user/throttle, car/speed, car/distance`, etc), in the html, but it was a trial and error exercise. I gave up on the idea to hook that up dynamically to the tub content - for example `car/speed, car/distance` is not something other people might have in their tubs. Hence the new UI dynamically reads the fields from the tub. All this functionality is easy in the new tub UI. Additionally handling large tubs is super performant. You will need a lot of patience if you want to edit a 300k tub in the web UI.  I spent too many hours in front of the web UI :-/

I agree that the code in the web UI can be better documented and I can do some small re-factorings and break it into smaller parts. For now, I thought we just offer this as an alternative. The `decompose` function is used for data stored in vectors, which we support in the tub. E.g.  my tubs store IMU data in two 3-d vectors instead of 6 scalars. The UI will support both and decompose the vectors and you can individually view only some components. 

I am also thinking of porting the UI to kivy, but this will only affect the look and not the functionality. Kivy is quite sleek, also on Linux, and uses graphics card OpenGL rendering.

donkeycar/pipeline/types.py
tikurahul: I am saying that the config file should contain something like:

```python
def filter_record(record: TubRecord):
   """
   Return True if you want the `record` to be included for training.
   """
    # defaults are to consider every record 
   return True

FILTER = filter_record
```

There is very little value in us parsing out strings to convert them to lambdas or invent a DSL.

donkeycar/management/tub_gui.py
DocGarbanzo: This is used only for storing the config file of the app. Json is a bit inconvenient to edit with all the `=` and `{`, etc,  and you might need to modify this file if you want a graphical progress bar for unknown data. So adding a row into this file is very easy. Compare it to a `.condarc` or similar file.

donkeycar/parts/tub_v2.py
DocGarbanzo: Makes sense.

donkeycar/parts/tub_v2.py
DocGarbanzo: Looking into this.

donkeycar/pipeline/types.py
DocGarbanzo: I haven't found an alternative that can parse expressions like `user/throttle > 0.1 and _index < 1000`. Apart from that, it is not used anywhere in the running vehicle, only in the selection of data for training, and it is non-destructive, it just omits records. You can test the expression in the UI, it shows you if it evaluates to true or false. It's neither put into a public web server - that wouldn't be a good idea, nor using an unchecked string as input. I don't think there is any more safety concern for malfunction than opening a python prompt on your machine and running the same code there in case anyone would put `shutil.rmtree('/')`.

donkeycar/pipeline/types.py
DocGarbanzo: Got it, yeah we can do this.

donkeycar/parts/tub_v2.py
tikurahul: Thinking about this a bit more, we should move all the `Tub` specific information to a file called `tub.manifest` which is a `Seekable` as well or just a standard `file` with some JSON content. That might be a good way to decouple it from the actual `catalog` itself. Because the actual `session` information is orthogonal.

donkeycar/parts/tub_v2.py
tikurahul: No, i meant the caller should include `session` and `str` in `inputs` and `types`. I feel like we are muddling the responsibilities a bit. The `run` method's responsibility should provide the `session_id` and not the other way around. The `Tub` should **not** have to discover the `session_id`. 

I think this will become easier when you store the `Tub` specific metadata into its own `manifest` file. See my comment below.

donkeycar/pipeline/types.py
tikurahul: I think if we switch to the actual lambda in the `config` then we don't need to do anything. (That is what i meant)

tikurahul (changes requested):

donkeycar/parts/tub_v2.py
DocGarbanzo: Yeah we could as well add a line to the `manifest.json` - because the only thing we need is a list of `session_id`s.

donkeycar/pipeline/types.py
DocGarbanzo: I've switched to that in training.

donkeycar/parts/datastore_v2.py
```diff
@@ -318,7 +344,36 @@ def _update_catalog_metadata(self, update=True):
         self.catalog_metadata = catalog_metadata
         self.seekeable.writeline(json.dumps(catalog_metadata))
 
+    def _write_metadata_contents(self):
+        self.seekeable.truncate_until_end(2)
+        self.seekeable.writeline(json.dumps(self.metadata))
+
+    def create_session_id(self):
```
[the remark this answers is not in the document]
DocGarbanzo: Ok - changed.

donkeycar/parts/datastore_v2.py
```diff
@@ -318,7 +344,36 @@ def _update_catalog_metadata(self, update=True):
...
+        """ Creates a new session id and appends it to the metadata."""
+        last_session_id = None
+        last_session_ids = self.manifest_metadata.get('session_ids')
+        if last_session_ids:
+            last_session_id = last_session_ids[-1]
+        last_num = -1
```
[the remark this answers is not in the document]
DocGarbanzo: Ok, makes sense, changed to storing a dictionary and avoid parsing.

donkeycar/parts/datastore_v2.py
```diff
@@ -230,50 +240,66 @@ def __init__(self, base_path, inputs=[], types=[], metadata=[],
...
+        new_catalog = self.current_index > 0 \
+                      and (self.current_index % self.max_len) == 0
         if new_catalog:
             self._add_catalog()
 
         self.current_catalog.write_record(record)
         self.current_index += 1
         # Update metadata to keep track of the last index
         self._update_catalog_metadata(update=True)
+        self._got_updated = True
```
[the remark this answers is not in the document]
DocGarbanzo: We use this flag to track if any records have been written into the tub. Only then we update the session information in the manifest metadata. Otherwise if you'd open an existing Tub, it would always append a new session_id. I changed the name.

donkeycar/tests/test_tub_v2.py
```diff
@@ -1,41 +1,70 @@
...
-            print('Record %s' % (record))
+            print(f'Record: {record}')
             count += 1
 
         self.assertEqual(count, (write_count - len(delete_indexes)))
         self.assertEqual(len(self.tub), (write_count - len(delete_indexes)))
 
-    def tearDown(self):
-        shutil.rmtree(self._path)
+    def test_basic_tubwriter_operations(self):
```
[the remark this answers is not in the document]
DocGarbanzo: Done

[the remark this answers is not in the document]
DocGarbanzo: Moved into own test.

donkeycar/parts/datastore_v2.py
```diff
@@ -280,7 +280,7 @@ def write_record(self, record):
         self.current_index += 1
         # Update metadata to keep track of the last index
         self._update_catalog_metadata(update=True)
-        self._got_updated = True
+        self._updated_session = True
```
tikurahul, line 283: This only needs to be happen when a new session is created right ?
Why are we doing this on every `write_record` ?

donkeycar/parts/datastore_v2.py
```diff
@@ -344,36 +344,27 @@ def _update_catalog_metadata(self, update=True):
...
-            self.manifest_metadata['session_ids'] = [session_id]
-        return session_id
+            sessions['all_full_ids'] = []
+        this_id = last_id + 1
+        date = time.strftime('%y-%m-%d')
+        this_full_id = date + '_' + str(this_id)
+        sessions['last_id'] = this_id
+        sessions['last_full_id'] = this_full_id
+        sessions['all_full_ids'].append(this_full_id)
+        self.manifest_metadata['sessions'] = sessions
```
tikurahul, line 361: This should be the only case where we update the session. Keep track of it and subsequently add it to the records when writing.

donkeycar/parts/datastore_v2.py
tikurahul: I meant something like:

```python
if not self._updated_session:
    # Ensure session metadata is persisted when the `Tub` is closed.
    self._updated_session = True
```

DocGarbanzo: If we set this to True when the tub is created, we would append the `_session_id` to the metadata, even if no record had been added. In fact we would only need to do this when `write_record()` is called once, but programatically, I don't think we'll gain any performance benefit of doing a check before that. But let me add this for clarity, I think that's what you were asking for.

donkeycar/parts/datastore_v2.py
DocGarbanzo: That's correct. We update the sessions in the metadata when creating the `_session_id`. However this gets only saved to the metadata file if any records were received.

sctse999:
Shall we merge this?

tikurahul (approved):

PR #731, opened 2021-01-10
Fixed erroneous comment hashes

zlite:
I made a mistake and put comment hashes in front of the required config lines. (I forgot that Donkey automatically adds them when creating the myconfig.py file). So this is just taking them out


tikurahul:
I think you accidentally rewrote git history again.

PR #751, opened 2021-01-18
Add back training history plot

sctse999:
This is copied from v3. We had this plot before.

### DIFF
--- a/donkeycar/parts/keras.py
+++ b/donkeycar/parts/keras.py
@@ -140,6 +140,40 @@ class KerasPilot(ABC):
             workers=1,
             use_multiprocessing=False
         )
+            
+        try:
+            import matplotlib.pyplot as plt
+            from pathlib import Path
+        
+            plt.figure(1)
+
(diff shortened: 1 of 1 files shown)


donkeycar/parts/keras.py
```diff
@@ -140,6 +140,40 @@ def train(self,
             workers=1,
             use_multiprocessing=False
         )
+            
+        try:
```
DocGarbanzo, line 144: We used to have a flag that enabled/disabled that part which I think is useful because the plot actually blocks the call. So if you run training in a batch it doesn't return until you click the matplot image away, for that we need to be able to suppress that.

DocGarbanzo (changes requested):

donkeycar/parts/keras.py
sctse999: Right. I have temporarily commented out `plt.show()`. Let's add it back later.

DocGarbanzo: Ah I see you're right.

PR #752, opened 2021-01-18
Pin tf to 2.2.0 in setup for pip and yaml for conda.

DocGarbanzo:
This solves the issue of getting error: `W tensorflow/core/kernels/data/generator_dataset_op.cc:103] Error occurred when finalizing GeneratorDataset iterator: Failed precondition: Python interpreter state is not initialized. The process may be terminated.` - this seems to occur for all versions >= 2.3.0.

### DIFF
--- a/install/envs/mac.yml
+++ b/install/envs/mac.yml
@@ -32,7 +32,7 @@ dependencies:
   - pytorch-lightning
   - numpy
   - pip:
-    - tensorflow>=2.2.0,<=2.3.2
+    - tensorflow=2.2.0
     - git+https://github.com/autorope/keras-vis.git
     - simple-pid
     - opencv-python-headless
--- a/install/envs/pc.yml
+++ b/install/envs/pc.yml
@@ -31,7 +31,7 @@ dependencies:
   - torchaudio
   - pytorch-lightning
   - numpy
-  - tensorflow>=2.2.0,<=2.3.2
+  - tensorflow=2.2.0
   - pip:
     - git+https://github.com/autorope/keras-vis.git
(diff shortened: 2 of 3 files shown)


tikurahul (approved):

custom-build-robots:
Hi, not sure if that helps you. Today I installed anaconda on a fesh new windows 10 laptop. The version of anaconda was the latest one with python 3.8. I tried to install tensorflow==2.1.0 to get rid of that tensorf flow error at the end of the training. But the installation was not possible because the error message told me that I have to use python < 3.8. Nor I am downgrading my anaconda installation to python==3.7. Hopefully I am now able to install tf 2.1.0 and to train my model with no error.

DocGarbanzo:
@custom-build-robots - the version of python that is in your anaconda base environment is irrelevant because you install donkey in an own environment. If you do the installation as per the doc with `conda env create -f install\envs\pc.yml` then this will create a conda env with py 3.7 and tf 2.2.0 which should work - at least it does on OSX and Ubuntu.

DocGarbanzo:
Sorry I have to patch this as I had the wrong operator `=` instead of `==` in the mac.yml file.

custom-build-robots:
> 
> 
> @custom-build-robots - the version of python that is in your anaconda base environment is irrelevant because you install donkey in an own environment. If you do the installation as per the doc with `conda env create -f install\envs\pc.yml` then this will create a conda env with py 3.7 and tf 2.2.0 which should work - at least it does on OSX and Ubuntu. Let us know the outcome on windows.

Sorry that was my fault. You are right.

PR #753, opened 2021-01-18
Support imgaug on nano and pi

DocGarbanzo:
Move imgaug from pc section to install_requires so it can be used on nano and pi as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,8 @@ setup(name='donkeycar',
           "simple_pid",
           'progress',
           'typing_extensions',
-          'pyfiglet'
+          'pyfiglet',
+          'imgaug',
       ],
       extras_require={
           'pi': [
@@ -65,7 +66,6 @@ setup(name='donkeycar',
           ],
           'pc': [
               'matplotlib',
-              'imgaug',
           ],
           'dev': [
               'pytest',


tikurahul:
`imgaug` needs a lot of native binaries, and take a long time to compile on the nano because of the lack of precompiled wheels. I don't recommend doing this.

DocGarbanzo:
> `imgaug` needs a lot of native binaries, and take a long time to compile on the nano because of the lack of precompiled wheels. I don't recommend doing this.

@tikurahul - we have now cropping and also the trapezoidal image modifications going through `imgaug`. If we train on such modified images, I would think we want to use the same image modification in the car.

sctse999:
How about moving this under `extras_require`?

tikurahul:
I feel like augmentation is not going to be attempted by a lot of users; and the users who know about augmentation know how to install `imgaug`. I leave it to you. I am fine either way.

DocGarbanzo:
I agree with @tikurahul - if the installation already causes lengthy compilations  (and even complications) for all users on nano and pi then we should avoid it. Using the `Augmentation` part in the car flow is not supported in any template, so people who want to use that should be fine to go through the necessary installations themselves. @sctse999 - moving from `pc` to `extras_require` is what I did, but I wasn't aware of the negative implications.

PR #762, opened 2021-01-22
Fix tf2 lite nano camera support

cloud-rocket:
Please suggest if there is a better place to put this fix. (Original `input_data` contains `float64` values, while tflite requires `float32`)


DocGarbanzo (approved):
My comments were more suggestions, no need to change, I'm approving this as your fix for `tflite` is important.

PR #766, opened 2021-01-24
new: convert prints to python logging

cloud-rocket:
The idea is to convert all "prints" into standard Python logging (which by default will look like "prints" in console but can be customized via config) and also sent to MQTT monitoring app.

This is a preliminary PR for review only.

Things to do:
* Convert all "prints" to `logger.info(`

### DIFF
--- a/donkeycar/templates/cfg_complete.py
+++ b/donkeycar/templates/cfg_complete.py
@@ -183,6 +183,11 @@ MM1_SHOW_STEERING_VALUE = False
 #  eg.'/dev/tty.usbmodemXXXXXX' and replace the port accordingly
 MM1_SERIAL_PORT = '/dev/ttyS0'  # Serial Port for reading and sending MM1 data.
 
+#LOGGING
+HAVE_CONSOLE_LOGGING = True
+LOGGING_LEVEL = 'INFO'          # (Python logging level) 'NOTSET' / 'DEBUG' / 'INFO' / 'WARNING' / 'ERROR' / 'FATAL' / 'CRITICAL'
+LOGGING_FORMAT = '%(message)s'  # (Python logging format - https://docs.python.org/3/library/logging.html#formatter-objects
+
 #TELEMETRY
 TELEMETRY_DONKEY_NAME = 'my_robot1234'
 TELEMETRY_PUBLISH_PERIOD = 1
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -17,6 +17,7 @@ Options:
 import os
 import time
 
+import logging
 from docopt import docopt
 import numpy as np
 
@@ -31,6 +32,9 @@ from donkeycar.parts.behavior import BehaviorPart
 from donkeycar.parts.file_watcher import FileWatcher
 from donkeycar.parts.launch import AiLaunch
 from donkeycar.utils import *
+
+
+logger = logging.getLogger()
 
 
 def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type='single', meta=[]):
@@ -60,7 +64,14 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type
     #Initialize car
     V = dk.vehicle.Vehicle()
 
-    print("cfg.CAMERA_TYPE", cfg.CAMERA_TYPE)
+    #Initialize logging before anything else to allow console logging
+    if cfg.HAVE_CONSOLE_LOGGING:
+        logger.setLevel(logging.getLevelName(cfg.LOGGING_LEVEL))
+        ch = logging.StreamHandler()
+        ch.setFormatter(logging.Formatter(cfg.LOGGING_FORMAT))
+        logger.addHandler(ch)
+
+    logger.info("cfg.CAMERA_TYPE %s"%cfg.CAMERA_TYPE)
     if camera_type == "stereo":
 
         if cfg.CAMERA_TYPE == "WEBCAM":
--- a/donkeycar/vehicle.py
+++ b/donkeycar/vehicle.py
@@ -8,11 +8,13 @@ Created on Sun Jun 25 10:44:24 2017
 
 import time
 import numpy as np
+import logging
 from threading import Thread
 from .memory import Memory
 from prettytable import PrettyTable
 import traceback
 
+logger = logging.getLogger(__name__)
 
 class PartProfiler:
     def __init__(self):
@@ -34,7 +36,7 @@ class PartProfiler:
         self.records[p]['times'][-1] = delta
 
     def report(self):
-        print("Part Profile Summary: (times in ms)")
+        logger.info("Part Profile Summary: (times in ms)")
         pt = PrettyTable()
         field_names = ["part", "max", "min", "avg"]
         pctile = [50, 90, 99, 99.9]
@@ -52,7 +54,7 @@ class PartProfiler:
                    "%.2f" % (sum(arr) / len(arr) * 1000)]
             row += ["%.2f" % (np.percentile(arr, p) * 1000) for p in pctile]
             pt.add_row(row)
-        print(pt)
+        logger.info(pt)
 
 
 class Vehicle:
@@ -89,7 +91,7 @@ class Vehicle:
         assert type(threaded) is bool, "threaded is not a boolean: %r" % threaded
 
         p = part
-        print('Adding part {}.'.format(p.__class__.__name__))
+        logger.info('Adding part {}.'.format(p.__class__.__name__))
         entry = {}
         entry['part'] = p
         entry['inputs'] = inputs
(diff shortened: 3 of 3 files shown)


donkeycar/parts/telemetry.py
```diff
@@ -38,7 +42,20 @@ def __init__(self, cfg, default_inputs=None, default_types=None):
...
-        print(f"Telemetry MQTT server connected (publishing: {', '.join(self._default_inputs)}")
+        if cfg.TELEMETRY_LOGGING_ENABLE:
+            self.setLevel(logging.getLevelName(cfg.TELEMETRY_LOGGING_LEVEL))
+            self.setFormatter(logging.Formatter(cfg.TELEMETRY_LOGGING_FORMAT))
+            logger.addHandler(self)
+
+    def add_step_inputs(self, inputs, types):
+   
+        # Add inputs if supported and not yet registered
+        for ind in range(0, len(inputs or [])):
```
DocGarbanzo, line 53: probably better to check `inputs` before the `for` statement

cloud-rocket: To check how? Not exactly understand the comment. 

(I changed the way it worked before to reduce the code in `manage.py` and make it more readable.)

DocGarbanzo: You only want to run this loop if `inputs` is not empty, i.e. `range(0, len([]))` is empty which is a bit convoluted way of saying `if inputs: .... else:...` or `if not inputs: return ...`

DocGarbanzo (commented):
@cloud-rocket - can you disentangle the changes for the mqtt part from the logger? Using proper logging instead of simple shell output might be worth considering, but it is not adding much value if we have this in one part only. As all other parts are still writing to the shell as usual, this results in a very inhomogeneous shell output. In fact DC used to have logging in early stages and was then converted to raw shell output (not sure why). I would suggest that if we go for proper logging, then we need to do this for the other parts as well.

cloud-rocket:
@DocGarbanzo,
I can surely split the changes to 2 different PRs (I converted Shell output in `vehicle.py` just to emphasize the ability to preserve the same output logs when `logging` is used):
* 1st one will deal with logging configuration and converting of the whole codebase into `logging`
* 2nd will update Telemetry as another logging exporter.
  
But if maintainers are not approving the whole conversion to `logging` - there is no much sense to update MqttTelemetry.

So I suggest to agree on this 1st.

DocGarbanzo:
Can you show a shell output of `manage.py drive ...` when we switch the logging on? Principally I think that using python logging is a good thing, but having a mixed shell output combining few `logging` statements with many `print` statements might be difficult to read.

tikurahul:
I am not sure if we actually want to redirect all `print`s to `log`s. If you want to introduce logging, and log the important bits, that is okay. I don't think this is actually a good idea.

DocGarbanzo:
I did a quick check myself. If one doesn't provide any `format` in the config the logging output is the same as the print output (I think it goes to `stderr`), so would be fine to turn it on w/o disturbing the output. Then we could turn more `print` into `logging` bit by bit. I see the benefit of sending the whole logging through mqtt or similar.

cloud-rocket:
@tikurahul , @DocGarbanzo - I added logging configuration to config.py (to follow donkey project config paradigm).

The default configuration does not introduces any changes to how the stdout looks like when donkey is loaded (even though part of the outputs are coming from logging and another part from `print`s. 

Using logging for all elements is beneficial (but not mandatory) since it will be possible to reroute the output to external systems:
* Telemetry
* Mobile apps (which are under development) - CC 
* Web interface
* Embedded screens

cloud-rocket:
> Can you show a shell output of `manage.py drive ...` when we switch the logging on? Principally I think that using python logging is a good thing, but having a mixed shell output combining few `logging` statements with many `print` statements might be difficult to read.

@DocGarbanzo this is how my output looks like:

("Adding part" stuff is using logging - I modified `Vehicle.py`, "Performance monitor activated." string is printed with `print` and it appears in the correct location)

```
$ python3 manage.py drive
________             ______                   _________
___  __ \_______________  /___________  __    __  ____/_____ ________
__  / / /  __ \_  __ \_  //_/  _ \_  / / /    _  /    _  __ `/_  ___/
_  /_/ // /_/ /  / / /  ,<  /  __/  /_/ /     / /___  / /_/ /_  /
/_____/ \____//_/ /_//_/|_| \___/_\__, /      \____/  \__,_/ /_/
                                 /____/

using donkey v4.1.0-dev ...
loading config file: /home/meir/mycar/config.py
loading personal config over-rides from myconfig.py
cfg.CAMERA_TYPE CSIC
Adding part CSICamera.
Starting My Serial Controller
Adding part MySerialController.
Adding part ThrottleFilter.
Adding part PilotCondition.
Adding part RecordTracker.
Adding part IMU.
Started Web FPV server. You can now go to jetson.local:8890 to view the car camera
Adding part WebFpv.
Adding part DriveMode.
Adding part AiLaunch.
Adding part AiRunCondition.
PWM Steering created
Init ESC
PWM Throttle created
Adding part PWMSteering.
Adding part PWMThrottle.
Performance monitor activated.
Adding part PerfMonitor.
Using catalog /home/meir/mycar/data/catalog_18.catalog
Adding part TubWriter.
Adding part MqttTelemetry.
Warming Serial Controller
Telemetry MQTT server connected (publishing: pilot/angle, pilot/throttle, recording, user/angle, user/throttle, user/mode, imu/acl_x, imu/acl_y, imu/acl_z, imu/gyr_x, imu/gyr_y, imu/gyr_z, perf/cpu, perf/mem, perf/freq)
Starting vehicle at 30 Hz
GST_ARGUS: Creating output stream
CONSUMER: Waiting until producer is connected...
GST_ARGUS: Available Sensor modes :
GST_ARGUS: 3264 x 2464 FR = 21.000000 fps Duration = 47619048 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: 3264 x 1848 FR = 28.000001 fps Duration = 35714284 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: 1920 x 1080 FR = 29.999999 fps Duration = 33333334 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: 1640 x 1232 FR = 29.999999 fps Duration = 33333334 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: 1280 x 720 FR = 59.999999 fps Duration = 16666667 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: 1280 x 720 FR = 120.000005 fps Duration = 8333333 ; Analog Gain range min 1.000000, max 10.625000; Exposure Range min 13000, max 683709000;

GST_ARGUS: Running with following settings:
   Camera index = 0
   Camera mode  = 5
   Output Stream W = 1280 H = 720
   seconds to Run    = 0
   Frame Rate = 120.000005
GST_ARGUS: Setup Complete, Starting captures for 0 seconds
GST_ARGUS: Starting repeat capture requests.
CONSUMER: Producer has connected; continuing.
CSICamera loaded.. .warming camera
```

cloud-rocket:
How do you suggest to proceed?

tikurahul:
Your branch needs to be rebased correctly.

cloud-rocket:
> Your branch needs to be rebased correctly.

done

donkeycar/parts/telemetry.py
cloud-rocket: I can add a check if you insist (personally I prefer to reduce the code if not necessary). Please let me know

cloud-rocket:
@tikurahul , @DocGarbanzo - which of the following options do you prefer?

1. Merge as-is
2. Convert all `prints` to log
3. Leave this pull request for logging changes only and separate MQTT Telemetry changes into separate PR

Please advise.

blown302:
> @tikurahul , @DocGarbanzo - which of the following options do you prefer?
> 
>     1. Merge as-is
> 
>     2. Convert all `prints` to logging (over the whole code)
> 
>     3. Leave this pull request for logging changes only and separate MQTT Telemetry changes into separate PR
> 
> 
> Please advise.

Implementing an actual logger will give the most benefits here. I've been away for awhile but this is something we've needed for some time. I haven't seen this telemetry app yet but I I'd hold off on the MQTT integration until the logger interface is in place for the whole codebase.

cloud-rocket:
@blown302 - the MQTT was already integrated before, all I did was adding the logging support to it. 

But I'll move it to another PR

cloud-rocket:
@blown302, @DocGarbanzo 

Removed all MQTT changes....

DocGarbanzo (approved):
I think this is a good start and we can convert more prints into logs in a couple of subsequent PRs.

Issue #690, opened 2020-11-29
Real execution frequency on embedded board is considerably lower than 20Hz and changes between drive and algo modes


cloud-rocket:
* My own setup runs at **8hz** on Jetson Nano in `drive` mode and drops to **3hz** in `local_angle` mode.
  * The difference is even bigger when disabling OLED part (20hz in `drive` mode vs 3.5zh in `local_angle` mode)
* Non of the CPU cores (4 on Nano) was fully loaded during the test
* Algorithm performance is probably suffers, because training and runtime is done on different frequency
* Performance was monitored via the following code - https://github.com/autorope/donkeycar/pull/689

This is my setup:
```
+--------------------+--------+--------+--------+--------+--------+--------+--------+
|        part        |  max   |  min   |  avg   |  50%   |  90%   |  99%   | 99.9%  |
+--------------------+--------+--------+--------+--------+--------+--------+--------+
|     CSICamera      |  0.09  |  0.01  |  0.02  |  0.01  |  0.02  |  0.05  |  0.09  |
| MySerialController |  0.09  |  0.02  |  0.02  |  0.02  |  0.03  |  0.06  |  0.08  |
|   ThrottleFilter   |  0.10  |  0.01  |  0.02  |  0.01  |  0.02  |  0.05  |  0.10  |
|   PilotCondition   |  0.08  |  0.01  |  0.01  |  0.01  |  0.02  |  0.04  |  0.07  |
|   RecordTracker    |  0.07  |  0.01  |  0.01  |  0.01  |  0.02  |  0.03  |  0.06  |
|        IMU         |  0.07  |  0.02  |  0.02  |  0.02  |  0.03  |  0.06  |  0.07  |
|       WebFpv       |  0.04  |  0.01  |  0.02  |  0.02  |  0.03  |  0.03  |  0.04  |
|    PerfMonitor     | 13.45  |  0.92  |  2.35  |  1.33  |  5.27  | 11.13  | 13.20  |
|     DriveMode      |  0.08  |  0.02  |  0.03  |  0.03  |  0.04  |  0.07  |  0.08  |
|      AiLaunch      |  0.04  |  0.02  |  0.02  |  0.02  |  0.02  |  0.04  |  0.04  |
|   AiRunCondition   |  0.05  |  0.01  |  0.02  |  0.01  |  0.02  |  0.02  |  0.05  |
[...]

PR #768, opened 2021-01-26
Fix OLED performance issue

fengye:
My config is an RPi 4 with Adafruit OLED 128x32(https://www.adafruit.com/product/3527) so I decided to use OLED part available. However, the performance of the part is poor and took about 170ms every the loop. Even I boosted the i2c baudrate to 1MHz in the RPi /boot/config.txt the average time for running OLED part is around 20ms. Looking at the code, it was intended to run in threaded mode, but it wasn't really. The update() function didn't create a loop thus it quits when it starts, while run_threaded() is calling update() directly, which results in a performance hit on the main loop.

The solution is just creating a loop in update() and let the thread do the heavy lifting job while the main loop only updates the statistics.


fengye:
The OLED part issue was also mentioned in issue #690

sctse999:
Just found below old PR not merged yet ... I believe a lot of effort has been put into that PR. Wish we can take a look of it too.

https://github.com/autorope/donkeycar/pull/512

fengye:
That PR seems dead though. I skimmed through the implementation, the performance is gained by queuing the OLED i2c request and process it once each call, which is still taking about 16ms on average on the author's device. Mine is 0.41ms on my Pi 4. I think it's worth to merge this one then see if we can do more about other models of OLED and generalisation.

```
Part Profile Summary: (times in ms)
+------------------------------+------+------+------+------+------+------+-------+
|             part             | max  | min  | avg  | 50%  | 90%  | 99%  | 99.9% |
+------------------------------+------+------+------+------+------+------+-------+
|           PiCamera           | 0.06 | 0.01 | 0.02 | 0.02 | 0.02 | 0.03 |  0.06 |
| PS3SwappedJoystickController | 0.05 | 0.02 | 0.02 | 0.02 | 0.03 | 0.04 |  0.05 |
|        ThrottleFilter        | 0.06 | 0.01 | 0.01 | 0.01 | 0.01 | 0.02 |  0.05 |
|            WebFpv            | 0.05 | 0.01 | 0.01 | 0.01 | 0.02 | 0.03 |  0.04 |
|           OLEDPart           | 0.66 | 0.43 | 0.47 | 0.45 | 0.52 | 0.61 |  0.65 |
|         LedHeadlight         | 3.18 | 0.20 | 0.41 | 0.39 | 0.47 | 0.53 |  2.62 |
|        PilotCondition        | 0.06 | 0.01 | 0.01 | 0.01 | 0.02 | 0.03 |  0.04 |
|          DriveMode           | 0.09 | 0.01 | 0.01 | 0.01 | 0.02 | 0.03 |  0.07 |
|         PWMSteering          | 2.99 | 0.35 | 0.76 | 0.70 | 0.91 | 2.24 |  2.98 |
|         PWMThrottle          | 5.95 | 0.33 | 1.02 | 0.95 | 1.07 | 2.95 |  5.27 |
+------------------------------+------+------+------+------+------+------+-------+

```

sctse999:
Ok i have tested this and it works as described. Merging it now.

PR #782, opened 2021-02-03
Add datastore metadata function and test

sctse999:


tikurahul (changes requested):
Whats the intended usecase ?

sctse999:
E.g. allowing user to put remarks on a dataset (we allow this on the app)

tikurahul:
Lets hold off on making this change. There are a few things I want to change before we land this.

sctse999:
Just wondering when this could get merged?

tikurahul:
Actually i thought more about this, and I don't want to add this feature to the `Tub` itself. This is because `metadata` is already a very overloaded term. 

You can keep this data in a separate file, and correlate it with the  `session_id`.

PR #790, opened 2021-02-18
readme.md_fixme_import_TubWriter_from_datastore_v2.py

stefanjay:
due to following the readme while trying to deploy donkeycar in an DGX env we faced the problem that the file ./datastore_v2.py is missing the `class TubWriter`. Please merge to fix the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ The donkey car is controlled by running a sequence of events
 import time
 from donkeycar import Vehicle
 from donkeycar.parts.cv import CvCam
-from donkeycar.parts.datastore_v2 import TubWriter
+# FIXME from donkeycar.parts.datastore_v2 import TubWriter => needs class TubWriter in ./datastore_v2.py
+from donkeycar.parts.datastore import TubWriter
 V = Vehicle()
 
 IMAGE_W = 160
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,8 @@ The donkey car is controlled by running a sequence of events
 import time
 from donkeycar import Vehicle
 from donkeycar.parts.cv import CvCam
-from donkeycar.parts.datastore_v2 import TubWriter
+# FIXME from donkeycar.parts.datastore_v2 import TubWriter => needs class TubWriter in ./datastore_v2.py
+from donkeycar.parts.datastore import TubWriter
```
DocGarbanzo, line 41: You are right, the import is not correct, but it should be from `donkeycar.parts.tub_v2` rather, can you change that and remove the #fixme, then we can merge this.

DocGarbanzo (changes requested):

stefanjay:
commited as requested by @DocGarbanzo

README.md
stefanjay: commited

DocGarbanzo:
Hi @stefanjay - I already corrected this on master and will also merge all doc changes from master back to dev very soon. Thanks for bringing this up. For changes to docs only you can do a PR against master in the future.

PR #793, opened 2021-02-24
Need to add "./" before running shell file

zlite:


tikurahul (approved):

zlite:
How do we make those bogus Travis checks go away?

PR #805, opened 2021-03-11
Add "./" before shell script so it will run

zlite:


docs/guide/robot_sbc/setup_raspberry_pi.md
```diff
@@ -221,7 +221,7 @@ git checkout master
 pip install -e .[pi]
 pip install numpy --upgrade
 wget "https://raw.githubusercontent.com/PINTO0309/Tensorflow-bin/master/tensorflow-2.3.1-cp37-none-linux_armv7l_download.sh"
-chmod u+x tensorflow-2.3.1-cp37-none-linux_armv7l_download.sh
+./chmod u+x tensorflow-2.3.1-cp37-none-linux_armv7l_download.sh
```
DocGarbanzo, line 224: I would expect you rather want `chmod u+x ./tensorflow-2.3.1-cp37-none-linux_armv7l_download.sh` because not everyone might have `.` in their PATH and `chmod` is a command.

zlite: Actually, that was me just typing too fast as I did this for the sixth time to try to get through Travis.  I've now revised to what I meant (modifying the line that is supposed to run the script, not the one setting that script to runable)

DocGarbanzo (approved):
LGTM

PR #806, opened 2021-03-13
Add automatic pushing of docs

DocGarbanzo:
Adding GitHub Action that automatically pushes docs to `docs.donkeycar.com` when the content of the `docs/` folder changes.

### DIFF
--- a/.github/linters/.python-black
+++ b/.github/linters/.python-black
@@ -1,0 +1,10 @@
+[tool.black]
+line-length = 80
+target-version = ['py37']
+multi_line_output = 3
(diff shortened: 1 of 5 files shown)


zlite (approved):
I don't quite get how the below commit is part of this, but it seems benign at least:

"### After a reboot, I don't see the (donkey) in front of the prompt, and I get python errors when I run?

DocGarbanzo:
@zlite - that was single line change to docs, want to check if this already triggers (probably triggers only after this is merged, let's see).

PR #810, opened 2021-03-14
Add missing checkout to auto push docs

DocGarbanzo:
This was missing in previous PR.


DocGarbanzo:
I can push these changes to master directly... I think that resolves it faster.

PR #821, opened 2021-03-26
[add] install guide command changes

wisehackermonkey:
Best practice for new users is to include moving to there home directory before installing the github
and updating the apt sources

### DIFF
--- a/docs/guide/robot_sbc/setup_raspberry_pi.md
+++ b/docs/guide/robot_sbc/setup_raspberry_pi.md
@@ -180,6 +180,7 @@ Choose `<Finish>` and hit enter.
 ## Step 8: Install Dependencies
 
 ```bash
+sudo apt-get update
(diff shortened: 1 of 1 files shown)


docs/guide/robot_sbc/setup_raspberry_pi.md
```diff
@@ -180,6 +180,7 @@ Choose `<Finish>` and hit enter.
 ## Step 8: Install Dependencies
 
 ```bash
+sudo apt-get update
```
DocGarbanzo, line 183: This command is already run in step 6.

docs/guide/robot_sbc/setup_raspberry_pi.md
```diff
@@ -215,6 +216,7 @@ cd projects
 * Get the latest donkeycar from Github.
 
 ```bash
+cd ~
```
DocGarbanzo, line 219: This is not necessary as you are in the home directory automatically after ssh'ing into the RPi.

PR #823, opened 2021-03-30
New Donkey UI

DocGarbanzo:
# New donkey app
ui-pilot-arena

* Migration of previous tub UI from `tkinter` to `kivy`- this is a x-platform open-gl supported python UI framework
* Extending the functionality of replacing the browser-based `donkey tub clean` with the following additional features
    - A pilot arena screen: here we can benchmark two pilots agains each other and compare the model steering/throttle forecast versus the recorded data. It also supports data augmentation so input images can be blurred and brightness tweaked
    - A trainer screen: here we can train a pilot using the select tub from the tub screen. Trained pilots are displayed in a database which is automatically updated when training finishes. There is a comment field in the UI that allows to put useful text for training into the database. The trainer screen also can overwrite the configuration fields.
    - Changes to the training routine: Now supports auto-generated model names like `pilot-21-03-29_1.h5` using the same YY-MM-DD_X pattern that is also used in the tub's session ID. Also, re-introduction of the transfer model, so training can start with a model that is already pre-trained. As models store the solver information in the file, the training will continue where it finished.
    - A car connector screen: this supports move tub data from the car to the PC and moving pilots back to the car. The app uses `ssh` and `rsync` as subprocesses here, hence this will not be working under windows out of the box. It's a more experimental feature anyway.

* Corresponding updates for config files (including minor bug fixes) and the documentation
* Improvement in the tub to delete/restore whole sets - this avoids cycling through single delete operations which are very slow as the `manifest.json` is re-written every time. Now there is only a single write process if a whole list is deleted or restored.


donkeycar/management/makemovie.py
```diff
@@ -80,30 +81,30 @@ def run(self, args, parser):
         clip = mpy.VideoClip(self.make_frame, duration=((num_frames - 1) / self.cfg.DRIVE_LOOP_HZ))
         clip.write_videofile(args.out, fps=self.cfg.DRIVE_LOOP_HZ)
 
-    def draw_user_input(self, record, img):
-        '''
-        Draw the user input as a green line on the image
-        '''
-
+    @staticmethod
+    def draw_line_into_image(angle, throttle, is_left, img, color):
```
DocGarbanzo, line 85: Factored out common code.

tikurahul:
Still looking at this. Sorry, it is taking a while.

donkeycar/pipeline/database.py
```diff
@@ -0,0 +1,83 @@
...
+    def generate_model_name(self) -> Tuple[str, int]:
+        if self.entries:
+            df = self.to_df()
+            # otherwise this will be a numpy int
+            last_num = int(df.index.max())
+            this_num = last_num + 1
+        else:
+            this_num = 0
+        date = time.strftime('%y-%m-%d')
+        name = 'pilot_' + date + '_' + str(this_num)
```
tikurahul, line 36: Consider using a formatted string consistently.

donkeycar/vehicle.py
```diff
@@ -54,7 +55,7 @@ def report(self):
                    "%.2f" % (sum(arr) / len(arr) * 1000)]
             row += ["%.2f" % (np.percentile(arr, p) * 1000) for p in pctile]
             pt.add_row(row)
-        logger.info(pt)
+        logger.info('\n' + str(pt))
```
tikurahul, line 58: A formatted string here as well.

tikurahul (approved):
This is awesome work ! :+1:

PR #824, opened 2021-03-31
Take out Tensorflow from automated installation

zlite:
……because it's causing an error with latest conda

I'll add it into the documentation to manually install with the --user flag after the conda evn is built


zlite:
I've now split pc.yml into two: windows.yml and ubuntu.yml

zlite:
Dirk, once you approve this I can update the docs for the Ubuntu installation to switch from pc.yml to ubuntu.yml

DocGarbanzo (approved):
Looks good, but we need to check that these files are also mentioned in the docs.

PR #829, opened 2021-03-31
adding support for L298N motor controller

raskaman:
adding logic to control the motors with a L298N
using the enable pins for pwm


DocGarbanzo (approved):
Looks good to me, thanks.

PR #830, opened 2021-04-01
Fix pytorch lightning error

EricWiener:
This resolves the issue @DocGarbanzo mentioned to me about PyTorch tests failing. All tests now pass.

The issue was caused when upgrading PyTorch Lightning from 1.1.0 to 1.2.4 when additional value checks were added in.


DocGarbanzo (approved):
LGTM - thanks.

PR #831, opened 2021-04-03
Some changes in the docs regarding the Sim API

Maximellerbach:
Updated the sim API in the doc,
more coming in the future


DocGarbanzo (approved):
Looks good to me.

PR #833, opened 2021-04-08
Added support for Odometry

zlite:
Supports both a simple rotary encoder and a quadrature encoder via an Arduino


donkeycar/parts/encoder.py
```diff
@@ -1,19 +1,69 @@
...
+            print(item)  # list all the serial ports
+        self.ser = serial.Serial('/dev/ttyACM0', 115200, 8, 'N', 1, timeout=0.1)
+        # initialize the odometer values
+        self.ser.write(str.encode('r'))  # restart the encoder to zero
+        self.ticks = 0
+        self.debug = debug
+        self.on = True
+
+    def update(self):
+        global lasttick
```
tikurahul, line 35: Please avoid the use of globals.

zlite: Fixed and pushed

zlite:
I fixed the error that was causing the CI fail. It was in the Basic Template, which is no longer used but is still included in the CI coverage

PR #840, opened 2021-04-12
Donkey UI: Add disclaimer and documentation for car connector screen

DocGarbanzo:
* Added disclaimer on the connector screen that only OSX/Ubuntu as supported and ssh needs to be set up correctly
*  Added ssh setup to the UI documentation


zlite:
Looks good and works fine. Thanks for updating the docs too, which is very helpful,

PR #851, opened 2021-04-25
Remove HAVE_ROBHAT option, which is no longer used

zlite:
It was redundant to the DRIVE_TRAIN_TYPE and  CONTROLLER_TYPE options


DocGarbanzo (approved):
LGTM.

PR #871, opened 2021-05-23
Update OLED part to use non-deprecated library and add rotation

zlite:
The Adafruit OLED library were were using has been deprecated and replaced by a CircuitPython-based version. This PR updates the OLED part to use the new library and adds a rotation option. 

Changes include: 

- Updated part
- Updated config file to add rotation option
- Updated install script to install CircuitPython dependencies

Tested on RPi.

### DIFF
--- a/donkeycar/parts/oled.py
+++ b/donkeycar/parts/oled.py
@@ -1,43 +1,54 @@
-import Adafruit_SSD1306
-
 from PIL import Image
 from PIL import ImageDraw
 from PIL import ImageFont
 import subprocess
 import time
+from board import SCL, SDA
+import busio
+from PIL import Image, ImageDraw, ImageFont
+import adafruit_ssd1306
 
 class OLEDDisplay(object):
     '''
     Manages drawing of text on the OLED display.
     '''
-    def __init__(self, bus_number=1):
+    def __init__(self, rotation=0):
         # Placeholder
         self._EMPTY = ''
         # Total number of lines of text
         self._SLOT_COUNT = 4
-        self.bus_number = bus_number
         self.slots = [self._EMPTY] * self._SLOT_COUNT
         self.display = None
+        self.rotation = rotation
 
     def init_display(self):
         '''
         Initializes the OLED display.
         '''
         if self.display is None:
-            # Use gpio = 1 to prevent platform auto-detection.
-            self.display = Adafruit_SSD1306.SSD1306_128_32(rst=None, i2c_bus=self.bus_number, gpio=1)
-            # Initialize Library
-            self.display.begin()
-            # Clear Display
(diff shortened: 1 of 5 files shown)


donkeycar/parts/oled.py
```diff
@@ -1,43 +1,54 @@
-import Adafruit_SSD1306
-
 from PIL import Image
 from PIL import ImageDraw
 from PIL import ImageFont
 import subprocess
 import time
+from board import SCL, SDA
+import busio
+from PIL import Image, ImageDraw, ImageFont
```
DocGarbanzo, line 8: Pls remove lines 1-3 otherwise these imports are duplicated.

install/pi/install.sh
```diff
@@ -40,8 +40,8 @@ source ~/env/bin/activate
 #make sure the virtual environment is active
 source ~/env/bin/activate
 
-# install pandas and numpy
+# install pandas and numpy and Adafruit CircuitPython
 pip install pandas #also installs numpy
-
+pip install adafruit-circuitpython-lis3dh
```
DocGarbanzo, line 45: Can we not install this with `pip` by adding it into the `.[pi]` section of `setup.py`? I don't think this install file is used.

DocGarbanzo (changes requested):

zlite:
Good catch on double imports. Other change made, too

donkeycar/parts/oled.py
```diff
@@ -1,43 +1,51 @@
...
+import adafruit_ssd1306
 
 class OLEDDisplay(object):
     '''
     Manages drawing of text on the OLED display.
     '''
-    def __init__(self, bus_number=1):
```
tikurahul, line 13: We need to provide the `I2C_BUSNO` on a Jetson Nano for this to work.

tikurahul (changes requested):

donkeycar/parts/oled.py
zlite: @tikurahul The way the Adafruit CircuitPython library framework works is that all those details, such as bus number, are automatically defined as part of the overall framework at installation time so they don't have to be redefined for each part. So there is no need (or way) to define a bus number in the new OLED library. 

That said, I have now added the installation of the CircuitPython framework to the overall Nano setup so people won't have to do it manually if they're using an OLED

DocGarbanzo (approved):
LGTM.

DocGarbanzo:
@tikurahul - Chris has addressed your issue, I'll approve & merge this.

donkeycar/parts/oled.py
tikurahul: That is awesome ! Thanks for clearing that up.

PR #872, opened 2021-05-24
Update Lidar part to use new Adafruit library

zlite:
The old Skoltech RPLidar library is depreciated and no longer supports latest RPLidar firmware


DocGarbanzo (approved):
LGTM.

PR #873, opened 2021-05-27
Revert to passing full path to model in training 

DocGarbanzo:
This got accidentally broken in 4.2 master. 
* Currently we have `donkey train --tub data --model pilot.h5` which will create the model in the directory `/model/pilot.h5`. 
* Previously we had explicitly give path to the models directory: `donkey train --tub data --model model/pilot.h5`. 

Here, we revert to the previous behaviour.


DocGarbanzo:
If this is merged, I will create a PR against master & do a patch release 4.2.1.

tikurahul (approved):

PR #877, opened 2021-06-01
Fix deletion of last records in tub

DocGarbanzo:
There is a bug in deleting the last n records of a tub. We fix it by correctly building the set of non-deleted indexes and removing the largest n entries from that.

### DIFF
--- a/donkeycar/parts/tub_v2.py
+++ b/donkeycar/parts/tub_v2.py
@@ -75,9 +75,11 @@ class Tub(object):
         self.manifest.delete_records(record_indexes)
 
     def delete_last_n_records(self, n):
-        last_index = self.manifest.current_index
-        first_index = max(last_index - n, 0)
-        self.manifest.delete_records(range(first_index, last_index))
+        # build ordered list of non-deleted indexes
+        all_alive_indexes = sorted(set(range(self.manifest.current_index))
+                                   - self.manifest.deleted_indexes)
+        to_delete_indexes = all_alive_indexes[-n:]
+        self.manifest.delete_records(to_delete_indexes)
 
     def restore_records(self, record_indexes):
         self.manifest.restore_records(record_indexes)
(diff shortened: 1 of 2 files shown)


donkeycar/tests/test_tub_v2.py
```diff
@@ -7,11 +7,12 @@
 
 class TestTub(unittest.TestCase):
 
-    def setUp(self):
-        self._path = tempfile.mkdtemp()
+    @classmethod
+    def setUpClass(cls) -> None:
```
tikurahul, line 11: Out of curiosity, why did you convert this to a `@classmethod` and not leave it as-is. Not that it matters, but I just wanted to understand if there is a reason.

donkeycar/parts/tub_v2.py
```diff
@@ -75,9 +75,11 @@ def delete_records(self, record_indexes):
         self.manifest.delete_records(record_indexes)
 
     def delete_last_n_records(self, n):
-        last_index = self.manifest.current_index
-        first_index = max(last_index - n, 0)
-        self.manifest.delete_records(range(first_index, last_index))
+        # build ordered list of non-deleted indexes
+        all_alive_indexes = sorted(set(range(self.manifest.current_index))
```
tikurahul, line 79: nit: call this `live_indexes`. It reads better.

tikurahul (approved):

donkeycar/tests/test_tub_v2.py
DocGarbanzo: Because `setUp`  and `tearDown` are called before/after each test function in the class. But here I want to keep the tub alive, because we fill it in the first test method and then use it in the second. `setUpClass` and `tearDownClass` are only called before the first test method and after the last one.

donkeycar/parts/tub_v2.py
DocGarbanzo: Yeah, will probably do this as part of next change, so I can merge this now.

PR #884, opened 2021-06-13
Add support for all models

DocGarbanzo:
# Add back support for all models

## Overview
* Factor out model from interpreter so each model can be saved and run in .h5 / .savedmodel keras format, in tflite format or in tensorrt format
* Improved KerasPilot interfaces to support the above
* Added test for all models and all interpreters (TFflite excluded for 3d CNN and TRT excluded for LSTM as these are not supported respectively)

## UI improvements
* Add support for all model / interpreter combinations in training and model arena
* Add options to sync only h5, savedmodel, tflite or tensorrt pilots to car

## Small improvements / fixes
* Added comment to status in train screen
* Added better error message in trt conversion
* Added seq length to lstm model str
* Added shebang to profile.py
* In UI: fixed disabling of keyboard when switching from one screen to another.

## Change on models
* Convert LSTM and 3D-CNN model from Sequential to Functional form, because transfer learning, i.e. continuing to train an already trained model, only works in the functional form of these two models.
* Added new model KerasLinearMemory which reads in the last n actions, this facilitates smoother steering
* Added support to inference from dictionary - passing by dictionary is used in training and supports augmentation
* Renamed output layers alphabetically in order to avoid tflite bug which returns output tensors in that order

## General code changes
* Add Sequencer class which returns list of continuous records so sequence models, like rnn, 3dcnn, memory can be trained. Add test.
* Add comparison string (called CString) class to better manage printing available user options. Used in get_model_by_type.
* Reduced tolerance on keras test as we still see relative differences of 2e-5 (which is ok).
* Check either relative or absolute tolerance as untrained models might return values >> 1 in keras test
* Added config parameters CREATE_TF_LITE, CREATE_TENSOR_RT to switch automatic creation of tflite and tensorrt models.


donkeycar/management/graph.py
```diff
@@ -100,8 +100,10 @@ def get_x(self, num_points):
         return x_trafo
 
     def transform_y(self, y):
-        y_scale = (self.bounding_box[1][1] - self.bounding_box[0][1]) \
-                  / (y.max() - y.min())
+        denom = (y.max() - y.min())
```
tikurahul, line 103: Just add a very small epsilon value (something like `1e-6` so you don't have have a divide by zero case.

donkeycar/parts/interpreter.py
```diff
@@ -0,0 +1,296 @@
...
+                                                 tags=[tag_constants.SERVING])
+        graph_func = saved_model_loaded.signatures[
+            signature_constants.DEFAULT_SERVING_SIGNATURE_DEF_KEY]
+        self.frozen_func = convert_var_to_const(graph_func)
+        self.input_shapes = [inp.shape for inp in graph_func.inputs]
+
+    def predict(self, img_arr: np.ndarray, other_arr: np.ndarray) \
+            -> Sequence[Union[float, np.ndarray]]:
+        # first reshape as usual
+        img_arr = img_arr.reshape((1,) + img_arr.shape).astype(np.float32)
```
tikurahul, line 262: Use `unsqueeze()` here and above. Its a lot more easier to understand. Also a comment would be super useful.

donkeycar/pipeline/types.py
```diff
@@ -71,27 +78,68 @@ class TubDataset(object):
     """
 
     def __init__(self, config: Config, tub_paths: List[str],
-                 shuffle: bool = True) -> None:
+                 seq_size: int = None) -> None:
```
tikurahul, line 81: `None` is never a good default for an `int`. Either use `Option[int]` or just use a number like `-1` to represent a sentinel value.

donkeycar/pipeline/types.py
```diff
@@ -71,27 +78,68 @@ class TubDataset(object):
...
+                    record = TubRecord(self.config, tub.base_path, underlying)
+                    if not self.train_filter or self.train_filter(record):
+                        self.records.append(record)
+            if self.seq_size:
+                seq = Sequencer(self.seq_size, self.records)
+                self.records = list(seq)
+        return self.records
+
+
+class Sequencer(Iterable[List[TubRecord]]):
```
tikurahul, line 106: Consider calling this `Collator` because that is what other ML frameworks use.

donkeycar/utils.py
```diff
@@ -16,23 +16,39 @@
...
 
+logger = logging.getLogger()
 
 '''
 IMAGES
 '''
 ONE_BYTE_SCALE = 1.0 / 255.0
 
 
+class CString:
```
tikurahul, line 33: Nit: Might want to come up with a better name ?

donkeycar/parts/interpreter.py
```diff
@@ -0,0 +1,296 @@
...
+        return [inp.shape for inp in self.model.inputs]
+
+    def compile(self, **kwargs):
+        assert self.model, 'Model not set'
+        self.model.compile(**kwargs)
+
+    def invoke(self, inputs):
+        outputs = self.model(inputs, training=False)
+        # for functional models the output here is a list
+        if type(outputs) is list:
```
tikurahul, line 135: Prefer `squeeze` instead.

donkeycar/parts/interpreter.py
```diff
@@ -0,0 +1,296 @@
...
+            # the additional dimension here again
+            output = [output[0].numpy() for output in outputs]
+            return output
+        # for sequential models the output shape is (1, n) with n = output dim
+        else:
+            return outputs[0].numpy()
+
+    def predict(self, img_arr: np.ndarray, other_arr: np.ndarray) \
+            -> Sequence[Union[float, np.ndarray]]:
+        img_arr = img_arr.reshape((1,) + img_arr.shape)
```
tikurahul, line 146: Use `unsqueeze` instead

donkeycar/parts/interpreter.py
```diff
@@ -0,0 +1,296 @@
...
+        inputs = img_arr
+        if other_arr is not None:
+            other_arr = other_arr.reshape((1,) + other_arr.shape)
+            inputs = [img_arr, other_arr]
+        return self.invoke(inputs)
+
+    def predict_from_dict(self, input_dict):
+        for k, v in input_dict.items():
+            input_dict[k] = v.reshape((1,) + v.shape)
```
tikurahul, line 155: Same as above..

donkeycar/pipeline/types.py
```diff
@@ -71,27 +78,68 @@ class TubDataset(object):
...
-        for tub in self.tubs:
-            for underlying in tub:
-                record = TubRecord(self.config, tub.base_path, underlying)
-                if not self.train_filter or self.train_filter(record):
-                    self.records.append(record)
-
-        return train_test_split(self.records, shuffle=self.shuffle,
-                                test_size=(1. - self.config.TRAIN_TEST_SPLIT))
+    @staticmethod
+    def is_continuous(rec_1: TubRecord, rec_2: TubRecord) -> bool:
```
tikurahul, line 117: I think we use `@classmethod` over `@staticmethod`. Use a `cls` as the first parameter if switching over.

tikurahul:
Also wanted to say, this is really amazing work !

Thanks for doing this.

donkeycar/parts/interpreter.py
DocGarbanzo: done.

donkeycar/parts/interpreter.py
DocGarbanzo: done.

donkeycar/parts/interpreter.py
DocGarbanzo: done.

donkeycar/pipeline/types.py
DocGarbanzo: Ok, we can use `0` here for a non-defined value, as only positive integers make sense. Changed to that.

donkeycar/pipeline/types.py
DocGarbanzo: Done.

donkeycar/pipeline/types.py
DocGarbanzo: I think, here '@staticmethod' makes more sense, because it doesn't need a class, it's just a helper function that is used on two `TubRecords`. Because it is used in the implementation of `__iter__()`.  Alternatively, I could make it a member function on `TubRecord` taking another `TubRecord`, that would make sense, too.

donkeycar/utils.py
DocGarbanzo: Yep, done.

DocGarbanzo:
@tikurahul - I've addressed all your points above and pushed a 3rd commit with these changes for review. I can rebase and squash everything afterwards (or squash merge the 3 commits), let me know your prefs.

donkeycar/utils.py
```diff
@@ -30,7 +30,7 @@
 ONE_BYTE_SCALE = 1.0 / 255.0
 
 
-class CString:
+class ComparisonString:
```
tikurahul, line 33: Nit: Call this `EqMemorizedString` perhaps.

tikurahul (approved):

PR #886, opened 2021-06-21
fix saliency map bugs

BillyCheung10botics:
This PR fixed the following two bugs:
- **tf2 version incomplatibility**
First,  the codes constructing saliency models  were incompatible with the tensorflow version 2. 
https://github.com/autorope/donkeycar/blob/20bf83c4f8bfe24bda151659e271a488bfa91f1b/donkeycar/management/makemovie.py#L186-L196
We should use the `GradientTape()` of tensorflow to calculate the gradient.

- **uint8 conversion bug**
Second, the conversion of RGB values from float (0.0, 1.0) to int (0, 255) was wrong, 
https://github.com/autorope/donkeycar/blob/20bf83c4f8bfe24bda151659e271a488bfa91f1b/donkeycar/management/makemovie.py#L239-L240
which creates artifacts as the bottom left figure shown. The red/yellow/blue parts are artifacts with RGB values over 255. Since some float values of the `image` array can be great than 1.0. I use `cv2.normalize` to properly convert the RGB values from float to uint. The correct saliency map is represented as the yello-purple clould as the right figure shown.
wrong RGB conversion
(left): video with artifacts  (right): video with correct saliency mask

### DIFF
--- a/donkeycar/management/makemovie.py
+++ b/donkeycar/management/makemovie.py
@@ -1,13 +1,11 @@
 import moviepy.editor as mpy
 from tensorflow.python.keras import activations
 from tensorflow.python.keras import backend as K
-
+import tensorflow as tf
+import cv2
+from matplotlib import cm
 try:
-    from vis.visualization import visualize_saliency, overlay
     from vis.utils import utils
-    from vis.backprop_modifiers import get
-    from vis.losses import ActivationMaximization
-    from vis.optimizer import Optimizer
 except:
     raise Exception("Please install keras-vis: pip install git+https://github.com/autorope/keras-vis.git")
 
@@ -105,7 +103,7 @@ class MakeMovie(object):
         user_throttle = float(record["user/throttle"])
         green = (0, 255, 0)
         self.draw_line_into_image(user_angle, user_throttle, False, img, green)
-        
+          
     def draw_model_prediction(self, img):
         """
         query the model for it's prediction, draw the predictions
@@ -183,24 +181,27 @@ class MakeMovie(object):
         model.layers[layer_idx].activation = activations.linear
         # build salient model and optimizer
         sal_model = utils.apply_modifications(model)
-        modifier_fn = get('guided')
-        sal_model_mod = modifier_fn(sal_model)
-        losses = [
-            (ActivationMaximization(sal_model_mod.layers[layer_idx], None), -1)
-        ]
-        self.opt = Optimizer(sal_model_mod.input, losses, norm_grads=False)
+        self.sal_model = sal_model
         return True
 
     def compute_visualisation_mask(self, img):
-        grad_modifier = 'absolute'
-        grads = self.opt.minimize(seed_input=img, max_iter=1, grad_modifier=grad_modifier, verbose=False)[1]
+        img = img.reshape((1,) + img.shape)
+        images = tf.Variable(img, dtype=float)
+        with tf.GradientTape() as tape:
+            pred = self.sal_model(images, training=False)
+            loss = 0
+            for p in pred:
+                loss = loss + p
+        grads = tape.gradient(loss, images)
+        grads = tf.math.abs(grads)
+
         channel_idx = 1 if K.image_data_format() == 'channels_first' else -1
-        grads = np.max(grads, axis=channel_idx)
+        grads = np.sum(grads, axis=channel_idx)
         res = utils.normalize(grads)[0]
         return res
 
     def draw_salient(self, img):
-        import cv2
+
         alpha = 0.004
         beta = 1.0 - alpha
         expected = self.keras_part.model.inputs[0].shape[1:]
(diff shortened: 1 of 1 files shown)


donkeycar/management/makemovie.py
```diff
@@ -236,19 +236,18 @@ def make_frame(self, t):
 
         if self.do_salient:
             image = self.draw_salient(image)
-            image = image * 255
-            image = image.astype('uint8')
-        
+            image = cv2.normalize(src=image, dst=None, alpha=0, beta=255, norm_type=cv2.NORM_MINMAX, dtype=cv2.CV_8U)
```
DocGarbanzo, line 239: Just a comment: I don't think there is a material difference between upscaling the [0, 1.0] image to [0, 255] integer, except probably the rounding.

donkeycar/management/makemovie.py
```diff
@@ -213,10 +214,9 @@ def draw_salient(self, img):
 
         norm_img = normalize_image(img)
         salient_mask = self.compute_visualisation_mask(norm_img)
-        z = np.zeros_like(salient_mask)
-        salient_mask_stacked = np.dstack((z, z))
-        salient_mask_stacked = np.dstack((salient_mask_stacked, salient_mask))
-        blend = cv2.addWeighted(img.astype('float32'), alpha, salient_mask_stacked, beta, 0.0)
+        salient_mask_stacked = cm.inferno(salient_mask)[:,:,0:3]
+        salient_mask_stacked = cv2.GaussianBlur(salient_mask_stacked,(3,3),cv2.BORDER_DEFAULT)
```
DocGarbanzo, line 218: Is using the Gaussian Blur here the reason we are not seeing the respective pixels of the mask as color variations?

DocGarbanzo (changes requested):
This looks good, but I'm wondering if the Gaussian Blur is the source of having the salient mask showing much less localised impact than in the original algorithm. Could you make this switchable, so the user could choose between the 'old' version of overlay vs the new one?
It seems you have done this completely without using the `keras-viz` package, that is great. From what I recall that was using a single step of backprop to utilise a single gradient calculation, but it seems this can be done much better with using the keras gradient (or `GradientTape` in tf 2.0) directly between the input and output of the salient-adjusted model.  Can you confirm?

donkeycar/management/makemovie.py
```diff
@@ -1,13 +1,11 @@
 import moviepy.editor as mpy
 from tensorflow.python.keras import activations
 from tensorflow.python.keras import backend as K
-
+import tensorflow as tf
+import cv2
+from matplotlib import cm
 try:
```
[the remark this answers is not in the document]
BillyCheung10botics: Thank you so much for reviewing. Yes this can be done without `keras-viz`! And we can definitely remove the try/catch. Some codes use the utilities from `keras.vis` which are `utils.normalize` and `utils.apply_modifications`, but I will implement them directly without importing them from `keras.vis` (they are just simple methods).

donkeycar/management/makemovie.py
BillyCheung10botics: Unfortunately some pixels of the `image` contain RGB values over 1.0 due the blending of `cv2.addWeighted` https://github.com/autorope/donkeycar/blob/20bf83c4f8bfe24bda151659e271a488bfa91f1b/donkeycar/management/makemovie.py#L219-L220
With the original approach, https://github.com/autorope/donkeycar/blob/20bf83c4f8bfe24bda151659e271a488bfa91f1b/donkeycar/management/makemovie.py#L239-L240
the multiplication by 255 produces pixels with values over 255. Then the values are modulo 255 under the uint8 conversion of numpy (`image.astype('uint8')`). For example, a float value of 1.1 is multipled to (1.2*255 = 306), then it is converted to a uint8 value of (306 mod 255 = 51). It is better to use `cv2.normalize` since it can always normalize the pixels' RGB correctly regardless of whether some pixels exceed the range.

BillyCheung10botics: For example, a float value of 1.2 is multipled to (1.2*255 = 306), then it is converted to a uint8 value of (306 mod 255 = 51).

**typo

donkeycar/management/makemovie.py
BillyCheung10botics: The `cv2.GaussianBlur` is to smooth the saliency map and hence gives a betterlooking visualization. 

I would like to clarify two points.
- First of all, please note that the original codes represent the saliency map with blue color as the following lines shown https://github.com/autorope/donkeycar/blob/20bf83c4f8bfe24bda151659e271a488bfa91f1b/donkeycar/management/makemovie.py#L216-L218 but I represent it using a colormap `inferno` (black-purple-yellow) https://github.com/robocarstore/donkeycar/blob/d5226637b668df534c680628e8c040b66e60d22a/donkeycar/management/makemovie.py#L217 . saliency compare
(left): saliency represented by inferno colormap (right): saliency represented by blue color

- Secondly, the sharp red, yellow, blue colors of the video produced by the original codes are not saliency map but artifacts due to the overflow of numpy uint8 conversion. 
Drawing (1)

DocGarbanzo: Ok, that was a severe bug then - so the original salient wasn't highlighting the activations but just showing overflowing bright pixels. Great that you discovered that. That's a big improvement.

donkeycar/management/makemovie.py
DocGarbanzo: Makes sense, thanks for explaining.

donkeycar/management/makemovie.py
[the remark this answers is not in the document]
DocGarbanzo: You're right, I missed the `utils` module. We can have a go at that in a separate change. Your help would be appreciated!

DocGarbanzo (approved):
Great change, many thanks.

PR #897, opened 2021-07-13
Added fps counter part

JoeSiu:
An fps counter that shows the current fps while running, and shows the fps list when shutdown.

PR #898, opened 2021-07-13
Added info overlay part

JoeSiu:
a simple Info overlay that adds texts overlay to the camera images, might be useful for debugging. It can show the current fps if enabled (#897) as well as the current throttle and angle, or any other info by adding the inputs or by calling the 'add' function in InfoOverlayLogger.


DocGarbanzo:
@JoeSiu - thanks for your PR. Why would you want to put this data into the images, as they are used for training. This would distort the ability to learn. You have another PR where the car frequency data gets logged. This is more useful I think.

JoeSiu:
This part is mainly used for easier debugging. For example, when testing the stop sign detection, It's easier to see the current pilot's throttle as well as the bounding box for the stop sign directly on the camera image while using the mobile app or web controller, instead of checking the output from the shell in a separate window which I think is less convenient. It will indeed affect the training through which I didn't consider at first, sorry about that!

donkeycar/parts/info_overlay.py
```diff
@@ -0,0 +1,75 @@
+import cv2
+
+info_list = []
```
DocGarbanzo, line 3: Please don't use global variables.

donkeycar/parts/info_overlay.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class InfoOverlayLogger(object):
```
DocGarbanzo, line 6: Please combine both classes into one, this should be one part only.

donkeycar/parts/info_overlay.py
```diff
@@ -0,0 +1,75 @@
...
+        self.add("current fps = {}".format(fps))
+
+        # truncate input so it won't cover up the screen
+        user_throttle = self.truncate(user_throttle)
+        user_angle = self.truncate(user_angle)
+        pilot_throttle = self.truncate(pilot_throttle)
+        pilot_angle = self.truncate(pilot_angle)
+
+        if user_mode == "user":
+            self.add("user throttle = {}".format(user_throttle))
```
DocGarbanzo, line 29: Pls use f-strings.

donkeycar/parts/info_overlay.py
```diff
@@ -0,0 +1,75 @@
...
+    def add(self, string):
+        info_list.append(string)
+
+    def truncate(self, input):
+        if input is not None:
+            input = str(input)[:4]
+        return input
+
+    def run(self, fps, user_mode, user_throttle, user_angle, pilot_throttle, pilot_angle):
+        self.add("current fps = {}".format(fps))
```
DocGarbanzo, line 20: Pls use f-strings.

donkeycar/templates/complete.py
```diff
@@ -186,6 +186,11 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None,
             
         V.add(cam, inputs=inputs, outputs=outputs, threaded=threaded)
 
+    if cfg.INFO_OVERLAY:
+        from donkeycar.parts.info_overlay import InfoOverlayLogger, InfoOverlayWritter
+        V.add(InfoOverlayLogger(), inputs=['fps/current', 'user/mode', 'user/throttle', 'user/angle', 'pilot/throttle', 'pilot/angle'], outputs=['info/info_list'])
+        V.add(InfoOverlayWritter(cfg.IMAGE_W, cfg.IMAGE_H), inputs=['cam/image_array'], outputs=['cam/image_array'])
```
DocGarbanzo, line 192: You can dispatch the different images to different operations. So the original image goes into the tub and the overlay image goes into the web server or mobile app, i.e. your output here becomes `'overlay/image_array'` and you use that accordingly.

DocGarbanzo (changes requested):
We can do this, pls have a look at the comments.

JoeSiu:
@DocGarbanzo I have rebased the branch, thank you!

DocGarbanzo (commented):
Sorry, I have a couple of only minor comments, can you have a look at them?

donkeycar/parts/info_overlay.py
```diff
@@ -0,0 +1,73 @@
+import cv2
+
+
+class InfoOverlayer(object):
+    """
+    Add a info overlay to the camera image
+    """
+
+    # Store the infos which will be write to image
+    info_list = []
```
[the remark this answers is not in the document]
JoeSiu: @DocGarbanzo It is used so that info can be written directly from other parts like this and have a shared `info_list`
```
from donkeycar.parts.info_overlay import InfoOverlayer
InfoOverlayer().add("hello")
```
"hello" can then be written onto the image along with other info like throttle or angle. However, it is supposed to be used as a temporary method in this case for easier debugging.

PR #899, opened 2021-07-13
Updated stop sign detector part to allow reversing for few iterations

JoeSiu:
Added a feature to force reverse the car for few iterations when it detected a stop sign, then stop the car until the stop sign is undetected. As when the car is driving at high speed, the stop sign could already be passed and is out of view before the car stops, which makes the car drive again. Therefore this reverse feature is added.

### DIFF
--- a/donkeycar/parts/object_detector/stop_sign_detector.py
+++ b/donkeycar/parts/object_detector/stop_sign_detector.py
@@ -23,7 +23,7 @@ class StopSignDetector(object):
         if not os.path.isfile(filename):
             urllib.request.urlretrieve(url, filename)
 
-    def __init__(self, min_score, show_bounding_box, debug=False):
+    def __init__(self, min_score, show_bounding_box, max_reverse_count=0, reverse_throttle=-0.5, debug=False):
         MODEL_FILE_NAME = "ssd_mobilenet_v2_coco_quant_postprocess_edgetpu.tflite"
         LABEL_FILE_NAME = "coco_labels.txt"
 
@@ -41,6 +41,12 @@ class StopSignDetector(object):
         self.min_score = min_score
         self.show_bounding_box = show_bounding_box
         self.debug = debug
+
+        # reverse throttle related
+        self.max_reverse_count = max_reverse_count
+        self.reverse_count = max_reverse_count
+        self.reverse_throttle = reverse_throttle
+        self.is_reversing = False
 
     def convertImageArrayToPILImage(self, img_arr):
         img = Image.fromarray(img_arr.astype('uint8'), 'RGB')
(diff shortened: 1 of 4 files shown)


donkeycar/parts/throttle_filter.py
```diff
@@ -10,15 +10,16 @@ def __init__(self):
 
     def run(self, throttle_in):
         throttle_out = throttle_in
+        # In case pilot's throttle and angle is None type
+        if throttle_in != None:
```
DocGarbanzo, line 14: This shouldn't be checked here, `throttle_in` is expected to be a number. Do you see a `None` being produced by the pilot?

donkeycar/parts/object_detector/stop_sign_detector.py
```diff
@@ -107,9 +112,20 @@ def run(self, img_arr, throttle, debug=False):
...
+        if traffic_light_obj or self.is_reversing:
+            if self.show_bounding_box and traffic_light_obj != None:
                 self.draw_bounding_box(traffic_light_obj, img_arr)
-            return 0, img_arr
+            
+            # Set the throttle to reverse within the max reverse count when detected the traffic light object
+            if self.reverse_count < self.max_reverse_count:
+                self.is_reversing = True
+                self.reverse_count += 1
+                print(f"reversing, reverseCount = {self.reverse_count}/{self.max_reverse_count}")
```
DocGarbanzo, line 123: We shouldn't be printing into the shell in each loop.

donkeycar/parts/object_detector/stop_sign_detector.py
```diff
@@ -107,9 +112,20 @@ def run(self, img_arr, throttle, debug=False):
...
+                return -1, img_arr
```
DocGarbanzo, line 124: This means full throttle backward. Shouldn't we go for a lower speed?

donkeycar/parts/object_detector/stop_sign_detector.py
```diff
@@ -23,7 +23,7 @@ def download_file(self, url, filename):
         if not os.path.isfile(filename):
             urllib.request.urlretrieve(url, filename)
 
-    def __init__(self, min_score, show_bounding_box, debug=False):
+    def __init__(self, min_score, show_bounding_box, max_reverse_count, debug=False):
```
DocGarbanzo, line 26: Can you provide a default (probably 0 here), so we retain the previous behaviour if the variable is not set?

donkeycar/templates/cfg_complete.py
```diff
@@ -357,3 +357,4 @@
 STOP_SIGN_DETECTOR = False
 STOP_SIGN_MIN_SCORE = 0.2
 STOP_SIGN_SHOW_BOUNDING_BOX = True
+STOP_SIGN_MAX_REVERSE_COUNT = 10    # How many times should the car reverse when detected a stop sign, set to 0 to disable reversing
```
DocGarbanzo, line 360: EOF blank line missing (see comments in other PR).

DocGarbanzo (changes requested):

donkeycar/parts/throttle_filter.py
JoeSiu: As it seems the pilot throttle is `NoneType` at start until I switch the drive mode to local pilot or local angle, then it will become `numpy.float32` afterward.

Here is the result if I remove the checking:
terminal output

DocGarbanzo:
Hi @JoeSiu - sorry, took me some time to look at this, this looks good to me. Can you please rebase & squash all your commits into one and then force-push to your branch. This means all your change will be in a single commit. I will then merge this.

JoeSiu:
@DocGarbanzo I have rebased the commits, thanks!

DocGarbanzo (approved):
Thanks.

PR #900, opened 2021-07-13
Fix bugs and Improve the approach of makemovie

BillyCheung10botics:
## 1.  fix the bugs originated from the latest commit b544241fec146edcbb708edf58c813cef0b727b2
The commit b544241fec146edcbb708edf58c813cef0b727b2 made a modification to the keras.py part on which the makemovie.py heavily depends. This results in errors of the code calling the keras part, https://github.com/autorope/donkeycar/blob/50f0831f824302693d7c2f479b95365fd7e41c05/donkeycar/management/makemovie.py#L76 https://github.com/autorope/donkeycar/blob/50f0831f824302693d7c2f479b95365fd7e41c05/donkeycar/management/makemovie.py#L147-L148 https://github.com/autorope/donkeycar/blob/50f0831f824302693d7c2f479b95365fd7e41c05/donkeycar/management/makemovie.py#L207
## 2.  improve the approach to calculate saliency maps,
2.1 evaluate the effects of all output layers instead of one
2.2 correct the problematic math (blending the gradients of angle and throttle output layers together) from direct sum to square norm
2.3 calculate the saliency depending on the model type "linear" and "categorical"

## 3.  modify the "draw_user_input", "draw_model_prediction", and "draw_steering_distribution" methods
I add an argument taking an image array to each method, such that the labels are drawn on this image array instead of the same image for inference. For instance,  if we draw the user input green line on the input image, and then pass the same image for inference, that green line could affect the model prediction and give an incorrect prediction blue line. 
Screenshot from 2021-07-13 18-28-53
These labels (yellow-purple cloud, blue line, green line, red line and grey lines) could affect the model prediction. It is better to seperate the image array for inference and the image array on which labels are drawn.

### DIFF
--- a/donkeycar/management/makemovie.py
+++ b/donkeycar/management/makemovie.py
@@ -52,6 +52,7 @@ class MakeMovie(object):
                 parser.print_help()
                 return
 
+        self.model_type = args.type
         self.tub = Tub(args.tub)
 
         start = args.start
@@ -73,7 +74,7 @@ class MakeMovie(object):
             self.keras_part = get_model_by_type(args.type, cfg=self.cfg)
             self.keras_part.load(args.model)
             if args.salient:
-                self.do_salient = self.init_salient(self.keras_part.model)
+                self.do_salient = self.init_salient(self.keras_part.interpreter.model)
 
         print('making movie', args.out, 'from', num_frames, 'images')
         clip = mpy.VideoClip(self.make_frame, duration=((num_frames - 1) / self.cfg.DRIVE_LOOP_HZ))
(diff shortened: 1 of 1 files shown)


sctse999:
@DocGarbanzo  @naisy any further comment? If no I will  merge this.

donkeycar/management/makemovie.py
```diff
@@ -204,7 +215,7 @@ def draw_salient(self, img):
 
         alpha = 0.004
         beta = 1.0 - alpha
-        expected = self.keras_part.model.inputs[0].shape[1:]
+        expected = self.keras_part.interpreter.model.inputs[0].shape[1:]
```
DocGarbanzo, line 218: There is a member function `get_input_shapes()` in `KerasPilot`.

DocGarbanzo (approved):
Looks good to me. Just a minor comment.

Issue #912, opened 2021-08-04
Cannot start 2D lidar if using a RealSense D435 camera


Ezward:
If you try to configure `CAMERA_TYPE="D435"` and `LIDAR_TYPE="RP"` simultaneously, the D435 part will NOT be correctly started when manage.py is run, but the RPLidar will not.  RPLidar works fine if camera is `MOCK`.

The problem is a bug in `manage.py` (created from `donkeycar/templates/complete.py`).  The code that configures that camera for some reason includes the lidar config.  So on line 90 of https://github.com/autorope/donkeycar/blob/dev/donkeycar/templates/complete.py we start configuration of the camera with the 'stereo' camera, then on line 113 we 'elif' camera is "D435", then on line 136 we 'else' the other cameras.  That 'else' block also contains the lidar configuration.  So if we configure the D435 then the else is never executed and so the lidar part is not configured.

Solution: move the code block starting on line 166, https://github.com/autorope/donkeycar/blob/5e2c5608fb6b1158300266613d10807369085e56/donkeycar/templates/complete.py#L166, to outside and after the else, so it gets executed no matter what kind of camera is being configured.

TCIII:
@Ezward,

This solution only works if you have not created mycar.

I believe that we can modify manage.py in the mycar directory to correct this issue?

TCIII:
@Ezward,

Okay, as a quick fix I moved the # add lidar code to after "V.add(cam, inputs=inputs, outputs=outputs, threaded=threaded)" and on the same indent as the "else" in manage.py and here is the result:

(env) thomas@thomas-desktop:~/mycar$ python manage.py drive
________             ______                   _________
___  __ \_______________  /___________  __    __  ____/_____ ________
__  / / /  __ \_  __ \_  //_/  _ \_  / / /    _  /    _  __ `/_  ___/
_  /_/ // /_/ /  / / /  ,<  /  __/  /_/ /     / /___  / /_/ /_  /
/_____/ \____//_/ /_//_/|_| \___/_\__, /      \____/  \__,_/ /_/
                                 /____/

using donkey v4.3.0 ...
loading config file: /home/thomas/mycar/config.py
loading personal config over-rides from myconfig.py
INFO:__main__:PID: 10323
cfg.CAMERA_TYPE D435
INFO:__main__:cfg.CAMERA_TYPE D435
Depth Scale is:  0.0010000000474974513
INFO:donkeycar.vehicle:Adding part RealSense435i.
[...]

PR #913, opened 2021-08-04
Fix issue 912: Cannot start 2D lidar if using a RealSense D435

Ezward:
https://github.com/autorope/donkeycar/issues/912

If you try to configure `CAMERA_TYPE="D435"` and `LIDAR_TYPE="RP"` simultaneously, the D435 part will NOT be correctly started when manage.py is run, but the RPLidar will not. RPLidar works fine if camera is MOCK.

The problem is a bug in manage.py (created from donkeycar/templates/complete.py). The code that configures that camera for some reason includes the lidar config. So on line 90 of https://github.com/autorope/donkeycar/blob/dev/donkeycar/templates/complete.py we start configuration of the camera with the 'stereo' camera, then on line 113 we 'elif' camera is "D435", then on line 136 we 'else' the other cameras. That 'else' block also contains the lidar configuration. So if we configure the D435 then the else is never executed and so the lidar part is not configured.

Solution: move the code block starting on line 166, ` # add lidar`, in  donkeycar/donkeycar/templates/complete.py to outside and after the else, so it gets executed no matter what kind of camera is being configured.

### DIFF
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -163,16 +163,7 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None,
         else:
             raise(Exception("Unkown camera type: %s" % cfg.CAMERA_TYPE))
 
-        # add lidar
-        if cfg.USE_LIDAR:
-            from donkeycar.parts.lidar import RPLidar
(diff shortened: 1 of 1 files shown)


DocGarbanzo (approved):
LGTM.

Issue #910, opened 2021-08-01
Integrate 2D Lidar data into Donkeycar


Ezward:
This is a discussion issue.  The topic is how to integrate 2D lidar data, as from an RPLidar or YDLidar, into the DonkeyCar framework.  It is likely that this will result in a number of following-on issues to actually implement the design that we decide upon.   I'll start the discussion by adding a strawman as the first comment in this ticket.

Ezward:
# Integrating 2D Lidar data into Donkeycar

## The 2D Lidar data

A spinning lidar captures data in 360 degree arc around it.  It provides each data point as a distance from the lidar and the angle at which that reading was taken.  It has some maximum working range.  That range can be thought of as the radius of a circle where the lidar is at the center of that circle.  Beyond that range we may actually get some readings, but we treat those a noise.

The lidar's 'zero' angle is relative to itself and so is all the data that it provides to us (it's super narcissistic).  But we care about where the points are relative to the vehicle that the lidar is mounted on.  If we treat directly forward as 'zero' angle for the vehicle, then we need to know the offset of the zero angle on the lidar relative to the forward direction of the vehicle.  We can then adjust the angle component of each data point provided by the lidar to turn it into an angle relative to the vehicle.  It may well be that the way the lidar is mounted that the zero angle of the lidar exactly coincides with forward on the vehicle, but we should include this offset (which in that case would be zero) anyway so we have a more general model that can be applied to other lidars that may not have their zero in the same place.

## Plotting the 2D Lidar data
So once we turn the lidar data into a distance and angle relative to the vehicle, then we can use trigonometry to calculate (x,y) coordinates relative to the vehicle; the vehicle is defined as being at (0,0).  


## A complication

Another thing to think about is the skew inherent in the data taken from a moving vehicle.  Remember that the each data point provided by the lidar is given as a distance from the lidar and the angle at which the lidar was pointing when it took the reading.  But if the lidar is on a moving vehicle, then the position from which each point is taken is changing.  The lidar has some rate at which it can make a full 360 degree sweep.  This is typically 5 to 10 hz.  If the vehicle is stopped, then all data points are from a circle with the same center.  However, as the vehicle moves, each point is taken from a different location.  If the vehicle is moving 1 meter per second, which is pretty slow for a race car, and the lidar is spinning at 5hz, then the lidar has moved 1/5 of a meter between taking the first point in a sweep and the last point.  There is additional skew that happens if the vehicle is turning.  

## A solution

[...]

PR #918, opened 2021-08-15
Next gen lidar parts

Ezward:
There is a desire to integrate 2D lidar data more completely into the donkeycar pipeline - see [issue 910](https://github.com/autorope/donkeycar/issues/910).  In particular the idea is to integrate the lidar data with the front-facing camera image.  In addition there are some issues with regarding accuracy and latency.  This pull request endeavours to tackle both problems by creating new lidar parts for acquiring data from the RPLidar and plotting data. 

There a  few problems with the current RPLidar and LidarPlot parts.
- The RPLidar part would wait for a full scan to be available before updating any data.  For a lidar spinning at 5hz, this would incur a 200ms latency between data updates.
- The RPLidar part forced angles to be on integer degrees.  This is probably ok for near-obstacle avoidance since the slight error won't matter much.  However, for longer distances this will cause significant deviation from actual readings if those readings include fractional angles (as a higher-end lidar could do). 
- The RPLidar spins in a clockwise direction and so readings from it have angles that increase in clockwise direction.  However, mathematically we want angles to increase in a counter-clockwise direction. 
- The LidarPlot part was written to be consistent with this clockwise increasing angles by using pixels coordinates where the y-axis increases going down.
- There was no way to easily visualize what these parts were doing without running a donkeycar.

These issues have been addressed by adding new versions of the parts (RPLidar2 and LidarPlot2) which presume mathematical angles that increase in a counter clockwise direction and are plotted on the Cartesian plane where y increases going up.  This can be adjusted with parameters.  In addition the new  RPLidar2 streams measurements and makes them available immediately without waiting for a full scan.  A __main__ has been added to visualizing the output of the lidar without having to run a full donkeycar.

The current RPLidar and LidarPlot parts remain for those that have code that depends upon them.

Added RPLidar2 part for streaming measurements  from an RPLidar.
- added a new part that streams measurements rather than scans in order to provide the freshest data possible with low latency.
- angles are no longer forced into a 360 element bucket; we simply return what the lidar provides (after filtering). The lidar seems to return about 258 measurements per scan, but the part will actually adapt to whatever the lidar is providing for a full scan, which can actually change from scan to scan.  This has the advantage of returning less data if less data is returned.  Further, if the lidar provides denser data, then this will support that.
- The part also will filter measurements by angle and distance on the fly for efficiency.
- The part no longer attempts to sort the measurements (I can't see any reason we need to sort, but any client of the part can do the sort if they want).
- The part will also adjust angles so that they increase in a counter-clockwise direction as is expected generally.  The default is still angles increasing in a clockwise direction,  since that is what the RPLidar does naturally.
- The part will also adjust the zero-heading.  This will adjust for different mounting of the lidar such that the physical front of the lidar is not longer pointing in the logical front direction.  By default this is zero and the most common mounting of the RPlidar on a vehicle would use the default.

Also added LidarPlot2, which fixes a few issues in the old plotter.
- It now assumes Cartesian coordinates where y increases going up. The old LidarPlot (which remains untouched) used the pixels coordinate system where y increased going down.  This combined with the fact that the RPLidar was producing angles that increased in a clockwise direction, allowed the plotting of RPLidar work correctly.  However, since we are now assumin that angles increase in a counter-clockwise direction, we need to plot in cartesian coordinates.
- This will also take an argument for rotating the plot. By default the zero heading is on the positive x axis, which is 'East'.  The rotating 90 degrees, the zero  heading can point 'North' for instance.  This is useful when aligning to an image take by a FPV camera.
- This draws a line along the zero heading.

Added a __main__ so the part can be run from the command line
- This will draw the scans to a window.
- Scan rate can be specified.
- total number of scans can be specified
- filtering for angle and distance can be specified.
- The plot can be rotate.
- In the future I hope to make this support other lidars.

### DIFF
--- a/donkeycar/parts/lidar.py
+++ b/donkeycar/parts/lidar.py
@@ -63,7 +63,7 @@ class RPLidar2(object):
                  max_distance = sys.float_info.max,
                  forward_angle = 0.0,
                  angle_direction=CLOCKWISE,
-                 batch_size=int(1846 / 20),
+                 batch_ms=50,  # how long to loop in run()
                  debug=False):
         
         self.lidar = None
@@ -134,7 +134,7 @@ class RPLidar2(object):
         self.full_scan_index = 0
         self.total_measurements = 0
         self.iter_measurements = self.lidar.iter_measurements()
-        self.measurement_batch_size = batch_size
+        self.measurement_batch_ms = batch_ms
 
         self.running = True
 
@@ -243,9 +243,19 @@ class RPLidar2(object):
         return []
     
     def run(self):
-        for i in range(self.measurement_batch_size):
+        if not self.running:
+            return []
+        #
+        # poll for 'batch' and return it
+        # poll for time provided in constructor
+        #
+        batch_time = time.time() + self.measurement_batch_ms / 1000.0
+        while True:
             self.poll()
-        return self.run_threaded()
+            time.sleep(0)  # yield time to other threads
+            if time.time() >= batch_time:
+                break
+        return self.measurements
 
     def shutdown(self):
         self.running = False
@@ -850,8 +860,7 @@ if __name__ == "__main__":
             min_distance=args.min_distance, max_distance=args.max_distance,
             forward_angle=args.forward_angle,
             angle_direction=args.angle_direction,
-            batch_size=int(1846/args.rate)) # based on empirical measurements
-                                            # for A1M8-r6 model
+            batch_ms=1000.0/args.rate)
         
         #
         # construct a lidar plotter
--- a/donkeycar/parts/lidar.py
+++ b/donkeycar/parts/lidar.py
@@ -37,9 +37,6 @@ def angle_in_bounds(angle, min_angle, max_angle):
     """
     Determine if an angle is between two other angles.
     """
-    angle = angle
-    min_angle = min_angle
-    max_angle = max_angle
     if min_angle <= max_angle:
         return min_angle <= angle <= max_angle
     else:
@@ -221,7 +218,7 @@ class RPLidar2(object):
                         self.full_scan_index += 1
                             
             except serial.serialutil.SerialException:
-                logger.info('SerialException from RPLidar.')
+                logger.error('SerialException from RPLidar.')
 
     def update(self):
         start_time = time.time()
@@ -698,8 +695,7 @@ class LidarPlot2(object):
         
         # background
         draw.rectangle(bounds, fill=self.background_color)
-        
-        
+
         # bounding perimeter and zero heading
         plot_polar_bounds(draw, bounds, self.border_color,
                           self.angle_direction, self.rotate_plot)
--- a/donkeycar/parts/lidar.py
+++ b/donkeycar/parts/lidar.py
@@ -5,6 +5,7 @@ Lidar
 # requies glob to be installed: "pip3 install glob2"
 # requires Adafruit RPLidar driver to be installed: pip install Adafruit_CircuitPython_RPLIDAR
 
+import sys
 import time
 import math
 import pickle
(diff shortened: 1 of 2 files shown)


donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+                 min_angle = 0.0, max_angle = 360.0,
+                 min_distance = sys.float_info.min, max_distance = sys.float_info.max,
+                 forward_angle = 0.0,
+                 angle_direction=CLOCKWISE,
+                 debug=False):
+        
+        self.lidar = None
+        self.port = None
+        self.on = False
+
```
DocGarbanzo, line 36: Too many blank lines (pep-8).

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+          
+        if forward_angle < 0 or forward_angle > 360:
+            help.append("forward_angle must be 0 <= forward_angle <= 360")
+            
+        if angle_direction != CLOCKWISE and angle_direction != COUNTER_CLOCKWISE:
+            help.append("angle-direction must be 1 (clockwise) or -1 (counter-clockwise)")
+
+        if len(help) > 0:
+            msg = "Could not start RPLidar; bad parameters passed to constructor"
+            print(msg)
```
DocGarbanzo, line 59: We moved from printing to logging. - This is a nice way to assemble a comprehensive help message, very useful!

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+        temp_list = glob.glob ('/dev/ttyUSB*')
+        result = []
+        for a_port in temp_list:
+            try:
+                s = serial.Serial(a_port)
+                s.close()
+                result.append(a_port)
+                port_found = True
+            except serial.SerialException:
+                pass
```
DocGarbanzo, line 88: As you are not using the context manager, don't you need the `close()` of the serial stream here? Probably change to context manager.

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+        if not port_found:
+            print("No RPLidar is connected.")
```
DocGarbanzo, line 90: Logger.

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+            raise Error("No RPLidar is connected.")
+        
+        # initialize
+        self.port = result[0]
+        self.lidar = RPLidar(None, self.port, timeout=3)
+        self.lidar.clear_input()
+        time.sleep(1)
+        
+        print(self.lidar.info)
+        print(self.lidar.health)
```
DocGarbanzo, line 100: Logger

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+                    if new_scan:
+                        full_scan_count += 1
+                        full_scan_index = 0
+                        measurement_count = measurement_index  # number of points in this full scan
+                        measurement_index = 0   # start filling in next scan
+                        
+                    # rplidar spins clockwise, but we want angles to increase counter-clockwise
+                    if self.spin_reverse:
+                        angle = (360.0 - (angle % 360.0)) % 360.0
+                        pass
```
DocGarbanzo, line 128: This line looks obsolete.

donkeycar/parts/lidar.py
```diff
@@ -13,6 +14,187 @@
...
+                    # check for start of new scan
+                    if new_scan:
+                        full_scan_count += 1
+                        full_scan_index = 0
+                        measurement_count = measurement_index  # number of points in this full scan
+                        measurement_index = 0   # start filling in next scan
+                        
+                    # rplidar spins clockwise, but we want angles to increase counter-clockwise
+                    if self.spin_reverse:
+                        angle = (360.0 - (angle % 360.0)) % 360.0
```
DocGarbanzo, line 127: Is this correct? 360.0 - (angle % 360.0) should be in [0.0, 360.0] by construction, why do you still need the % 360.0 afterwards

donkeycar/parts/lidar.py
```diff
@@ -218,6 +401,114 @@ def run(self, distances, angles):
...
+
+        if self.angle_direction != COUNTER_CLOCKWISE:
+            theta = (360.0 - (theta % 360.0)) % 360.0
+            
+        theta = np.radians(theta)
+        sx = int(cx + math.cos(theta) * dist)
+        sy = int(cy - math.sin(theta) * dist)
+        ex = int(cx + math.cos(theta) * (dist + 2 * self.rad))
+        ey = int(cy - math.sin(theta) * (dist + 2 * self.rad))
+        fill = 128
```
DocGarbanzo, line 476: This looks pretty much like a copy of line 435ff - can you factor out that common code in order to reduce duplication / redundancy?

DocGarbanzo (commented):
This looks good to me, only some minor comments.

donkeycar/parts/lidar.py
Ezward: I removed these.

donkeycar/parts/lidar.py
Ezward: removed pass

donkeycar/parts/lidar.py
Ezward: 360 - (angle % 360) would turn zero into 360; I want it to be zero.

donkeycar/parts/lidar.py
Ezward: I started doing this (by breaking out the draw methods into functions that the class methods would call) but I found that the resulting methods would be tightly coupled and fragile because the part treat we treat coordinates and angles and how they output data differently.  I don't think these two lidar parts will coexist forever; I think we should deprecate LidarPlot in favour of LidarPlot2, but I don't want to remove it right away and break anyones code.  I would like to give them time to move to the newer part before we remove the older part and I would like to get the new part field tested.  So there is not much upside to trying to share code when we will just delete one anyway and there is  downside because it makes the code coupled and complex.  However, the thought process did get me going and I refactored the drawing code so it is stand-alone.  That is most of the code anyway.  So if we wanted to we could make the older RPLidar part use it (again, not worth the effort if we are going to remove it in the future).   We may want to move the drawing code into its own file.

donkeycar/parts/lidar.py
Ezward: Done.

donkeycar/parts/lidar.py
Ezward: I thought the same thing when I looked at the original code (I just copied this from the older part).  The thing is, we only open the serial port, then immediately close it.  We just want to know if it is available.  If it won't open, we get an exception.  If it does open we immediately close it.  So the context manager doesn't make sense here; we are not doing any work with the serial port.  In terms of if the code should call serial close; if we get an exception, we never got a serial port to close, so we can't close it.

donkeycar/parts/lidar.py
Ezward: I did a pep-8 on this code and got 321 lines of output.  "name 'e' does not conform to snake-case"  PEP-8 was written by an abuser.  Create a language that depends on invisible characters, then write tools that say you should be able to see the invisible characters. I've gone back and put one space between methods in classes and two spaces between classes.

Ezward:
@DocGarbanzo thank you very very much for the code review.  It made me go back and look more deeply at the code.  If found a couple of bugs while doing so.

donkeycar/parts/lidar.py
Ezward: Thanks.  I switch to logging everywhere except the main, which is using console io.

donkeycar/parts/lidar.py
DocGarbanzo: 👍

donkeycar/parts/lidar.py
DocGarbanzo: 👍

donkeycar/parts/lidar.py
DocGarbanzo: 👍

donkeycar/parts/lidar.py
DocGarbanzo: 👍

donkeycar/parts/lidar.py
DocGarbanzo: Thanks for explaining, that makes a lot of sense. Maybe put a one-line comment into the code to explain that the old part is expected to be deprecated.

donkeycar/parts/lidar.py
```diff
@@ -13,6 +16,257 @@
...
+    return angle
+
+
+def angle_in_bounds(angle, min_angle, max_angle):
+    """
+    Determine if an angle is between two other angles.
+    """
+    angle = angle
+    min_angle = min_angle
+    max_angle = max_angle
```
DocGarbanzo, line 42: These 3 lines are obsolete.

donkeycar/parts/lidar.py
```diff
@@ -13,6 +16,257 @@
...
+                        if self.measurement_index >= len(self.measurements):
+                            self.measurements.append(measurement)
+                            self.measurement_count = self.measurement_index + 1
+                        else:
+                            self.measurements[self.measurement_index] = measurement  # noqa
+                        self.measurement_index += 1
+                        self.full_scan_index += 1
+                            
+            except serial.serialutil.SerialException:
+                logger.info('SerialException from RPLidar.')
```
DocGarbanzo, line 224: This probably should be a `logger.error(...)`

donkeycar/parts/lidar.py
```diff
@@ -13,6 +16,257 @@
...
+            self.poll()
+            time.sleep(0)  # yield time to other threads
+        total_time = time.time() - start_time
+        scan_rate = self.full_scan_count / total_time
+        measurement_rate = self.total_measurements / total_time
+        logger.info("RPLidar total scan time = {time} seconds".format(time=total_time))
+        logger.info("RPLidar total scan count = {count} scans".format(count=self.full_scan_count))
+        logger.info("RPLidar total measurement count = {count} measurements".format(count=self.total_measurements))
+        logger.info("RPLidar rate = {rate} scans per second".format(rate=scan_rate))
+        logger.info("RPLidar rate = {rate} measurements per second".format(rate=measurement_rate))
```
DocGarbanzo, line 238: We generally switched to f-strings for readability and brevity (they also perform better, but I think this is less of a concern).

donkeycar/parts/lidar.py
```diff
@@ -219,6 +472,252 @@ def shutdown(self):
...
+        measurements: list of polar coordinates as (distance, angle) tuples
+        '''
+            
+        self.frame = Image.new('RGB', self.resolution, (255, 255, 255))
+        bounds = (0, 0, self.frame.width, self.frame.height)
+        draw = ImageDraw.Draw(self.frame)
+        
+        # background
+        draw.rectangle(bounds, fill=self.background_color)
+
```
DocGarbanzo, line 701: additional blank line

setup.py
```diff
@@ -58,13 +58,15 @@ def package_files(directory, strip_leading):
...
               'adafruit-circuitpython-ssd1306',
+              'adafruit-circuitpython-rplidar',
               'RPi.GPIO',
               'pyserial',
           ],
           'nano': [
               'Adafruit_PCA9685',
               'adafruit-circuitpython-lis3dh',
               'adafruit-circuitpython-ssd1306',
+              'adafruit-circuitpython-rplidar',
```
DocGarbanzo, line 69: We would need corresponding changes also in the conda yaml files (not sure if these packages are available in conda channels or pypi only).

donkeycar/parts/lidar.py
Ezward: Thank you, I remove the redundant lines

donkeycar/parts/lidar.py
Ezward: And so it should be, and so it is.

donkeycar/parts/lidar.py
Ezward: You are right, but I'll leave it for now.

donkeycar/parts/lidar.py
Ezward: eliminated.  Thanks.

setup.py
Ezward: Do we need those for conda?  This is only installed on rpi and nano, although I guess we could install it on the host computers as well.   If we only install this on RPi and Nano do we need to update the conda files?

Ezward:
Rebased against main.

setup.py
```diff
@@ -24,7 +24,7 @@ def package_files(directory, strip_leading):
     long_description = fh.read()
 
 setup(name='donkeycar',
-      version="4.3.10",
+      version="4.3.11",
```
DocGarbanzo, line 27: I think this needs to be 4.3.14 now - don't know why GitHub doesn't show it as conflict.

DocGarbanzo (approved):
I merged from main and resolved the version conflict.

PR #919, opened 2021-08-15
Added an image buffer to fix the artificial latency in sim

Maximellerbach:
We previously used time.sleep to simulate the latency, but this method slows down the whole loop.
I added an image/info buffer to fix this issue.

what it does:
- when an image is received, store it into the buffer
- if an image has been long enough in the buffer use it as self.frame and remove it from the buffer.

If you have SIM_ARTIFICIAL_LATENCY=100, you should expect to have from 4 to 5 images in the buffer

tested on my machine with donkeycar in the sim

### DIFF
--- a/donkeycar/parts/dgym.py
+++ b/donkeycar/parts/dgym.py
@@ -23,7 +23,8 @@ class DonkeyGymEnv(object):
         conf["exe_path"] = sim_path
         conf["host"] = host
         conf["port"] = port
-        conf['guid'] = 0
+        conf["guid"] = 0
+        conf["frame_skip"] = 1
         self.env = gym.make(env_name, conf=conf)
         self.frame = self.env.reset()
         self.action = [0.0, 0.0, 0.0]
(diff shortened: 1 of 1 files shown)


donkeycar/parts/dgym.py
```diff
@@ -34,24 +35,48 @@ def __init__(self, sim_path, host="127.0.0.1", port=9091, headless=0, env_name="
...
+        to_remove = []
+        for it in range(len(self.buffer)):
+            if now-self.buffer[it][0] >= self.delay:
+                to_remove.append(it)
+                self.frame = self.buffer[it][1]
+            else:
+                break
+
+        # clear the buffer
+        for _ in range(len(to_remove)):
```
DocGarbanzo, line 61: Why do you need a loop here, it looks like you are deleting the same thing again and again.

DocGarbanzo (changes requested):

donkeycar/parts/dgym.py
Maximellerbach: if there is multiple images "old enough" in the buffer, it will delete them from the buffer

Maximellerbach: as they are in a list, removing [0] will remove the oldest

DocGarbanzo: You are right. This will pop `len(to_remove)` times from the front. I'm still a bit confused about what the `to_remove` array is meant to be doing? Aren't you just counting in line 55 and then popping `self.buffer` that count times from the front? So couldn't `to_remove` be just an `int` here that we count up in line 55 and then in 62 we just say `del self.buffer[:to_remove]` ?

Maximellerbach: Yes you are right, I will refactor that

donkeycar/parts/dgym.py
```diff
@@ -34,24 +35,46 @@ def __init__(self, sim_path, host="127.0.0.1", port=9091, headless=0, env_name="
...
+                self.frame, _, _, self.info = self.env.step(self.action)
 
     def run_threaded(self, steering, throttle, brake=None):
         if steering is None or throttle is None:
             steering = 0.0
             throttle = 0.0
         if brake is None:
             brake = 0.0
-        if self.delay > 0.0:
-            time.sleep(self.delay / 1000.0)
```
DocGarbanzo, line 54: You are recalculating this number in every loop. Why don't you initialise `self.delay` already with the scaled number in the constructor in line 38?

donkeycar/parts/dgym.py
```diff
@@ -34,24 +35,46 @@ def __init__(self, sim_path, host="127.0.0.1", port=9091, headless=0, env_name="
...
+        self.buffer = []
+
+    def delay_buffer(self, frame, info):
+        now = time.time()
+        buffer_tuple = (now, frame, info)
+        self.buffer.append(buffer_tuple)
+
+        # go through the buffer
+        num_to_remove = 0
+        for it in range(len(self.buffer)):
```
DocGarbanzo, line 53: You can now make this more pythonic by not using the random access iterator at all but just: 
```python
for buf in in self.buffer:
    if now - buf[0] >= self.delay
    ...
```

DocGarbanzo:
Great, this looks good to me, I only have one minor comment. Can you please rebase and squash all the commits into one and force-push your branch so we keep git history linear? I will merge this then.

DocGarbanzo (approved):
LGTM.

PR #922, opened 2021-08-20
Fix for memory model inferencing

DocGarbanzo:
# Fix the `KerasMemory.run()` function 

_Currently_: the function expects the previous steering throttle values in the second argument
_Fix_: ignore the second argument and cache the previous return steering and throttle values from the network for the memory vector to be passed to the model.

Also fix the UI code accordingly. There are no changes to the car templates required.


donkeycar/templates/complete.py
```diff
@@ -378,8 +378,7 @@ def show_record_count_status():
         inputs = ['cam/image_array',
                   'imu/acl_x', 'imu/acl_y', 'imu/acl_z',
                   'imu/gyr_x', 'imu/gyr_y', 'imu/gyr_z']
-    elif cfg.USE_LIDAR:
```
DocGarbanzo, line 381: These lines are duplicated from 369/370

tikurahul (approved):

Issue #925, opened 2021-08-23
In the complete.py template mm_per_tick is hard coded with the value 0.306096 in the "HAVE_ODOM" cfg.


TCIII:
The robot wheel that I have an encoder attached to is 6.5 inches in diameter for a circumference of  20.42 inches. The encoder wheel has 20 slots so there are 1.02 inches/tick or 25.93 mm/tick. Therefore one rotation of the wheel should equal 518.69 mm and 10 rotations of the wheel should equal 5,185.94  mm or a little over 5 meters.

If I start manage.py drive and rotate the wheel 10 rotations while collecting records, I consistently get  between 0.06 - 0.07 meters of travel after shutting down "manage.py drive" when I should have a little over 5 meters.

I can set "MM_PER_TICK =" in myconfig.py and I am using 25.9347. In the RotaryEncoder class the default is "mm_per_tick=0.306096".

No matter what value I set for "MM_PER_TICK =" in either myconfig.py or the RotaryEncoder class, the distance travelled is always 0.06 m for 10 turns of the wheel when it should be a little over 5 meters.

If I do the following: (0.306096 mm/tick X 20 ticks/revolution) X 10 revolutions, I get 61.21 mm or 0.0612 meters of travel.
If I do the following: (0.306096 mm/tick X 20 ticks/revolution) X 5 revolutions, I get 30.61 mm or 0.0306 meters of travel.

The 0.0612 meters is approximately the distance I get when I run manage.py drive to collect records and turn the wheel 10 times. The 0.0306 meters is approximately the distance I get when I run manage.py drive to collect records and turn the wheel 5 times.

Therefore I believe that the code is using "mm_per_tick=0.306096" no matter what value is input in MM_PER_TICK =" in myconfig.py.

TCIII:
Ed Murphy reviewed the HAVE_ODOM cfg in the complete.py template and found that "mm_per_tick=0.306096" is hard coded in "enc = RotaryEncoder(mm_per_tick=0.306096, pin = cfg.ODOM_PIN, debug = cfg.ODOM_DEBUG)".

Recommended solution:

Change "mm_per_tick=0.306096" to "mm_per_tick=cfg.MM_PER_TICK"

Ezward:
[...]

PR #929, opened 2021-08-31
fix-encoder-pipeline

Ezward:
- the complete.py template hooked up the encoder parts
  using hard-coded value for mm_per_tick and ignored
  the configuration for MM_PER_TICK.
- The RotaryEncoder was also using a default
  value for the poll_delay so polling was incorrect
  in anything except the default vehicle loop rate.
- This commit fixes the building of the pipeline
  in complete.py to use the correct configuration.
  It also calculates the poll_delay for the RotaryEncoder
  based on the DRIVE_LOOP_HZ configuration.

### DIFF
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -78,11 +78,11 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None,
     if cfg.HAVE_ODOM:
         if cfg.ENCODER_TYPE == "GPIO":
             from donkeycar.parts.encoder import RotaryEncoder
-            enc = RotaryEncoder(mm_per_tick=0.306096, pin = cfg.ODOM_PIN, debug = cfg.ODOM_DEBUG)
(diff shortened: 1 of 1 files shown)


TCIII:
I integrated Ezward's updated complete.py into my DC v 4.3, created a new mycar, and verified that it ran without error. 
In myconfig.py I was able to set the "MM_PER_TICK = 25.9347" in the "ODOMETRY" and I was able to set "STEERING_PWM_INVERTED = True" in "STEERING FOR PIGPIO_PWM OUTPUT".
Running manage.py drive resulted in a "Distance Travelled" of ~5238 mm which is within less than a 1 percent deviation from the calculated value. Also I was able to observe correct differential steering of my 6WD robot chassis whether it was being controlled by my RC transmitter or the output of my SBC Rpi 4B.

DocGarbanzo (approved):
LGTM.

Ezward:
Fixes https://github.com/autorope/donkeycar/issues/925

PR #905, opened 2021-07-19
Support for sixad ps3controller on Nano

fengye:
It's hard to pair PS3 controller to Jetson Nano however we still can use sixad, which continuously maintained by Retropie team, to pair and connect the controller. Details of instructions [here ](https://retropie.org.uk/docs/PS3-Controller/#for-older-versions-of-retropie) and [there(out of date)](http://booting-rpi.blogspot.com/2012/08/dualshock-3-and-raspberry-pi.html), though referred to Pi but works on Nano too.

After successfully connecting the PS3 controller to Nano, there's a need to support the sixad version of PS3 button/axis mapping. Users can swap the PS3Joystick object with PS3JoystickSixAd in PS3JoystickController.init_js(). Tested with OEM PS3 controller on Nano 4GB kit.

Side note: we might want to update the docs about PS3 controller compatibility on Nano too.

PR #931, opened 2021-09-04
Support for sixad ps3controller on Nano(2nd try)

fengye:
Same as PR #905 but rebased and squashed.

It's hard to pair PS3 controller to Jetson Nano however we still can use sixad, which continuously maintained by Retropie team, to pair and connect the controller. Details of instructions here and there(out of date), though referred to Pi but works on Nano too.

After successfully connecting the PS3 controller to Nano, there's a need to support the sixad version of PS3 button/axis mapping. Users can swap the PS3Joystick object with PS3JoystickSixAd in PS3JoystickController.init_js(). Tested with OEM PS3 controller on Nano 4GB kit.

Side note: we might want to update the docs about PS3 controller compatibility on Nano too.

### DIFF
--- a/donkeycar/parts/controller.py
+++ b/donkeycar/parts/controller.py
@@ -346,6 +346,44 @@ class JoystickCreator(Joystick):
 
         return button, button_state, axis, axis_val
 
+class PS3JoystickSixAd(Joystick):
+    '''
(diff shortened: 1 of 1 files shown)


DocGarbanzo (approved):
Looks good to me - thanks!

PR #933, opened 2021-09-09
Add ability to read config variables from env

altexdim:
Add ability to read config variables from env

Usage example:

`DONKEYCAR_CFG_MAX_LOOPS=2000 python manage.py drive`

this command will override config variable MAX_LOOPS and set it to 2000
this is the equivalent of having MAX_LOOPS=2000 in the myconfig.py

### DIFF
--- a/donkeycar/config.py
+++ b/donkeycar/config.py
@@ -62,19 +62,19 @@ def load_config(config_path=None, myconfig="myconfig.py"):
                 config_path = local_config
     
     logger = getLogger()
-    logger.info('loading config file: {}'.format(config_path))
+    logger.info('loading config file: {}' . format(config_path))
     cfg = Config()
     cfg.from_pyfile(config_path)
 
     # look for the optional myconfig.py in the same path.
     personal_cfg_path = config_path.replace("config.py", myconfig)
     if os.path.exists(personal_cfg_path):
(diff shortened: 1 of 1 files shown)


donkeycar/config.py
```diff
@@ -51,19 +61,20 @@ def load_config(config_path=None, myconfig="myconfig.py"):
             if os.path.exists(local_config):
                 config_path = local_config
     
-    print('loading config file: {}'.format(config_path))
+    logger = getLogger()
+    logger.info('loading config file: {}' . format(config_path))
```
DocGarbanzo, line 65: Great, we should be moving to logging. Can you please also use f-strings?

DocGarbanzo (changes requested):
This is a great addition. Can you please also add a small test? Afterwards, please rebase (might do nothing...) and squeeze your commits into one, then I'll merge this.

DocGarbanzo:
@altexdim - do you want to address the comments so we can merge this?

altexdim:
Yes, absolutely, thank you for the review, I'll fix it as I find some time

Ezward:
@DocGarbanzo  if we like this feature (I like it), we should change that to f-string ourselves or merge as is.

PR #934, opened 2021-09-10
Add AiCatapult - an improved version of AiLauncher which also controls steering angle

altexdim:
Add AiCatapult - an improved version of AiLauncher which also controls steering angle

Comparing to AiLaunch this particular class also controls the angle.

To be able to use it you need to change <mycar>/manage.py:

old one
```
from donkeycar.parts.launch import AiLauncher

aiLauncher = AiLauncher(cfg.AI_LAUNCH_DURATION, cfg.AI_LAUNCH_THROTTLE, cfg.AI_LAUNCH_KEEP_ENABLED)
V.add(aiLauncher,
      inputs=['user/mode', 'throttle'],
      outputs=['throttle'])
```
new one
```
from donkeycar.parts.launch import AiCatapult

aiLauncher = AiCatapult(cfg.AI_LAUNCH_DURATION, cfg.AI_LAUNCH_THROTTLE, cfg.AI_LAUNCH_KEEP_ENABLED)
V.add(aiLauncher,
      inputs=['user/mode', 'throttle', 'angle'],
      outputs=['throttle', 'angle'])
```

### DIFF
--- a/donkeycar/parts/launch.py
+++ b/donkeycar/parts/launch.py
@@ -1,4 +1,6 @@
 import time
+from datetime import datetime
+from logging import getLogger
 
 class AiLaunch():
     '''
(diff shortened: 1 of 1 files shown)


donkeycar/parts/launch.py
```diff
@@ -46,3 +48,73 @@ def run(self, mode, ai_throttle):
 
         return new_throttle
 
+
+class AiCatapult():
```
DocGarbanzo, line 52: This class is effectively just a copy of the above class with a very small change to return 0 angle in launch mode. We should probably drive this with a single bool switch in the `AiLaunch` class above. Also the `()` should go. Then we should add a config parameter in `cfg_complete.py` that can trigger the 0 angle behaviour and pass this in `complete.py`

DocGarbanzo (changes requested):

DocGarbanzo:
@altexdim - do you want to have a look at the above code review and make the corresponding changes?

altexdim:
Yes, absolutely, thank you for the review, I'll fix it as I find some time

PR #935, opened 2021-09-10
now donkey gets version from setup.py

Ezward:
- installation was failing because setup was trying to
  import from donkeycar before it was setup.
- setup.py imports __version__ from donkeycar;
  that causes __init__.py to be loaded,
  which tries to import pyfiglet.
  We can't import from donkeycar in setup.py because it is not yet setup
- this fix uses pkg_resources.get_distribution() to
  get the version set in setup.py and make it
  available to the donkeycar code.

From now on we set the version in setup.py; it will be imported into the donkeycode.  This avoids the circular setup issue.

### DIFF
--- a/donkeycar/__init__.py
+++ b/donkeycar/__init__.py
@@ -1,10 +1,12 @@
 import sys
 from pyfiglet import Figlet
 import logging
+from pkg_resources import get_distribution
+
+__version__ = get_distribution('donkeycar').version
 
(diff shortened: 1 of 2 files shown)


DocGarbanzo:
The pip installation isn’t working right? B/c our conda install on GH actions was working all the time.

DocGarbanzo (approved):
LGTM - thanks!

PR #936, opened 2021-09-19
Threading TubWriter for better performance

fengye:
The disk writing on RPi 4 is pretty slow compared to desktops, unless you mount a SSD on the car with it.

The idea is to unblock the TubWriter's IO from main thread. I see a performance boost from average 4ms to 0.2ms:
Before (IO on main thread):
```
+------------------------------+-------+------+------+------+------+-------+-------+
|             part             |  max  | min  | avg  | 50%  | 90%  |  99%  | 99.9% |
+------------------------------+-------+------+------+------+------+-------+-------+
|          TubWriter           | 14.07 | 2.46 | 4.41 | 3.47 | 7.24 | 11.31 | 13.32 |
+------------------------------+-------+------+------+------+------+-------+-------+
```

After (IO threaded):
```
+------------------------------+-------+------+------+------+------+------+-------+
|             part             |  max  | min  | avg  | 50%  | 90%  | 99%  | 99.9% |
+------------------------------+-------+------+------+------+------+------+-------+
|          TubWriter           | 79.88 | 0.07 | 0.19 | 0.08 | 0.09 | 0.13 | 21.36 |
+------------------------------+-------+------+------+------+------+------+-------+
```

### DIFF
--- a/donkeycar/parts/tub_v2.py
+++ b/donkeycar/parts/tub_v2.py
@@ -9,6 +9,11 @@ from PIL import Image
 
 from donkeycar.parts.datastore_v2 import Manifest, ManifestIterator
 
+from queue import Queue, Empty
+from threading import Lock
+import logging
+
+logger = logging.getLogger(__name__)
 
 class Tub(object):
     """
@@ -31,7 +36,7 @@ class Tub(object):
         if not os.path.exists(self.images_base_path):
             os.makedirs(self.images_base_path, exist_ok=True)
 
-    def write_record(self, record=None):
+    def write_record_ts(self, record, timestamp_sec):
         """
         Can handle various data types including images.
         """
@@ -64,12 +69,20 @@ class Tub(object):
                     image.save(image_path)
                     contents[key] = name
 
-        # Private properties
-        contents['_timestamp_ms'] = int(round(time.time() * 1000))
+        # Private properties    
+        contents['_timestamp_ms'] = round(timestamp_sec * 1000)
         contents['_index'] = self.manifest.current_index
         contents['_session_id'] = self.manifest.session_id
 
         self.manifest.write_record(contents)
+
+    def write_record(self, record):
+        """
+        write_record should be only used in single threaded mode, as the timestamp 
+        will be generated on the current thread.
+        """
+        self.write_record_ts(record, time.time())
+
 
     def delete_records(self, record_indexes):
         self.manifest.delete_records(record_indexes)
@@ -109,9 +122,21 @@ class TubWriter(object):
     """
     A Donkey part, which can write records to the datastore.
     """
+
+    class TubRecordCached:
+        """
+        Simple helper for storeing record in a queue
+        """
+        def __init__(self, record, timestamp):
+            self.record = record
+            self.timestamp = timestamp
+
     def __init__(self, base_path, inputs=[], types=[], metadata=[],
                  max_catalog_len=1000):
         self.tub = Tub(base_path, inputs, types, metadata, max_catalog_len)
+        self.record_queue = Queue()
+        self.mutex = Lock()
+        self.on = False
 
     def run(self, *args):
         assert len(self.tub.inputs) == len(args), \
@@ -120,13 +145,55 @@ class TubWriter(object):
         self.tub.write_record(record)
         return self.tub.manifest.current_index
 
-    def __iter__(self):
-        return self.tub.__iter__()
+    def run_threaded(self, *args):
+        assert len(self.tub.inputs) == len(args), \
+            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
+        record = dict(zip(self.tub.inputs, args))
+        # push the (record, timestamp) tuple into queue
+        self.record_queue.put( (record, time.time()) )
+        self.mutex.acquire()
+        count = self.tub.manifest.current_index
+        self.mutex.release()
+        return count
+
+    def dequeue_and_process_record(self):
+        # pop a (record, timestamp) tuple
+        record, timestamp = self.record_queue.get()
+        # write the record to disk
+        self.mutex.acquire()
+        self.tub.write_record_ts(record, timestamp)
+        self.mutex.release()
+        # notify this task is done
+        self.record_queue.task_done()
+
+    def update(self):
+        self.on = True
+        # This loop runs like a 'normal' threaded loop 
+        while self.on:
+            if not self.record_queue.empty():
+                try:
+                    self.dequeue_and_process_record()
+                except Empty:
+                    logger.warn('Nothing to pop from record_queue. This should not be happening. Check if there is any other thread accessing record_queue.')
+
+        # Main thread triggered shutdown(), finish up everything in the queue
+        while not self.record_queue.empty():
+            try:
+                self.dequeue_and_process_record()
+            except Empty:
+                logger.warn('Nothing to pop from record_queue. This should not be happening. Check if there is any other thread accessing record_queue.')
 
     def close(self):
+        self.record_queue.join()
+        self.mutex.acquire()
         self.tub.close()
+        self.mutex.release()
+        # Force everything written to disk
+        os.sync()
 
     def shutdown(self):
+        self.on = False
+        self.record_queue.join()
         self.close()
 
 
--- a/donkeycar/templates/basic.py
+++ b/donkeycar/templates/basic.py
@@ -193,7 +193,7 @@ def drive(cfg, model_path=None, model_type=None):
     # do we want to store new records into own dir or append to existing
     tub_path = TubHandler(path=cfg.DATA_PATH).create_tub_path() if \
         cfg.AUTO_CREATE_NEW_TUB else cfg.DATA_PATH
-    tub_writer = TubWriter(base_path=tub_path, inputs=inputs, types=types)
+    tub_writer = TubWriter(base_path=tub_path, inputs=inputs, types=types, threaded=True)
     car.add(tub_writer, inputs=inputs, outputs=["tub/num_records"],
             run_condition='recording')
     if not model_path and cfg.USE_RC:
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -693,7 +693,7 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None,
     tub_path = TubHandler(path=cfg.DATA_PATH).create_tub_path() if \
         cfg.AUTO_CREATE_NEW_TUB else cfg.DATA_PATH
     tub_writer = TubWriter(tub_path, inputs=inputs, types=types, metadata=meta)
-    V.add(tub_writer, inputs=inputs, outputs=["tub/num_records"], run_condition='recording')
+    V.add(tub_writer, inputs=inputs, outputs=["tub/num_records"], run_condition='recording', threaded=True)
 
     # Telemetry (we add the same metrics added to the TubHandler
     if cfg.HAVE_MQTT_TELEMETRY:
--- a/donkeycar/templates/simulator.py
+++ b/donkeycar/templates/simulator.py
@@ -399,7 +399,7 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None, camera_type
     tub_path = TubHandler(path=cfg.DATA_PATH).create_tub_path() if \
         cfg.AUTO_CREATE_NEW_TUB else cfg.DATA_PATH
     tub_writer = TubWriter(tub_path, inputs=inputs, types=types, metadata=meta)
-    V.add(tub_writer, inputs=inputs, outputs=["tub/num_records"], run_condition='recording')
+    V.add(tub_writer, inputs=inputs, outputs=["tub/num_records"], run_condition='recording', threaded=True)
 
     if cfg.PUB_CAMERA_IMAGES:
         from donkeycar.parts.network import TCPServeValue


zlite:
Wow! This is much needed. Hope we can integrate soon. Bravo!!

tikurahul (commented):
If you are using a queue then you don't really need a Mutex.

tikurahul (commented):
If you are using a queue then you don't really need a Mutex. 

You also might want to drain the queue before a shutdown, because you have a race condition now.

fengye:
The queue is thread-safe, true. The lock is introduced not because of the queue, it's because Tub itself isn't thread-safe. 

In shutdown(), the queue has to be depleted before it reaches it end.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +124,48 @@ def run(self, *args):
...
+                # pop a record
+                record = self.record_queue.get()
+                # write the record to disk
+                self.mutex.acquire()
+                self.tub.write_record(record)
+                self.mutex.release()
+                # notify this task is done
+                self.record_queue.task_done()
+            except Empty:
+                pass
```
DocGarbanzo, line 151: If we try/catch, then we should add a logger.error here, and not silently go over this. If the write fails, the tub might be corrupt.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +124,48 @@ def run(self, *args):
...
+        record = dict(zip(self.tub.inputs, args))
+        # push the record into queue
+        self.record_queue.put(record)
+        self.mutex.acquire()
+        count = self.tub.manifest.current_index
+        self.mutex.release()
+        return count
+
+    def update(self):
+        self.on = True
```
DocGarbanzo, line 139: Pls define the member already in the constructor.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +124,48 @@ def run(self, *args):
...
+        while self.on or not self.record_queue.empty() > 0:
+            try:
+                # pop a record
+                record = self.record_queue.get()
+                # write the record to disk
+                self.mutex.acquire()
```
DocGarbanzo, line 145: I don't think you need the lock here as this runs in a single continuous thread.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +124,48 @@ def run(self, *args):
         self.tub.write_record(record)
         return self.tub.manifest.current_index
 
+    def run_threaded(self, *args):
+        assert len(self.tub.inputs) == len(args), \
+            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
+        record = dict(zip(self.tub.inputs, args))
+        # push the record into queue
+        self.record_queue.put(record)
+        self.mutex.acquire()
```
DocGarbanzo, line 133: I believe you don't need a lock here as you are just reading the counter from the tub manifest.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +124,48 @@ def run(self, *args):
...
+                pass
+
+    """
+    Let's disable this first
     def __iter__(self):
         return self.tub.__iter__()
+    """
 
     def close(self):
+        self.mutex.acquire()
```
DocGarbanzo, line 160: The `queue.join()` should happen before we close the `tub` because this will close the files. I don't think you need a lock here, either.

DocGarbanzo (changes requested):

donkeycar/parts/tub_v2.py
```diff
@@ -112,6 +114,8 @@ class TubWriter(object):
     def __init__(self, base_path, inputs=[], types=[], metadata=[],
                  max_catalog_len=1000):
         self.tub = Tub(base_path, inputs, types, metadata, max_catalog_len)
+        self.record_queue = LifoQueue()
```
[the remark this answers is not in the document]
fengye: I thought Lifo or Fifo doesn't matter as it's just a container to buffer records. I didn't realise Tub.write_record() will not take time as arguments. So the whole thing needs to be reworked.

donkeycar/parts/tub_v2.py
fengye: I do believe you need a lock. Manifest class has nothing protect it from accessing from multiple threads. Here `Manifest.current_index` has a read access from main thread, and in `update`, which is called on part's worker thread, will in turn call `Manifest.write_record` which will change its `current_index`(write access). The read/write of a variable in Python isn't atomic. Unless I missed something.

donkeycar/parts/tub_v2.py
fengye: As I my previous comment said, `self.tub` is being accessed from two threads. On main thread, its `manifest.current_index` will be read, while in this thread, its `manifest.write_record()` will be called eventually. Those need to be taken care of. Of course it can be better here to implement a read/write lock. However due to the complexity in the whole datastore, I try to avoid to do so and a mutex is working fast enough for me.

donkeycar/parts/tub_v2.py
fengye: Calling `queue.join()` before tub is closed sounds good.  For the lock, here I added one because at the time `self.tub.close()` is called, there could be the worker thread still trying to access `self.tube.write_record`, and `Tub` class has no way to protect those two events happen one after another.

fengye (commented):
For the FIFO queue it seems I need more work to deal with the timestamp, as `Tub` will simply take `time.time()` when `write_record()` is called

donkeycar/parts/tub_v2.py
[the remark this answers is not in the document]
tikurahul: It should be a FIFO queue. The order of events is actually quite importnat.

donkeycar/parts/tub_v2.py
tikurahul: Who is using `manifest.current_index` from the main thread ? If its being used that way - we should break that contract and hide the `current_index` completely from the main thread. Otherwise we introduce a lot of other racy-ness.

donkeycar/parts/tub_v2.py
tikurahul: Sorry, i missed the call to `join()`. That should do the right thing here.

donkeycar/parts/tub_v2.py
fengye: `run_thread()` is using `manifest.current_index` from the main thread, as it needs to return the number of elements in the manifest. This behaviour aligns with `run()` but since now we're multithreading this writer, it needs to be protected. It doesn't need to be hidden, returning a local copy of the number should be fine.

fengye:
New code being squashed and force pushed. I'm still keeping the locks since I'm not convinced we're safe to leave them. I changed the other issues including the FIFO queue and timestamp issue. I did my local run the tub seems to be fine. Please have a review.

tikurahul (changes requested):
Just default `timespec_ms` to a timestamp and remove the `None` altogether.

donkeycar/parts/tub_v2.py
tikurahul: There is only 1 dedicated thread writing a record. So you don't need a lock.

donkeycar/parts/tub_v2.py
```diff
@@ -109,24 +117,79 @@ class TubWriter(object):
...
         return self.tub.manifest.current_index
 
-    def __iter__(self):
-        return self.tub.__iter__()
+    def run_threaded(self, *args):
+        assert len(self.tub.inputs) == len(args), \
+            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
+        record = dict(zip(self.tub.inputs, args))
+        recordCached = self.TubRecordCached(record, time.time())
+        # push the cached record into queue
```
tikurahul, line 148: TubRecordCached is not an instance variable. Also why are we introducing a new class for a cached TubRecord ?

donkeycar/parts/tub_v2.py
```diff
@@ -109,24 +117,79 @@ class TubWriter(object):
     """
     A Donkey part, which can write records to the datastore.
     """
+
+    class TubRecordCached:
```
tikurahul, line 121: We don't need this. Make a utility function if you absolutely need it.

tikurahul (changes requested):

donkeycar/parts/tub_v2.py
fengye: It's not an instance variable. TubRecordCached is a nested class. We need it because we want to cache the record in the queue, along with the timestamp at that time. We can't use `time.time()` in the writer thread as the timestamp will not be accurate. Thus we need this little class to do the job.

donkeycar/parts/tub_v2.py
fengye: Can you advice how to make a utility function for its job?

donkeycar/parts/tub_v2.py
fengye: > There is only 1 dedicated thread writing a record. So you don't need a lock.

There's only 1 deciated thread writing the records, true. But there's another thread reading some metadata of the records. In this case, it's `run_threaded` function that needs to return the number of the records already written. So here it's a one reader and one writer problem. It needs a lock. If we can avoid returning the number of records in `run_threaded`, then we can remove this lock here.

fengye:
> Just default `timespec_ms` to a timestamp and remove the `None` altogether.

I assume you're referring to the `write_record`'s arguments. The idea here is to keep `write_record` backward compatible and doesn't break the existing tests, because now if the `TubWriter` is running as threaded, its worker thread needs to feed `write_record` with a timestamp. I don't really know how this can be default and remove the `None`. A simple test like `def write_record(self, record, timestamp_sec=time.time())` doesn't work for me(`timestamp_sec` will be default to the time when its definition being parsed).

donkeycar/parts/tub_v2.py
tikurahul: We don't need to do that. The timestamp does not have to be precise. Also, it does not have to be nested either way.

tikurahul:
> > Just default `timespec_ms` to a timestamp and remove the `None` altogether.
> 
> I assume you're referring to the `write_record`'s arguments. The idea here is to keep `write_record` backward compatible and doesn't break the existing tests, because now if the `TubWriter` is running as threaded, its worker thread needs to feed `write_record` with a timestamp. I don't really know how this can be default and remove the `None`. A simple test like `def write_record(self, record, timestamp_sec=time.time())` doesn't work for me(`timestamp_sec` will be default to the time when its definition being parsed).

That's arguably better than what you are doing now because it reflects when the `Vehicle` intended to add it to the data store and that is more deterministic than what it being done here.

donkeycar/parts/tub_v2.py
fengye: Well I'm not an expert what timestamp is to do here but if there's a way to record it as precisely as before for both threaded and non-threaded mode at no cost then why not? I made it nested because I don't want to pollute the global namespace so it's just a little implementation details inside `TubWriter`.

fengye:
> > > Just default `timespec_ms` to a timestamp and remove the `None` altogether.
> > 
> > 
> > I assume you're referring to the `write_record`'s arguments. The idea here is to keep `write_record` backward compatible and doesn't break the existing tests, because now if the `TubWriter` is running as threaded, its worker thread needs to feed `write_record` with a timestamp. I don't really know how this can be default and remove the `None`. A simple test like `def write_record(self, record, timestamp_sec=time.time())` doesn't work for me(`timestamp_sec` will be default to the time when its definition being parsed).
> 
> That's arguably better than what you are doing now because it reflects when the `Vehicle` intended to add it to the data store and that is more deterministic than what it being done here.

I'm confused. If you are saying that worker thread of `TubWriter`(mostly IO heavy) should just generate the timestamp for a event happened on main thread, that is, the camera shot with all the driving parameters, then we definitely has less control over the precision of it. The main thread, if I remember correctly, has a proper time control to keep the loop running at a relatively stable frequency, while the worker thread can run at arbitrary speed depending on what it's doing and the os scheduling.

tikurahul:
You are needlessly complicating the design of `TubWriter`. It's single responsibility should be to `write_record`s posted in in FIFO queue. The expensive work is I/O and not generating timestamps (relatively an order of magnitude cheaper) so I am not even sure I understand why you are implying that timestamps are better cached / handled on the I/O thread.

`TubWriter` is only adding timestamps to maintain some notion of ordering. (I know, given I wrote `TubWriter` and the new datastore implementation; so you can trust me 😄)

You can store a tuple of (record, timestamp) and avoid introducing the `TubRecordCached` class altogether. 

I think all you need is to do a couple of things:

```python
def write_record(self, record, timestamp_sec=time.time()):
  # The actual current implementation of write_record, except you are using the timestamp passed in.
```
Then your `run_threaded(...)` method becomes:

```python
 def run_threaded(self, *args):
        assert len(self.tub.inputs) == len(args), \
            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
        record = dict(zip(self.tub.inputs, args))
        self.record_queue.put((record, time.time()))
```

and:

```python
    def dequeue_and_process_record(self):
        # Notice no mutexes here. 
        # pop a cached record
        record, timestamp = self.record_queue.get()
        # write the record to disk
        self.tub.write_record(record, timestamp)
        # notify this task is done
        self.record_queue.task_done()
```

fengye:
> You are needlessly complicating the design of `TubWriter`. It's single responsibility should be to `write_record`s posted in in FIFO queue. The expensive work is I/O and not generating timestamps (relatively an order of magnitude cheaper) so I am not even sure I understand why you are implying that timestamps are better cached / handled on the I/O thread.
> 
> `TubWriter` is only adding timestamps to maintain some notion of ordering. (I know, given I wrote `TubWriter` and the new datastore implementation; so you can trust me 😄)
> 
> You can store a tuple of (record, timestamp) and avoid introducing the `TubRecordCached` class altogether.
> 
> I think all you need is to do a couple of things:
> 
> ```python
> def write_record(self, record, timestamp_sec=time.time()):
>   # The actual current implementation of write_record, except you are using the timestamp passed in.
> ```
> 
> Then your `run_threaded(...)` method becomes:
> 
> ```python
>  def run_threaded(self, *args):
>         assert len(self.tub.inputs) == len(args), \
>             f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
>         record = dict(zip(self.tub.inputs, args))
>         self.record_queue.put((record, time.time()))
> ```
> 
> and:
> 
> ```python
>     def dequeue_and_process_record(self):
>         # Notice no mutexes here. 
>         # pop a cached record
>         record, timestamp = self.record_queue.get()
>         # write the record to disk
>         self.tub.write_record(record, timestamp)
>         # notify this task is done
>         self.record_queue.task_done()
> ```
It's all way around. I'm not saying timestamp should be generated on IO thread, instead it should be generated on the main thread and cached into a place(this can be anything, including a nested TubRecordCache class and you proposed tuple). That's why it needs to be generated and saved on main thread, which should be running fast, and written on the IO thread, which can be potentially slow.

> ```python
> def write_record(self, record, timestamp_sec=time.time()):
>```

This won't work. `timestamp_sec` will be fixed to a constant value when it's being parsed, unless the caller explicitly pass the parameter to the function.

Tuple of course can work, a nested class can too. If you insist, I can change for sure.

DocGarbanzo:
> Ah right. Python eagerly evaluates that expression. In that case, you can do something like this instead.
> 
> ```python
> def write_record(self, record, timestamp_fn = lambda: time.time()):
>     timestamp = timestamp_fn()
>     # Do something with timestamp
> ```
> 
> So `run_threaded` becomes:
> 
> ```python
> def run_threaded(self, *args):
>         assert len(self.tub.inputs) == len(args), \
>             f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
>         record = dict(zip(self.tub.inputs, args))
>         ts = time.time()
>         self.record_queue.put((record, lambda: ts))
> ```
> 
> and:
> 
> ```
>  def dequeue_and_process_record(self):
>         # Notice no mutexes here. 
>         # pop a cached record
>         record, timestamp_fn = self.record_queue.get()
>         # write the record to disk
>         self.tub.write_record(record, timestamp_fn)
>         # notify this task is done
>         self.record_queue.task_done()
> ```
> 
>  > Tuple of course can work, a nested class can too. If you insist, I can change for sure.
> 
> Tuples are perfect for this sort of use-case because they are fixed size containers.
> 
> One other note:
> 
> > ..That's why it needs to be generated and saved on main thread, which should be running fast, and written on the IO thread, which can be potentially slow. Note run_threaded despite its name, it's actually running on main thread. update is the function that running on its own thread.
> 
> It's `main` thread right now, but we will change the semantics of this soon. We need to do this because, there are limits to the number of threads we can spin up. A better way to manage the run loop is to potentially use co-routines.
> 
> Threaded parts don't necessarily have the best named APIs like you pointed out.
> 
> You can also choose to just introduce another method _overload_ if you don't want the `lambda` here (to keep things simpler).
> 
> So something like:
> 
> ```python
> def write_record(self, record):
>     timestamp = time.time()
>     return self.write_record_ts(record, timestamp)
> 
> def write_record_ts(self, record, ts):
>     # Write the record 
> ```

@tikurahul - I think this is better, so we don't have to pass lambdas in the interface. You also said the timestamp doesn't have to be accurate, but let's try to keep it as accurate as possible (in whole ms is fine), because if we look at sequence models we might run on different car frequencies between recording (say 40Hz) and inferencing (say 20Hz). If we have accurate timings of the records, then temporal interpolation can be handled in the model - for example the model just wants to know what happened 50ms, ago, 100ms ago, 150ms ago, etc.  With the approach of taking the time stamp from the main loop and passing this down into the `tub` this is achieved, so I think we are good.
@fengye - can you make those changes? - this is a great improvement.

fengye:
@DocGarbanzo I was away from my car for a few days. I'll be adding this `write_record_ts` to TubWriter instead. And the nested class will be removed and be replaced with tuples.

donkeycar/parts/tub_v2.py
DocGarbanzo: @tikurahul - do you have any objections left here?

donkeycar/parts/tub_v2.py
```diff
@@ -64,13 +69,21 @@ def write_record(self, record=None):
                     image.save(image_path)
                     contents[key] = name
 
-        # Private properties
-        contents['_timestamp_ms'] = int(round(time.time() * 1000))
+        # Private properties
```
tikurahul, line 72: Get rid of extra spaces in the end.

donkeycar/parts/tub_v2.py
```diff
@@ -109,9 +122,21 @@ class TubWriter(object):
     """
     A Donkey part, which can write records to the datastore.
     """
+
+    class TubRecordCached:
```
tikurahul, line 126: I thought you were planning to get rid of this completely and use a tuple instead ?

donkeycar/parts/tub_v2.py
```diff
@@ -109,9 +122,21 @@ class TubWriter(object):
...
+        """
+        def __init__(self, record, timestamp):
+            self.record = record
+            self.timestamp = timestamp
+
     def __init__(self, base_path, inputs=[], types=[], metadata=[],
                  max_catalog_len=1000):
         self.tub = Tub(base_path, inputs, types, metadata, max_catalog_len)
+        self.record_queue = Queue()
+        self.mutex = Lock()
```
tikurahul, line 138: We still don't need a lock.

donkeycar/parts/tub_v2.py
```diff
@@ -120,13 +145,55 @@ def run(self, *args):
...
 
-    def __iter__(self):
-        return self.tub.__iter__()
+    def run_threaded(self, *args):
+        assert len(self.tub.inputs) == len(args), \
+            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
+        record = dict(zip(self.tub.inputs, args))
+        # push the (record, timestamp) tuple into queue
+        self.record_queue.put( (record, time.time()) )
+        self.mutex.acquire()
```
tikurahul, line 154: No need for a mutex here. If you are worried about `current_index` gets being out of sync - that is always going to be the case as writes are happening on a dedicated thread. This won't help.

tikurahul (commented):
In general get rid of locks. You can guarantee ordering by using a Queue. Remove references to types that are no longer being used.

donkeycar/parts/tub_v2.py
fengye: One reader thread, one writer thread, when in doubt, use locks. I'm not worried `current_index` is "out of sync" nor I'm worry about `Queue` is not thread-safe. I'm following basic multi thread programming rule, to make the data store operation execute in sequence. Keep in mind Python can override attribute accessor quite easily, so there's no guarantee reading `current_index` is atomic nor it won't have interference with "writer" method like `write_record` without examine the implementation of `Manifest` thoroughly. So, without the knowledge of how `Manifest` works, it's better putting a lock around where we read anything from it, and around where we write anything to it. 

In addition, if you remove the lock, there won't be huge performance gain, but your code is more vulnerable to future changes.

Another side note, I've seen quite a lot places in this project with no proper thread protection, which makes the running wonky and sometimes unreliable. Understandable there are so many parts and implementations are from different contributor however it's better start to emphasise thread safety now.

donkeycar/parts/tub_v2.py
fengye: See my comments below

donkeycar/parts/tub_v2.py
fengye: Yes, this should be removed entirely.

donkeycar/parts/tub_v2.py
fengye: :)

donkeycar/parts/tub_v2.py
tikurahul: > One reader thread, one writer thread, when in doubt, use locks.

Use locks when thinking about contention. Not necessarily true otherwise. 
`current_index` being read from the manifest while writing is not going to be something that is going to be in sync even with locks.

fengye:
How can you be sure accessing manifest.current_index is a thread-safe read operation, without looking into Manifest class’s implementation?


On 1/12/2021, at 09:55, Rahul Ravikumar ***@***.***> wrote:

﻿

@tikurahul commented on this pull request.

________________________________

In donkeycar/parts/tub_v2.py<https://github.com/autorope/donkeycar/pull/936#discussion_r759652525>:

> @@ -120,13 +145,55 @@ def run(self, *args):
         self.tub.write_record(record)
         return self.tub.manifest.current_index

-    def __iter__(self):
-        return self.tub.__iter__()
+    def run_threaded(self, *args):
+        assert len(self.tub.inputs) == len(args), \
+            f'Expected {len(self.tub.inputs)} inputs but received {len(args)}'
+        record = dict(zip(self.tub.inputs, args))
+        # push the (record, timestamp) tuple into queue
+        self.record_queue.put( (record, time.time()) )
+        self.mutex.acquire()


One reader thread, one writer thread, when in doubt, use locks.

Use locks when thinking about contention. Not necessarily true otherwise.
current_index being read from the manifest while writing is not going to be something that is going to be in sync even with locks.

tikurahul:
> How can you be sure accessing manifest.current_index is a thread-safe read operation, without looking into Manifest class’s implementation?

It's not thread-safe (unless its being read) irrespective even with locks in place. `current_index` is only relevant when reading when you have a `Iterator`. The only entity that cares about `current_index` otherwise, is the writer itself, and given that is the one that is ordering writes with a `Queue` anyway there is no point having an additional lock.

My point is any use of `current_index` during a write is kind of moot. The only guarantee at this point is that the datastore write order is consistent, and for that we don't need to lock anything.

fengye:
> > How can you be sure accessing manifest.current_index is a thread-safe read operation, without looking into Manifest class’s implementation?
> 
> It's not thread-safe irrespective even with locks in place given you need to externally synchornize to not race in-between `write_record`s and to not miss `record` updates. `current_index` is only relevant when reading when you have a `Iterator`. The only entity that cares about `current_index` otherwise, is the writer itself, and given that is the one that is ordering writes with a `Queue` anyway there is no point having an additional lock.
> 
> My point is any use of `current_index` during a write is kind of moot. The only guarantee at this point is that the datastore write order is consistent, and for that we don't need to lock anything.

I'm not worried about `current_index` getting out-of-sync, as I need to say it a second time. 
You just said accessing `Manifest.current_index` is not thread safe. Okay, let's put it this way then. 
```
# Version 1
class A:
  A(self):
      self.index=0
  def methodB(self):
      while True:
        self.index = self.index+1

a = A() # Not thread-safe
t = threading.Thread(target=a.methodB)
t.start()
sleep(1)
print(a.index)
```
Sounds pretty okay right? By looking into the simple implementation you know `index` will be increasing forever and no one cares if printed `index` is a little falling behind threaded `methodB`. But now you have:

```
# Version 2
class A:
  A(self):
    self.i = 0
  def methodB(self):
      while True:
        self.doSomeStuff()
        self.i = self.i + 1
  def __getattr__(self, name):
    if name == 'index':
        self.doSomeStuff()
        return self.i
    else:
        raise AttributeError()
  def doSomeStuff(self):
      # Some complicated operations


a = A() # Still not thread-safe
t = threading.Thread(target=a.methodB)
t.start()
sleep(1)
print(a.index)
```
Now can you safely issue this `print(a.index)`? Probably not. Because you cannot assume `a.index` is a pure read-only operation. In fact, in this case, accessing `a.index` calls `doSomeStuff()` and same as `methodB` which is on another thread. How can you safely do this?
```
# Version 3
# Class A remains the same as in Version 2
class A:
  A(self):
    self.i = 0
  def methodB(self):
      while True:
        self.doSomeStuff()
        self.i = self.i + 1
  def __getattr__(self, name):
    if name == 'index':
        self.doSomeStuff()
        return self.i
    else:
        raise AttributeError()
  def doSomeStuff(self):
      # Some complicated operations

a = A() # Still not thread-safe
lock = Lock()

def methodC(a, lock):
  lock.acquire()
  a.methodB()
  lock.release()

t = threading.Thread(target=methodC, args=(a,lock))
t.start()
sleep(1)
lock.acquire()
print(a.index)
lock.release()
```
Yes that lock sounds redundant at the first glance. But without knowledge of class A, or the documentation says class A isn't thread-safe, then it's better adding a lock, even for something _looks_ like a pure read-only.

The whole point is, since a lot of classes and structures in donkeycar isn't implemented in thread safety in mind, and extremely lacks of document, here I locked the accessing of the instance of `Manifest` class, the instance of `Tub` class, and reference to their attributes. In this way, I can safely implement my features without knowing the logics behind `Tub` or `Manifest`.

If you find the lock is actually not necessary later, please remove them as you wish. I'm not commit my code without the locks.

tikurahul:
There can be classes that are not implicitly thread-safe; but **require external synchronization**. This is done all the time for performance sensitive code because most of the time you don't want to incur the costs of those locks.

If you don't want to make any more changes that is okay. We can add a threaded writer, which has most of your changes without locks.

Issue #945, opened 2021-10-11
Refactor motor drivers


Ezward:
Our motor drivers are pretty tightly bound to the libraries used to control PWM and GPIO pins.  This makes it hard to re-use code when building a new driver.

I'm in the process of enabling donkeycar software on a Duckiebot robot homebrew clone [here](https://github.com/Ezward/diyduckie).  It uses an TB6612FNG motor controller which is pin compatible with an L298N.  However, due to the evolution of the Duckiebot platform, one of it's motors get's it's enable signals from GPIO and the other from a PCA9685.  The other motor get's it's enable signals from GPIO and PWM from PCA9685.  

This would all be actually pretty easy to handle if we had our motor drivers accept a more generic interface for setting GPIO and PWM signals.  We could implement versions for the RPi.GPIO library, PiGPIO library and the PCA9685.  We could even write a version that controls this via an arduino using I2c or serial. 

- Create abstract base classes for  LogicPin and PwmPin.  
- Create concrete implementations for RPi.GPIO (and so Jetson.GPIO),  PCA9685 and PiGPIO
- Refactor motor drivers to take instances of LogicPin and PwmPin rather than raw pin numbers.
- update templates to instantiate instances of LogicPin and PwmPin when constructing drivetrain parts.

I would also refactor the templates by extractng the drivetrain construction from complete.py and making this a function that can be called by any template.  Then we could give other templates the flexibility to choose drivetrains, even if we only add a default in their config.  

Add more documentation on how to wire the motor drivers.  Each of our choices for motor implies a certain kind of hardware and wiring.  It would be useful to add a little more documentation and references on this.  Links to product descriptions on Amazon and Aliexpress tend to go stale very quickly, so we should focus on finding datasheets and tutorials that are likely to stay live longer.

This article does a very detailed job of describing the possible connections to an L298N https://www.etechnophiles.com/l298n-motor-driver-pin-diagram/  However, it does not explicitly describe how only the duty cycle inputs are used and the enable pins are not used.  The links to the product pages are 404ing.  I'm assuming this driver is for a 'mini' L298N.  That board uses just 4 wired to hookup 2 motors as this driver assumes.  Here is an article on how t use this with an arduino; https://www.instructables.com/Tutorial-for-Dual-Channel-DC-Motor-Driver-Board-PW/   Basically, the input pins are also used as PWM/duty cycle pins.  I assume you can use a full L298N module in this mode by leaving the enable pins jumpered and sending PWM via the input pint.

[...]

PR #951, opened 2021-11-03
945 refactor motor drivers

Ezward:
Addresses [Issue #945](https://github.com/autorope/donkeycar/issues/945)
Companion [documentation PR](https://github.com/autorope/donkeydocs/pull/25)

This should be QA'd in conjunction with the [documentation PR](https://github.com/autorope/donkeydocs/pull/25) using the [test plan](https://github.com/autorope/donkeycar/issues/945#issuecomment-999019364) in [Issue 945](https://github.com/autorope/donkeycar/issues/945) 

NOTE: I expect and desire a lot of input on this pull request.  If you like it and want to approve it, then great.  But please don't merge it.  I'd like to get consensus on this prior to merge.

This is a large refactor of the motor drivers.  The goal is to make the motor drivers more reusable.  I got the itch to do this because I have a Duckiebot differential drive robot.  In another branch I'm adding odometry/kinematics and pose estimation, but I could not test it on this bot; Duckiebot uses an L298N 3pin kind of motor setup and we have a drive configuration, DC_TWO_WHEEL_L298N, that almost works.  The issue is that this bot uses a PCA9685 to generate the duty cycle.  Actually it's worse than that, it mixes some TTL outputs from the GPIO and the PCA9685.  So our driver would not work, since it assumes only GPIO output (using RPi.GPIO library) for both TTL and PWM.  So I was in the position of writing yet another one-off driver.  I realized that if the differential drive code took a higher-level pin specification, then it would not be tied to any particular hardware or library.  Looking at actuator.py, there are a number of instances where we have almost the same driver logic, but re-implemented for different libraries or pwm hardware.  Worse, they diverge a little in how they actually work.  This then causes a proliferation of configuration options in myconfig.py.  There is very little documentation how the hardware maps to a particular drivetrain so how does a user know which drive train to choose for their hardware?   There are a number of classes for particular microcontrollers or pwm generators that are not actually hooked up or documented; they are just latent in the code; no one would know they are there or how to use them looking at the templates or myconfig.py.

So I did not want to add to this problem.   What I tried to accomplish was to refactor so we only need a few reusable drivetrains; one for servo/esc,  One for a server plus L298N,  one for differential drive L298N 3pin wiring and one for differential L298N 2pin wiring.  These drivers take pin-specifiers rather than raw pin numbers.  Pin specifiers indicate the source for the pin; PCA9685, RPi.GPIO/Jetson.GPIO or PIGPIO and other particulars.  These can all be used on raspberry pi or jetson nano, with our without a PCA9685.  This allows us to drastically cut down on the code in actuators.py and complete.py and the amount of configuration in config.py.

So this is pretty opinionated pull request.  Actuator.py has a lot of cruft in it.  I would like to see it paired down to what we really intent to support in the framework.  For those other edge cases, like a particular microcontroller implementation, I think those should be in their own repository with a readme that documents what they are and how the donkey code would need to be changes to use them.  Maybe that repo in is in autorope, maybe it should be in the author's own github account if it is particular to them.  If there is another way to handles these things, please let me know.

This pull request is the first of several I have lined up, so I hope to get a lot of input on this.  That will help guide me in how I manage the subsequent pull requests.  I would enjoy a lot of input.  I this this also needs a lot of testing, so I have when you are reviewing this that you will pull the code and try it on your donkeycar.
 
Here is the approach I took:

1. create a layer of abstraction over input/output/pwm pins.  I call this the pin provider api and it can be found in donkeycar/parts/pins.py.  There are 3 pin providers currently implemented; PCA9685, Rpi.GPIO (and so Jetsion.GPIO) and PIGPIO.  I'm trying to figure out of we need to support ServoBlaster and Arduino Firmata (or possibly our own Arduino sketch).  Pins are specified using a pin-spec string.  For instance, "PCA9685.1:40.13" specifies channel 13 on the PCA9685 at address 0x40 on I2C bus 1.  "RPI_GPIO.BOARD.33" specified board pin 33 driven by the RPi.GPIO library.  "PIGPIO.BCM.20" specifies broadcom gpio-20 driven by the pigpio library.

2. Refactor pwm controllers (those support set_pulse() methods) to use this new system, so we can eliminate the other library specific implementations that are tied to specific configurations.  This  allows us to be more general in our support of these underlying hardware architectures without users having to write their own parts.  For instance, the `PIGPIO_PWM` configuration supports an hardware layer that takes pwm pulses for both steering and throttle as you would do for an RC car with a servo and esc.  However, it requires the use of the PIGPIO libary, which does not work on Jetson Nano.  There is now a drivetrain named PWM_STEERING_THROTTLE and a generalized pulse controller called PulseController that uses the pins.py pin provider api to send servo style PWM pulses to Servo/ESC style RC car drivetrains.  This means we can retire PIGPIO_PWM and it's code (and I2C_SERVO drive train configuration).  A car with a servo/esc drivetrain can now be driven by RPi.GPIO, PIGPIO or a PCA9685 using just the PWM_STEERING_THROTTLE drivetrain configuration.

3. Refactor non-pwm-controller configurations, like the `DC_TWO_WHEEL` and `DC_TWO_WHEEL_L298N`, to use pin providers.  This has those same benefits; now a differential drive robot can use the GPIO header or a PCA9685 for PWM.  In fact, it can mix the two; I have a Duckiebot that does just that; now it can run donkeycar.  Duckies and Donkeys living together in peace!

4. Delete unused code.  We have a bunch of code that is not hooked up to anything OR is supporting boards that we really don't support.  So we have a bunch of code that is not in a template and has no configuration.  The only way to use it is to edit your manage.py (or create a fork and edit complete.py).  So that makes actuators a dumping ground for individuals' pet projects that are not really part of the framework.  That makes actuators much harder to maintain.  For the most part, I have not actually deleted this, but I have marked it as @deprecated and added a comment that describes why I think it should be deprecated and how we might actually rescue this code and make it part of the framework.  So with the exception of the DC_TWO_WHEEL** drivetrains, the old drivetrain code remains in place, but is marked as deprecated.  This allows us to merge this code into the 4.x branch without breaking changes (except to the DC_TWO_WHEEL drivetrains, as mentioned.  If we think these should also not be touched, I can put the original code back in and create a parallel set of configurations for the new pin-provider versions).  Come version 5.x, I believe we should either delete the deprecated code or make an effort to make it part of the framework.

5. Add a lot more commenting and documentation while we are at it.  I've added a lot of comments to the code.  There will be a second pull request for donkeydocs after this pull request has gone through some review.  Again, given that the old configurations are generally maintained, the old docs are still valid.

Here is a high level explanation of the file changes:
- donkeycar/parts/pins.py : this is a new file that contains that Pin Provider api and it's implementations.
- donkeycar/utilities/deprecated.py : adds a decorator for classes and functions that are deprecated.  This will log a warning when these are used, but will allow them to be used.
- donkeycar/parts/actuator.py: 
  - Added PulseController() as a general pwm controller that uses pin provider for pwm source.  This can replace the prior pulse controllers; PCA9685, PiGPIO_PWM, and RPi_GPIO_Servo and eliminate the slight differences those various pulse controllers have.  
  - Replaced L298N_HBridge_DC_Motor which was dedicated to RPi.GPIO with L298N_HBridge_3pin which does the same thing but uses pin provider so that it can be used on Jetson Nano (or with a PCA9685).  I can test this configuration.
  - Replaced Mini_HBridge_DC_Motor_PWM which was dedicated to RPi.GPIO with L298N_HBridge_2pin which does the same thing but uses pin provider so that it can be used on Jetson Nano (or with a PCA9685).  This has been tested with a logic analyzer, but it would be good to get this tested with a real robot that uses an L298N Mini HBridge controller.
  - Added methods to calculate 12bit pulse values from pulse length in microseconds and visa versa
  - Marked a lot of stuff as deprecated and commented those things.  Some of those things we may want to keep and integrate with the framework better.  This is where I need a lot of input from those individuals that added this code; is this worth keeping in this repository.  If so, should it be better integrated so that it is part of a template and has configuration defaults and documentation?  If not, do we move this out into it's own file and add comments on how to use it.  Do we just remove it?
- donkeycar/templates/complete.py/cfg_complete.py
  - added the PWM_STEERING_THROTTLE drivetrain configuration which uses the PulseController.  This is a drivetrain for RC cars with Servo/ESC.  It uses pin provider to specify the pwm pins.  The idea is that in version 5.x, we would remove the other Servo/ESC configurations and just have this one.  I can test this, but since this the Servo/ESC car is the majority of what is out there, it would be good to get others to test this.
  - modified the DC_TWO_WHEEL and DC_TWO_WHEEL_L298N drivetrains to use the pin provider versions of the drivers.  This is a place I need input; this replaces the old drivers and configuration such that users of the prior configuration will be forced to update their pin definitions.  Is this OK or should I keep the old code and deprecate it like I have done elsewhere.  I think this is not used very much, so the impact would be small, but I don't have data to support that hunch.
  - made DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE"  the default drivetrain configuration so that newly created cars will use the new pin provider drivers.  This would not affect older configurations, just newly created configurations.  It would be a way of getting a lot more real world testing on this prior to making this permanent in 5.x
- donkeycar/templates/calibrate.py and donkeycar/parts/web_controller/web.py and donkeycar/parts/web_controller/templates/calibrate.html: updated web calibration to support DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE".  I could use help testing this.
- donkeycar/management/base.py : added support to the `donkey calibrate` command for pin provider.  I've tested this with a PCA9685 servo/esc setup.  I could use help testing other setups.
- I added or updated a bunch of unit tests.

Verification before merging:
- [ ] [Test Plan](https://github.com/autorope/donkeycar/issues/945#issuecomment-999019364) passes for all configurations
- [ ] Code branch PR (this PR) is approved
- [ ] Corresponding Documentation branch PR is approved.


TCIII:
@Ezward,

Does not run on a NVIDIA Nano/Xavier as manage.py attempts to import RPi.GPIO as GPIO on a Nano/Xavier.

Ezward:
> @Ezward,
> 
> Does not run on a NVIDIA Nano/Xavier as pins.py attempts to import RPi.GPIO as GPIO on a Nano/Xavier. Pin.py needs to detect the SBC in use so that either Rpi.GPIO can be loaded for the Rpi or Jetson.GPIO (Jetson.GPIO comes preinstalled on Jetpack) can be loaded for the Nano/Xavier.

That is a bug in dev branch as well.  I've fixed it and pushed change to this branch.

donkeycar/management/base.py
```diff
@@ -196,20 +197,40 @@ def parse_args(self, args):
...
+            from donkeycar.parts import pins
+
+            pwm_pin = None
+            try:
+                pwm_pin = pins.pwm_pin_by_id(args.pwm_pin)
+            except ValueError as e:
+                print(e)
+                print("See pins.py for a description of pin specification strings.")
+                exit(-1)
+            print('init pin {}'.format(args.pwm_pin))
```
DocGarbanzo, line 220: We moved to f-strings instead of the `format` based strings.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+class InputPin(ABC):
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    @abstractmethod
+    def start(self, on_input=None, edge=PinEdge.RISING) -> None:
+        """
+        Start the pin in input mode.
+        on_input: function to call when an edge is detected, or None to ignore
```
DocGarbanzo, line 76: Should we use shpinx or numpy  syntax `:param on_input:` and `:return: ...` in order to get a nice render in an IDE?

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+    if pin_provider == PinProvider.PCA9685:
+        raise RuntimeError("PinProvider.PCA9685 does not implement InputPin")
+    if pin_provider == PinProvider.PIGPIO:
+        if pin_scheme != PinScheme.BCM:
+            raise ValueError("Pin scheme must be PinScheme.BCM for PIGPIO")
+        return InputPinPigpio(pin_number, pull)
+    raise RuntimeError("UnknownPinProvider ({})".format(pin_provider))
+
+
+def output_pin(pin_provider:str, pin_number:int, pin_scheme:str=PinScheme.BOARD, i2c_bus:int=0, i2c_address:int=40, frequency_hz:int=60) -> OutputPin:
```
DocGarbanzo, line 307: Rather lengthy line... we should probably try to stick to pep-8.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+# ----- RPi.GPIO/Jetson.GPIO implementations -----
+#
+try:
+    import RPi.GPIO as GPIO
+    # lookups to convert abstact api to GPIO values
+    gpio_pin_edge = [None, GPIO.RISING, GPIO.FALLING, GPIO.BOTH]
+    gpio_pin_pull = [None, GPIO.PUD_OFF, GPIO.PUD_DOWN, GPIO.PUD_UP]
+    gpio_pin_scheme = {PinScheme.BOARD: GPIO.BOARD, PinScheme.BCM: GPIO.BCM}
+except ImportError:
+    print("pigpio was not imported.")
```
DocGarbanzo, line 347: We switched to logging for these kind of shell outputs.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+
+def gpio_fn(pin_scheme, fn):
+    """
+    Convenience method to call GPIO function
+    using desired pin scheme.  This restores
+    the previous pin scheme, so it is safe to 
+    mix pin schemes.
+    """
+    prev_scheme = GPIO.getmode() or pin_scheme
+    GPIO.setmode(pin_scheme)
```
DocGarbanzo, line 359: Just an observation here - our PWM actuator parts are multithreaded for performance reasons so they don't clog up the vehicle loop. I hope there is no issue if these parts will talk to the actuators from different threads.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+    Output pin ttl HIGH/LOW using Rpi.GPIO/Jetson.GPIO
+    """
+    def __init__(self, pin_number:int, pin_scheme:str) -> None:
+        self.pin_number = pin_number
+        self.pin_scheme = gpio_pin_scheme[pin_scheme]
+        self._state = PinState.NOT_STARTED
+
+    def start(self, state:int=PinState.LOW) -> None:
+        if self.state() != PinState.NOT_STARTED:
+            raise RuntimeError("Attempt to start OutputPin that is already started.")
```
DocGarbanzo, line 420: Maybe add the PIN number to the exception message, so we know which ones is causing a problem.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,913 @@
...
+
+
+class OutputPinGpio(OutputPin):
+    """
+    Output pin ttl HIGH/LOW using Rpi.GPIO/Jetson.GPIO
+    """
+    def __init__(self, pin_number:int, pin_scheme:str) -> None:
+        self.pin_number = pin_number
+        self.pin_scheme = gpio_pin_scheme[pin_scheme]
+        self._state = PinState.NOT_STARTED
```
DocGarbanzo, line 416: I think all of these members are initiated equally in all of the derived classes, so maybe move this part of the construction into the base class?

donkeycar/templates/calibrate.py
```diff
@@ -54,6 +55,37 @@ def drive(cfg ):
...
+        throttle_controller = PulseController(
+            pwm_pin=pins.pwm_pin_by_id(cfg.PWM_THROTTLE_PIN), 
+            pwm_scale=cfg.PWM_THROTTLE_SCALE, 
+            pwm_inverted=cfg.PWM_THROTTLE_INVERTED)
+        throttle = PWMThrottle(controller=throttle_controller,
+                                            max_pulse=cfg.THROTTLE_FORWARD_PWM,
+                                            zero_pulse=cfg.THROTTLE_STOPPED_PWM, 
+                                            min_pulse=cfg.THROTTLE_REVERSE_PWM)
+
+        drive_train = dict()
```
DocGarbanzo, line 82: Is this variable used?

donkeycar/utilities/deprecated.py
```diff
@@ -0,0 +1,80 @@
...
+                category=DeprecationWarning,
+                stacklevel=2
+            )
+            warnings.simplefilter('default', DeprecationWarning)
+            return func2(*args, **kwargs)
+
+        return new_func2
+
+    else:
+        raise TypeError(repr(type(reason)))
```
DocGarbanzo, line 80: missing newline

DocGarbanzo (commented):
@Ezward - I think this is a great change, only minor comments / questions. In general I observed, that you don't use f-strings or logging and your line lengths are not generally pep-8 compliant. Maybe you can change that?

donkeycar/management/base.py
Ezward: I've changes .format() instances to f-strings

donkeycar/parts/pins.py
Ezward: I've changed print() to logger.debug()

donkeycar/parts/pins.py
Ezward: Done

donkeycar/parts/pins.py
Ezward: I'm doing my linting with 120 character line length.

donkeycar/management/base.py
DocGarbanzo: great

donkeycar/parts/pins.py
DocGarbanzo: not here, but I saw your changes in the other files.

Ezward: Fixed in pins.py now.  Thanks.

donkeycar/parts/pins.py
Ezward: Yes, I don't like this pattern where GPIO is a singleton with shared state.  I agree this could be an issue if there are points using different schemes.  One approach would be to require a global scheme to be set then just check against it and runtime fail in pin spec does not match.  Another approach is to use a semaphore/lock to make sure this section is not accessed by more than one thread.

donkeycar/parts/pins.py
Ezward: I generally do not like to use inheritance as a way to share code; I would prefer composition.  However, even that is not warranted here as these initializations are actually a little different except for a couple of cases.  Trying to share this simple code would create coupling that is probably worse than the cure.

donkeycar/templates/calibrate.py
Ezward: Yes, see line 117.  It assigns it; `    ctr.drive_train = drive_train`

donkeycar/utilities/deprecated.py
Ezward: Fixed, thanks.

donkeycar/parts/pins.py
Ezward: I fixed this in pins.py

Ezward: I'm still working on other files for this.

Ezward: Update actuators.py; that and pins.py are the files where I added the most code.

donkeycar/parts/pins.py
Ezward: I went and reviewed the RPi.GPIO source code and found that the method that sets the mode will only allow it to be set once!  See https://github.com/Tieske/rpi-gpio/blob/c36c6101712d5a0c370ef0023aae237b734bf51f/source/py_gpio.c#L514  So that means we should not actually be using mixed modes.  It's also been bugging me a little that using different libraries; PiGPIO and RPi.GPIO can get you in trouble.  So we may want to have a configuration that chooses the library and the pin scheme.  We would initialize the pin provider api with that configuraiton.  Pin specs will still include the scheme, but it much match the configuration or a runtime exception would be raised.  In this way we would prevent mixed libraries and mixed pin schemes.  Further, we could get rid of the code that tries to get/set/reset the pin mode, which causes a race condition.  What do you think @DocGarbanzo?

Ezward: The other thing I could do, which would protect if other code is also using RPi.GPIO, is to check the mode to see if it is already set to a different mode and raise and error if it is.  This is essentially what the library does, but we can do it in python and give a better error message.  This will also detect if other code is using RPi.GPIO in a different mode.  I'll probably do that.  I need to make sure I go back and make sure all the documentation says all pins must be in the same mode.

Ezward: So I changes the code so that it would detect any attempt to mix pin modes and raise an error.  So, unless there was no pin mode set yet, it will never try to change the pin mode.  This eliminates the race condition without a lock.

donkeycar/parts/actuator.py
```diff
@@ -81,25 +208,29 @@ def set_pulse(self, pulse):
...
+            raise ValueError("controller must have a set_pulse method")
 
         self.controller = controller
         self.left_pulse = left_pulse
         self.right_pulse = right_pulse
         self.pulse = dk.utils.map_range(0, self.LEFT_ANGLE, self.RIGHT_ANGLE,
                                         self.left_pulse, self.right_pulse)
         self.running = True
-        print('PWM Steering created')
+        logger.debug('PWM Steering created')
```
DocGarbanzo, line 233: Just an observation, I think we have statements like these constructor prints on the `info` level. Such initialisations are printed only once in the car app, so it doesn't clog up the shell anyway. Debug output would be for printing some data within the `run()` method or other detailed information that you really don't want to see, if you don't debug.

donkeycar/parts/pins.py
DocGarbanzo: Yeah composition > inheritance, if it's not worth to factor out common data (or even methods for that data), then that's fine.  If you say we cannot use mixed modes, do you you mean we cannot use PWM from PCA9685 for driving servo/esc and use PIGPIO, say for reading PWM simultaneously? Because this would be a drawback.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,1043 @@
...
+            self.set_high(channel)
+        elif duty_cycle == 0:
+            self.set_low(channel)
+        else:
+            # duty cycle is fraction of the 12 bits
+            pulse = int(4096 * duty_cycle)
+            try:
+                self.pwm.set_pwm(channel, 0, pulse)
+            except:
+                self.pwm.set_pwm(channel, 0, pulse)
```
DocGarbanzo, line 588: This looks spurious, if the `try` failed we should log an error, but not simply try again.

donkeycar/parts/pins.py
DocGarbanzo: ...? not in this version.. maybe still on your local change...

donkeycar/parts/pins.py
Ezward: I'm saying for the RPi.GPIO driver we can't mix BCM and BOARD pin numbers.  However, we can (and I have) mixed RPi.GPIO and PCA9685 pins.  You could also technically mix PIGPIO and PCA9685 pins.  Technically you could mix RPi.GPIO and PIGPIO, but that's not necessary and probably not advisable.

donkeycar/parts/pins.py
Ezward: I ported this.  Looks like a one-time retry.  I'll check with the original author why it is there, but I don't want to remove it without understanding that.  I can't really hurt leaving it, but it might hurt taking it out.

DocGarbanzo: I think it was Doug in d1efdfa4056cdbb0b6ae623397013dc9f625be74, and the code was in the PCA9685 before. I have taken this bit out of my fork a long time ago, as I saw regular issues where it was throwing from the `except` branch, because the I2C can sometimes produce a hick up. If a single pwm signal gets missed, we are just a missing a control for 1/vehicle_HZ seconds, so not too bad.

donkeycar/parts/pins.py
DocGarbanzo: Ok, cool, thanks for clarification.

donkeycar/parts/actuator.py
Ezward: Good point.  I switched the logging calls in the constructors to be info level.

donkeycar/parts/pins.py
Ezward: @DocGarbanzo Is there a reason not to do this retry?

donkeycar/templates/calibrate.py
DocGarbanzo: Got it, thanks.

donkeycar/parts/pins.py
DocGarbanzo: Yes, if there is an I2C error, which is something I have observed on my car, then the non-exception handled retry might throw which will stop the vehicle loop. The `except` code should simply log an error, and when the vehicle loop comes around again, it will `try` to set it again.

donkeycar/parts/actuator.py
```diff
@@ -32,6 +135,27 @@ def get_bus():
...
+        if duty_cycle == 1:
+            self.set_high()
+        elif duty_cycle == 0:
+            self.set_low()
+        else:
+            # duty cycle is fraction of the 12 bits
+            pulse = int(4096 * duty_cycle)
+            try:
+                self.pwm.set_pwm(self.channel, 0, pulse)
+            except:
```
DocGarbanzo, line 156: This issue is still here....

Ezward: I"m ok with the except: clause doing a retry.

Ezward:
Testing

for each hardware configuration:
- run `python manage.py` and open the web ui
- verify the drive wheels turn in the correct direction when moving forward and that the speed of the wheel is proportional to the amount of throttle.
- verify the drive wheels turn in the correct direction when reversing.
- verify the front wheels turn in the correct direction and proportional to the user's input

Hardware configurations
- [ ] Confirm that older configuration still works: RaspberryPi, DRIVE_TRAIN_TYPE = "I2C_SERVO" using PCA9685 to generate PWM pulses for a steering servo and ESC. 
- [ ] Confirm that the new configuration for the same hardware, but using pin specifiers, works: RaspberryPi, DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE" using PCA9685 to generate PWM pulses for a steering servo and ESC.  Verify the documentation branch describes the how to format pin specifiers for this configuration.

- [ ] Confirm that older configuration still works: Jetson Nano, DRIVE_TRAIN_TYPE = "I2C_SERVO" using PCA9685 to generate PWM pulses for a steering servo and ESC. 
- [ ] Confirm that the new configuration for the same hardware, but using pin specifiers, works: Jetson Nano, DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE" using PCA9685 to generate PWM pulses for a steering servo and ESC.  Verify that the documentation branch describes how to format pin specifiers for this configuration.
- [ ] Confirm that new configuration using control signals generated from GPIO rather that PCA9685 works.  This may require a little rewiring of the vehicle.  Verify that the documentation branch describes how to activate PWM output on the Jetson Nano GPIO bus and which two pins can generate PWM.   Verify that the documentation branch describes how to format pin specifiers for this configuration.

donkeycar/parts/pins.py
```diff
@@ -582,10 +582,7 @@ def set_duty_cycle(self, channel: int, duty_cycle: float):
         else:
             # duty cycle is fraction of the 12 bits
             pulse = int(4096 * duty_cycle)
-            try:
-                self.pwm.set_pwm(channel, 0, pulse)
-            except:
-                self.pwm.set_pwm(channel, 0, pulse)
+            self.pwm.set_pwm(channel, 0, pulse)
```
DocGarbanzo, line 585: I think this we should keep the try/catch and incorporate proper exception handling like: 
```python
try:
    self.pwm.set_pwm(channel, 0, pulse)
except Exception as e:
    logger.error(f'Error in talking to PCA9685: {str(e)}')
```

Ezward: Done.  Thanks.

Ezward:
I've rebased against master and updated the RC hat docs to reference the new drive train type.  I've completed most of the tests in https://github.com/autorope/donkeycar/issues/945#issuecomment-999019364 .  Those that are not completed are because of lack of hardware.  However, each of those have minimal risk because we have tested aspects of those configurations when we tested other configurations.  @DocGarbanzo I think this is good to go if you do.

DocGarbanzo:
I hope you have re-based against `dev` - can  you confirm? Can you also bump the version, that helps if folks come with questions and we see exactly which version of `dev` they are using.

donkeycar/parts/pins.py
```diff
@@ -0,0 +1,1043 @@
...
+            self.set_high(channel)
+        elif duty_cycle == 0:
+            self.set_low(channel)
+        else:
+            # duty cycle is fraction of the 12 bits
+            pulse = int(4096 * duty_cycle)
+            try:
+                self.pwm.set_pwm(channel, 0, pulse)
+            except Exception as e:
+                logger.error(f'Error on PCA9685 channel {channel}: {str(e)}')
```
DocGarbanzo, line 588: Awesome.

donkeycar/management/base.py
```diff
@@ -4,15 +4,14 @@
...
-from socket import *
 import logging
 
 from progress.bar import IncrementalBar
 import donkeycar as dk
 from donkeycar.management.joystick_creator import CreateJoystick
 from donkeycar.management.tub import TubManager
 from donkeycar.pipeline.types import TubDataset
-from donkeycar.utils import *
+from donkeycar.utils import normalize_image, load_image, math
```
DocGarbanzo, line 14: Yep, that's better, too.

DocGarbanzo (commented):
This all looks good to me!

Ezward:
Yes, I rebased against dev (NOT master, sorry.  I've fixed the comment).  I've bumped version to version="4.3.3". Thanks for all of your input on this complex branch.

DocGarbanzo (approved):
Awesome.

PR #958, opened 2021-11-29
Add back support for tubhist

DocGarbanzo:
The `donkey tubhist` command got lost when migrating to 4.x. We are bringing it back here.

### DIFF
--- a/donkeycar/management/base.py
+++ b/donkeycar/management/base.py
@@ -273,6 +273,60 @@ class MakeMovieShell(BaseCommand):
         mm.run(args, parser)
 
 
+class ShowHistogram(BaseCommand):
+
+    def parse_args(self, args):
+        parser = argparse.ArgumentParser(prog='tubhist',
+                                         usage='%(prog)s [options]')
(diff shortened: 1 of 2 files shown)


donkeycar/management/base.py
```diff
@@ -273,6 +274,59 @@ def run(self, args):
...
+        Produce a histogram of record type frequency in the given tub
+        '''
+        from matplotlib import pyplot as plt
+        from donkeycar.parts.tub_v2 import Tub
+
+        output = out or os.path.basename(tub_paths)
+        path_list = tub_paths.split(",")
+        records = [record for path in path_list for record
+                   in Tub(path, read_only=True)]
+        df = pd.DataFrame(records)
```
tikurahul, line 301: Pandas is a pretty heavy package to install on a SBC. If you can avoid it, that would be a lot nicer. We have been trying to avoid dependencies on it in the past.

DocGarbanzo: Good point, at least we can move the import into the function, one wouldn't run it on the RPi anyway as it directly prints into the screen.

tikurahul (approved):

PR #964, opened 2021-12-09
963 Install Jetson.GPIO on [nano]

Ezward:
- it was incorrectly installing the raspberry pi only library RPi.GPIO
- this caused issues where RPi.GPIO was imported.  This is because Jetson.GPIO, which is the Nvidia version of that library, creates an alias for RPi.GPIO so that code that uses RPi.GPIO will work on the nano but ONLY if it does NOT import RPi.GPIO
- Now we install Jetson.GPIO for the nano.


DocGarbanzo (approved):
LGTM.

PR #965, opened 2021-12-13
Install imgaug also on the robot and not only the pc, so the donkey c…

DocGarbanzo:
Install `imgaug` also on the robot and not only the pc, so the donkey car can use CROP without any additional manual installation. Also added a None type check in the augmentation part. I checked on RPi that this is working.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ with open("README.md", "r") as fh:
     long_description = fh.read()
 
 setup(name='donkeycar',
-      version="4.3.4",
+      version="4.3.5",
       long_description=long_description,
       description='Self driving library for python.',
       url='https://github.com/autorope/donkeycar',
--- a/setup.py
+++ b/setup.py
@@ -49,8 +49,7 @@ setup(name='donkeycar',
           'progress',
           'typing_extensions',
           'pyfiglet',
-          'psutil',
-          'imgaug'
+          'psutil'
       ],
       extras_require={
           'pi': [
@@ -60,6 +59,7 @@ setup(name='donkeycar',
               'adafruit-circuitpython-ssd1306',
               'RPi.GPIO',
               'pyserial',
+              'imgaug'
           ],
           'nano': [
               'Adafruit_PCA9685',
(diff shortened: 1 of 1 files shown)


donkeycar/pipeline/augmentations.py
```diff
@@ -117,5 +117,7 @@ def create(cls, aug_type: str, config: Config) -> iaa.meta.Augmenter:
 
     # Parts interface
     def run(self, img_arr):
+        if img_arr is None:
```
Ezward, line 120: This is a good change. This is orthogonal to the changes I made to the cameras

setup.py
```diff
@@ -50,6 +50,7 @@ def package_files(directory, strip_leading):
           'typing_extensions',
           'pyfiglet',
           'psutil',
+          'imgaug'
```
Ezward, line 53: the imgaug library break the CSIC and the opencv camera on the Jetson nano.  So we should probably add this to the pi for now, but leave it off the nano.

Ezward (changes requested):
The change to augmentations is good and is orthogonal to the changes I made to the cameras.  However installing imgaug on nano breaks the CSI camera, so we can add it to the pi, but not the nano.

donkeycar/pipeline/augmentations.py
DocGarbanzo: I think we don't need this any longer now, b/c you fixed the camera returning `None`.

setup.py
DocGarbanzo: I'll change this.

Ezward:
@DocGarbanzo This looks good.  If you can fix the conflicts then we can merge it.

donkeycar/pipeline/augmentations.py
DocGarbanzo: Removed through rebasing

DocGarbanzo:
Rebased on dev and bumped version

Ezward (approved):
Excellent.  Thanks.

Issue #32, opened 2017-02-03
Add Compound Eye line detection autopilot.


wroscoe:
I ported the line detection autopilot from the Compound Eye car into a python notebook. This should be added as an available autopilot. 

https://wroscoe.github.io/compound-eye-autopilot.html#compound-eye-autopilot

PR #966, opened 2021-12-14
cropping works

HansKristjan99:


Ezward:
@HansKristjan99 we need more before we can review this;
There is no description at all for this pull request.  We need some context before we can review this.  Remember that we also have to test this, so if you have tested this please tell us how and on what hardware and with what relevant configuration.
1. What: What is this pull request doing.  Something like , "Implements cropping for training and inference on the linear model" or 'Fixes bug described in Issue #32"
2. Why: explain the value of this feature or fix.
3. How: What changes did you make to implement the What.
4. Test: How should we test this and what other features might this affect?

Ezward:
@HansKristjan99 I've not looked deeply at this, but we do have a  cropping feature that can be applied to any model.  See TRANFORMATIONS and ROI_CROP_TOP in donkeycar/templates/cfg_complete.py.

DocGarbanzo (changes requested):
This code is not in a state where it can be reviewed.

Ezward:
I have not heard anything from the author so I am going to close this PR without merging.

Issue #970, opened 2021-12-17
Installing imgaug on Jetson Nano breaks CSIC camera


Ezward:
This has been found using Jetpack 4.6.  In order to use the 'CROP' transformation with an autopilot the imgaug dependency needs to be installed into the environment.  It installs without issue.  however, once installed the CSIC camera will no longer return frames.  This is not an issue in RaspberryPi.  I'm not sure if it is an issue in earlier versions of Jetpack.  

Using 'CROP' is a best practice when training an autopilot.  So it is a problem that one of our supported platforms cannot use it. Worse, the error that is produced provides no help in understanding the underlying issue.  We should try to find a fix or work-around for this issue.

- Determine if earlier versions of jetpack also suffer from this issue
- Determine if there is an specific version of imgaug or one of it's dependencies that can be installed to avoid this issue.
- If we can't find a way to make imgaug work then open a issue for implementing CROP and trapezoidal masks that don't use imgaug. There is an alternative library that used to use imgaug but has removed it because of this bug.  It may be a good replacement for us https://github.com/albumentations-team/albumentations#list-of-augmentations

It has also been determined that this breaks training on the nano, since the training code always tries to import imgaug.  If it cannot import then it fails.  If we do the pip install then it gets an runtime illegal instruction error.  So a fix for that, if we don't get the imgaug replaced, is to fix the training code to only import imgaug if it is used.  That may require some mock stub of the imgaug api we use; so if imgaug fails to import then use a mock that warns the user that augmentations are disabled.

Ezward:
See this related issue https://github.com/autorope/donkeycar/issues/749

Ezward:
when installing imgaug on the nano, I see a series of these errors while it is trying to install 
```
WARNING: Discarding https://files.pythonhosted.org/packages/44/ec/4eddbf9d17a917c51fb4ad159aa7137f506681e91ab559cf87d120e1d78d/Shapely-1.7.0.tar.gz#sha256=e21a9fe1a416463ff11ae037766fe410526c95700b9e545372475d2361cc951e (from https://pypi.org/simple/shapely/). Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
  Using cached Shapely-1.6.4.post2.tar.gz (225 kB)
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #978, opened 2022-01-10
970 imgaug workaround mock augmentations

Ezward:
Issue https://github.com/autorope/donkeycar/issues/970

Currently uses of Jetson Nano that use image augmentations or tranformations AND want to train on their Nano will get an exception because imgaug is not installed.  It is not installed because it conflicts with opencv, which is required for the CSIC camera.  The maintainer of imgaug has stated in related issues in the imgaug repo that he will not be fixing this.

This is a temporary hack to mock our ImageAugmentation class if imgaug cannot be imported.  This is a workaround so users of the Jetson Nano can train even if imgaug is not installed.  This allows them to train on their nano even if imgaug is not installed.

### DIFF
--- a/donkeycar/pipeline/augmentations.py
+++ b/donkeycar/pipeline/augmentations.py
@@ -1,121 +1,141 @@
 import cv2
 import numpy as np
 import logging
-import imgaug.augmenters as iaa
 from donkeycar.config import Config
 
 
 logger = logging.getLogger(__name__)
 
+# 
+# HACK: workaround for imgaug bug; mock our implementation
+# TODO: remove this when https://github.com/autorope/donkeycar/issues/970 is addressed.
+#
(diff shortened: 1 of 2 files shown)


Ezward:
For discussion: this is implemented to allow users to specify augmentations and transformations, but to ignore them.  We might want to throw an error instead and instruct them to remove the augmentations and/or transformations.

DocGarbanzo (approved):
Looks good to me.

DocGarbanzo (approved):
Perfect.

Ezward:
It failed Python package and test / set up python 3.7 (ubuntu-latest, install/envs/ubuntu.yml) (push) because of the conda bug; github eventually cancels the test.  This is preventing me from merging.

PR #979, opened 2022-01-15
Add argument parser to realsense435i.py main

Ezward:
Add argument parsing in the `__main__` self test program.
- prior this this PR it would require altering the code to test different features of the RealSense 435i driver.


DocGarbanzo (approved):
LGTM

Issue #980, opened 2022-01-15
CI is slow and Windows is not tested


DocGarbanzo:
We have two issues in the CI:
* The conda installation has become increasingly slow since beginning of Dec 2021, in particular on Ubuntu. On many occasions it took > 3h. 
*  The CI currently does not work for windows:
  - we have a slightly different TF install, where TF is not part of the conda yaml file
  - therefore, we can't easily add windows to the CI os matrix. Even when adding it temporarily we hit other issues

DocGarbanzo:
Fixed in #981

PR #981, opened 2022-01-17
Improve CI times by switching from conda to mamba

DocGarbanzo:
# Improve CI performance
The CI is  slow on OSX and very slow on Ubuntu as mentioned in #980. We are changing the packet manager from conda to mamba which performs the install much faster.
* Change installation from conda to mamba
* Improve the CI yaml by providing better named build steps and and a more granular break down between checkout, install, reporting and testing
* Remove push-docs actions

I spent some time trying to setup the CI also for windows, but many tests fail, because of the windows filesystem not coping with the Tub and Datastore tests, and the missing integration of trt in tensorflow. Also mamba is not supported, so currently it cannot be added easily and if the install is only patchy and many tests are not working then there is no point in maintaining integration test for windows. 

I will update the docs to describe the mamba install as this is helpful also outside of the CI after this gets merged.


DocGarbanzo:
Please note, the 'old' CI tests _set up python 3.7 (macos-latest|ubuntu-latest)_ are obsolete as they are being replaced by _Python package and test_.

Ezward (approved):
LGTM.  Thanks for doing this.

PR #983, opened 2022-01-19
clamp differential drive throttle and steering

Ezward:
- the throttle and steering passed to TwoWheelSteeringThrottle where just a 3/1000 over 1, but where triggering the check against a max of 1.
- this caused the differential drive cars to fail when trying to run the linear model.
- this commit simply clamps out of range values to the allowed range of -1 to 1 inclusive.

See @Rushil  issue in discord https://discordapp.com/channels/662098530411741184/671604367706554388/932156905818828861


DocGarbanzo (approved):
LGTM

Issue #972, opened 2021-12-18
Rename Auto modes in Web UI


zlite:
Rename "Local Pilot" to "Full Auto"
Rename "Local Angle" to "Auto Steering + Manual Throttle"

(I'm also not sure what the "(d)" means. Maybe remove?

![donkeyUI](https://user-images.githubusercontent.com/18071/146651740-e9d8086c-c3d1-4ab7-99a3-ca156d59ae0a.png)

Ezward:
Addressed in https://github.com/autorope/donkeycar/pull/985

PR #985, opened 2022-01-23
970 enable rc controller in webui

Ezward:
Addresses [Issue #973](https://github.com/autorope/donkeycar/issues/973) and [Issue #972](https://github.com/autorope/donkeycar/issues/972)

For issue #973:
- the web, RC and game controllers now the the user/mode and recording state as inputs and use them as defaults for their own internal state.
- the controllers also latch any changes they make to that state in their background thread, so it can be applied to outputs in the foreground thread (in run_threaded())
- those change synchronize changes between those controller in the server.  However, we also needed to make those changes get reflected in the browser ui.  So the local web controller was change to send any changes in state to the browser client via the websocket, so the web client and apply them and update the web ui.

For Issue #972:
- I updated the drive mode selections to `User`, `Full Auto` and `Auto Steering`.  I chose not to use the suggest of `Auto Steering and Manual Throttle` because it was too long for the ui.

In addition I added an explanation of the keyboard shortcuts for change the drive mode and recording in the web ui to the model help popup.

### DIFF
--- a/donkeycar/parts/web_controller/web.py
+++ b/donkeycar/parts/web_controller/web.py
@@ -246,7 +246,8 @@ class WebSocketCalibrateAPI(tornado.websocket.WebSocketHandler):
 
         if 'config' in data:
             config = data['config']
-            if self.application.drive_train_type == "I2C_SERVO":
+            if self.application.drive_train_type == "PWM_STEERING_THROTTLE" \
+                or self.application.drive_train_type == "I2C_SERVO":
                 if 'STEERING_LEFT_PWM' in config:
                     self.application.drive_train['steering'].left_pulse = config['STEERING_LEFT_PWM']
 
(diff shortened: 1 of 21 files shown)


donkeycar/parts/web_controller/web.py
```diff
@@ -143,34 +149,61 @@ def update(self):
...
-    def run_threaded(self, img_arr=None, num_records=0):
+    def update_wsclients(self, data):
+        if data:
+            for wsclient in self.wsclients:
+                try:
+                    data_str = json.dumps(data)
+                    logger.debug(f"Updating web client: {data_str}")
+                    wsclient.write_message(data_str)
+                except Exception as e:
+                    print(e)
```
DocGarbanzo, line 160: This should be a logger error.

DocGarbanzo (commented):
Looks good to me, only minor comment

DocGarbanzo (approved):
Looks good to me. Thank you.

PR #986, opened 2022-01-26
Adding FastAI(pytorch) as a first class citizen to the pipeline (squashed)

adricl:
This PR allows you to use Fast AI(pytorch) end to end in donkey car. I have tested training and driving with the simulator.

I have added a new type of model fastai_ so far linear is the only model implemented. More can be added in the future of course.
I have tried my best to emulate most of the function that Keras does but there are some idiosyncrasies between the two frameworks.

These normal training and driving commands work.
python train.py --tubs data/ --model models/fastai_linear_model.pth --type fastai_linear
python manage.py drive --model models/fastai_linear_model.pth --type fastai_linear

I do realise that this PR comes without a heads up or warning so if its rejected that's fine.
I do really like this project and the work that has been done, its just I prefer PyTorch to TensorFlow.

Any suggestions are welcome. I am new to python of this complexity so any pointers would be great.
I will also add some docs around this.

If this goes well I might also look at getting pytorch lightning into own part and working in the pipeline.

Please note this will not work with TensorRT or TF lite if there is demand I could look at this in the future.

Cheers

### DIFF
--- a/donkeycar/parts/fastai.py
+++ b/donkeycar/parts/fastai.py
@@ -1,31 +1,406 @@
-import os
-from fastai.vision import *
+"""
+
+fastai.py
+
+Methods to create, use, save and load pilots. Pilots contain the highlevel
+logic used to determine the angle and throttle of a vehicle. Pilots can
+include one or more models to help direct the vehicles motion.
+
+"""
+from abc import ABC, abstractmethod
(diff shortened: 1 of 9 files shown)


adricl:
This was a mistake

Issue #989, opened 2022-02-07
Steering does not work when using drive , but does in calibration .  Earlier commit works . 


c1505:
#### Issue description
Steering does not work when using drive , but does in calibration and it works when using an earlier commit.  Seeing that another person in discord reported the same problem and that [downgrading to 4.2.1 fixed it](https://discord.com/channels/662098530411741184/671604048218292233/935889345460699176) , I tried a more recent commit `d5b91301ff904873e70b9caba72823fe3454feb6`.  It fixed the problem.  I am using a Jetson nano 2gb and the other person has a raspberry pi so it seems to not be related to the particular board used.  I went back and reinstalled dev and created a new car to make sure that the problem replicated for me before opening this issue. 

#### Steps to reproduce the issue

1.  Follow software setup as described here for the [jetson nano ](https://docs.donkeycar.com/guide/robot_sbc/setup_jetson_nano/)
2. create donkeycar from template
3. perform calibration and modify myconfig.py
4. changes to myconfig.py for my car
```
PCA9685_I2C_BUSNUM = 1
CAMERA_TYPE = "CSIC"   # (PICAM|WEBCAM|CVCAM|CSIC|V4L|D435|MOCK|IMAGE_LIST)
IMAGE_W = 224
IMAGE_H = 224
PCA9685_I2C_BUSNUM = 1
STEERING_LEFT_PWM = 460         #pwm value for full left steering
STEERING_RIGHT_PWM = 260        #pwm value for full right steering
THROTTLE_FORWARD_PWM = 420      #pwm value for max forward throttle
THROTTLE_STOPPED_PWM = 370      #pwm value for no movement
THROTTLE_REVERSE_PWM = 320      #pwm value for max reverse throttle
``` 
5. initiate driving from the car folder `python manage.py drive`
6. navigate to browser
7. press j multiple times to try to turn wheels left
8. press k multiple times to try to turn wheels right
9. also tested on mobile with the touch interface

#### What's the expected result?

- wheels turn

#### What's the actual result?

- wheels do not turn

#### Additional details / screenshot

- the webrowser UI shows changes to the steering, but no changes happen in the car.  The throttle still works. 
- the working commit `d5b91301ff904873e70b9caba72823fe3454feb6` is from December 30th of 2021.  I have not tested other commits yet.  I figured that someone more familiar with the code would be able to narrow down where the problem originated better than me.  
[...]

PR #994, opened 2022-03-07
989 namespace drivetrain configurations

Ezward:
Addresses Issue https://github.com/autorope/donkeycar/issues/989

The PWM_STEERING_PIN configuration values are used in several DRIVETRAIN_TYPE configurations, each with a different default value. If you just set the DRIVETRAIN_TYPE = "PWM_STEERING_THROTTLE", but don't uncomment the PWM_STEERING_PIN value, then the default of a different drivetrain get's used.

So a workaround is to just uncomment PWM_STEERING_PIN in the PWM_STEERING_THROTTLE sections.

This fix changes the newer drivetrain configuration into python dictionaries so all configuration are namespaced.

### DIFF
--- a/donkeycar/templates/cfg_complete.py
+++ b/donkeycar/templates/cfg_complete.py
@@ -76,34 +76,44 @@ DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE"
 # Uses a PwmPin for steering (servo) and a second PwmPin for throttle (ESC)
 # Base PWM Frequence is presumed to be 60hz; use PWM_xxxx_SCALE to adjust pulse with for non-standard PWM frequencies
 #
-PWM_STEERING_PIN = "PCA9685.1:40.1"  # PWM output pin for steering servo
-PWM_STEERING_SCALE = 1.0   # used to compensate for PWM frequency differents from 60hz; NOT for adjusting steering range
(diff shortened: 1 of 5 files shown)


Ezward:
This branch still needs have all the updated drivetrains tested.

donkeycar/templates/cfg_complete.py
```diff
@@ -177,14 +186,40 @@
...
+# - the pwm pins produce a duty cycle from 0 (completely LOW)
+#   to 1 (100% completely high), which is proportional to the
+#   amount of power delivered to the motor.
+# - in forward mode, the reverse pwm is 0 duty_cycle,
+#   in backward mode, the forward pwm is 0 duty cycle.
+# - both pwms are 0 duty cycle (LOW) to 'detach' motor and
+#   and glide to a stop.
+# - both pwms are full duty cycle (100% HIGH) to brake
+#
+HBRIDGE_PIN_FWD = 18,       ß# provides forward duty cycle to motor
```
DocGarbanzo, line 218: Here is an unintentional `beta` as well as some unintentional commas here and below.

Ezward: Thank you for catching that.

DocGarbanzo (approved):
Looks good to me.

DocGarbanzo (approved):
LGTM - thank you.

PR #1002, opened 2022-03-26
Set LOG_LEVEL environment variable to set log level

Ezward:
Minor change to allow the root logger level to be set with an environment variable.  This makes it easy to run a different log level without changing code or configuration.

```python
LOG_LEVEL=DEBUG python manage.py drive
```


DocGarbanzo (approved):
Super idea.

PR #1003, opened 2022-03-26
921 tachometer odometer kinematics parts

Ezward:
Add new encoder, tachometer, odometer and kinematics parts
- the encoder/tachometer/odometer break up the these concerns into separate layers so we can avoid hoisting all that functionality into each separate encoder.
    - encoders simply count ticks
    - tachometers take ticks and output revolutions 
    - odometers take revolutions and output distance and velocity
- We support either one odometer for car-like vehicles or two encoders/odometers for differential robots.
- the kinematics are used to estimate the vehicle pose (x, y, heading).  This will eventually be used in two ways
   - we can use this as a source of pose in the path_follow template.  TBD in a separate pull request.
   - we can use this, along with closed loop speed control, to implement a velocity based neural network model.  TBD in a separate pull request.
- car-like vehicles uses bicycle kinematics
- differential drive vehicles uses unicycle kinematics

All of these parts have been run-tested for a few months in another branch using a highly modified path_follow.py.  That branch has a lot of changes, so I've created this pull request just for the encoder/odometer/kinematics pipeline.  Future pull requests will 1) add changes to path_follow to bring it up to snuff with the complete template 2) add closed-loop speed control and a velocity-based neural network model.

This is draft because I still need to test it on track.

### DIFF
--- a/arduino/mono_encoder/mono_encoder.ino
+++ b/arduino/mono_encoder/mono_encoder.ino
@@ -1,0 +1,213 @@
+/* 
+ * mono_encoder.ino
+ * 
+ * Read a single-channel encoder,
+ * debounce the input.  Ticks 
+ * are counted for each transition
+ * from open-to-closed and closed-to-open.
+ * 
+ * For example, a 20 slot optical rotary encoder
+ * will have be 40 ticks per revolution.
+ * 
+ * Implements the r/p/c command protocol 
+ * for on-demand sending of encoder value
+ * and continuous sending with provided delay.
+ * See the comment in the loop below for details.
+ * 
+ * Sends the encoder ticks and a timestamp
+ * as a comma delimited pair: ticks,timeMs
+ *
+ */
+#include <Arduino.h>
+ 
+#define ENCODER_PIN   (7)          // input pin for first encoder
+#define ENCODER_2_PIN (9)          // input pin for second encoder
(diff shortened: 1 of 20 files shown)


Ezward:
Status is that the kinematics are producing a repeatable path, but it is not accurate.  so the strange results is that if you record a path and put it in path follow mode then it will looks like it is doing a great job.  However, if you look at the path (I've added the ability to save path as .csv) in something like https://www.csvplot.com/  the path will not look anything like what your recorded.  So the path looks totally wrong, but because it is repeatably and reliably wrong in the same way each time the PID path follow algorithm can still 'follow' the wrong path and get the correct real-world output.

We've added in support for odometry/kinematics in the simulator and it seems to produce a similar issue.  NOTE that all of these tests are currently with a differential encoder setup.

I've checked and rechecked the math; it looks correct; same math as I have used in other projects and their paths look correct.

My current thinking is that we don't have a fast enough update loop.  The encoders and kinematics are running in non-threaded mode, so locked in at 20hz, which may be too slow for good pose estimation.  The kinematics formulas assume that vehicles has driven in a straight line between estimates; we may be build error very quickly because we break this assumption.  This theory is supported by the behavior we see; the path look correct if you just go straight, but creates weird loops when attempting to turn.

Ezward:
Superceded by PR https://github.com/autorope/donkeycar/pull/1089

PR #1004, opened 2022-03-30
Fix torch imports for rpi

DocGarbanzo:
# Fix pytorch import issue created in #986

There is an issue with the RPi version after merging of #986. The `pytorch` imports are not moved into the corresponding classes / functions, such now pytorch becomes a dependency when running code, like loading a model. This is non-wanted behaviour, because on RPi we don't install pytorch by default, and causes the `manage.py drive --model ` to fail. This fix moves the imports into the relevant classes/functions and fixes the issue.

### DIFF
--- a/donkeycar/parts/interpreter.py
+++ b/donkeycar/parts/interpreter.py
@@ -5,7 +5,6 @@ import numpy as np
 from typing import Union, Sequence, List
 
 import tensorflow as tf
-import torch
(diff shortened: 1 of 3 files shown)


francisduvivier:
Hi, I tried this patch and I can also confirm that this fix is working for me on the raspberry pi 4b with 32 bit buster for driving with my model.

Issue #996, opened 2022-03-09
JPG files don't recording in the data folder


Anton890304:
I found the issue in Donkey Car version 4.3.5 from the dev branch. Dataset writing doesn't work. JPG files don't recording in the data folder.
In the Donkey Car web interface, the (Start Recording [r]) button doesn't  switch to recording mode. At that time in the JupyterLab terminal I see only: `INFO:donkeycar.parts.web_controller.web:{'recording': False}`
When toggling recording on/off on the gamepad in the JupyterLab terminal, I see only: `recording: False`

- Hardware - JetsonNano
- myconfig.py parameter `AUTO_RECORD_ON_THROTTLE = True`

DocGarbanzo:
Hi @Anton890304, are you also driving? Because you will only record images when you use throttle. Also, what do you mean by JupyterLab terminal?

Anton890304:
> Привет @Anton890304 , ты тоже за рулем? Потому что вы будете записывать изображения только при использовании газа. Кроме того, что вы подразумеваете под терминалом JupyterLab?

I push the throttle, but recording doesn't work.
Yes i use JupyterLab + web interface - :8887/drive

Anton890304:
@DocGarbanzo 
I push the throttle, but recording doesn't work.
Yes i use JupyterLab + web interface - :8887/drive

BrianHenryIE:
I have the same issue on a fresh Donkey install on Jetson Nano (not a fresh Ubuntu install).

Same problem existed with PS4 controller, but this output is web controller. The only changes in `myconfig.py` are the camera and `PCA9685_I2C_BUSNUM = 1`.


```
(env) brian@briannano:~/projects/donkeycar$ python ~/mycar/manage.py drive 
________             ______                   _________              
___  __ \_______________  /___________  __    __  ____/_____ ________
__  / / /  __ \_  __ \_  //_/  _ \_  / / /    _  /    _  __ `/_  ___/
_  /_/ // /_/ /  / / /  ,<  /  __/  /_/ /     / /___  / /_/ /_  /    
/_____/ \____//_/ /_//_/|_| \___/_\__, /      \____/  \__,_/ /_/     
                                 /____/                              

using donkey v4.3.7 ...
WARNING:donkeycar.parts.pins:pigpio was not imported.
loading config file: /home/brian/mycar/config.py
loading personal config over-rides from myconfig.py
INFO:__main__:PID: 9427
cfg.CAMERA_TYPE CSIC
[...]

PR #1007, opened 2022-04-05
- fix web controller recording state so it is latched.

Ezward:
- fix for https://github.com/autorope/donkeycar/issues/996
- version="4.3.10"

### DIFF
--- a/donkeycar/parts/web_controller/web.py
+++ b/donkeycar/parts/web_controller/web.py
@@ -249,13 +249,14 @@ class WebSocketDriveAPI(tornado.websocket.WebSocketHandler):
 
     def on_message(self, message):
         data = json.loads(message)
-
         self.application.angle = data.get('angle', self.application.angle)
         self.application.throttle = data.get('throttle', self.application.throttle)
         if data.get('drive_mode') is not None:
             self.application.mode = data['drive_mode']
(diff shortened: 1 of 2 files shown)


donkeycar/parts/web_controller/web.py
```diff
@@ -249,13 +249,14 @@ def open(self):
...
     def on_message(self, message):
         data = json.loads(message)
-
         self.application.angle = data.get('angle', self.application.angle)
         self.application.throttle = data.get('throttle', self.application.throttle)
         if data.get('drive_mode') is not None:
             self.application.mode = data['drive_mode']
             self.application.mode_latch = self.application.mode
-        self.application.recording = data.get('recording', self.application.recording)
+        if data.get('recording') is not None:
```
DocGarbanzo, line 257: Just a comment for future improvements. Because you are already getting the entry from the dict, you could use it in the next line, w/o getting it again, i.e.
```python
recording = data.get('recording') 
if recording is not None:
    self.application.recording = recording
    self.application.recording_latch = self.application.recording
```

or even
```python
recording = data.get('recording') 
if recording is not None:
    self.application.recording_latch = self.application.recording = recording
```

DocGarbanzo (approved):
Looks good, thanks for fixing this.

Issue #1005, opened 2022-04-03
4.3.6.1 is not recording data


Ezward:
version: tag/4.3.6.1
RaspberryPi 4
Relevant config:
- `CAMERA_TYPE = "PICAM"`
- `DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE"`
- `CONTROLLER_TYPE = 'xbox'`
- `JOYSTICK_MAX_THROTTLE = 1 `
- `AUTO_RECORD_ON_THROTTLE = True`
- `JOYSTICK_DEADZONE = 0.1`
- `RECORD_DURING_AI = False`
- `AUTO_CREATE_NEW_TUB = False`

Ran with a fresh install of 4.3.6.1.  I had auto-save with throttle on; the configuration in myconfig.py that causes recording whenever the throttle is non-zero: `AUTO_RECORD_ON_THROTTLE = True`.  I drove a number of laps to create new data.  during that time I used the 'x' button to delete mistakes.  The results was that it created the catalog and images folder, but collected no data.  **The catalog was zero bytes and the images folder was empty.**

zlite:
I'm running 4.3.8 and it's saving the data correctly for me

DocGarbanzo:
Thanks @zlite - @Ezward, this is your #989 which fixes it then. That's the only 'real' difference between the two version (i.e. HEAD of main and 4.3.6.1)

Ezward:
This is actually a duplicate of another issue and that issue has a proposed fix https://github.com/autorope/donkeycar/issues/996#issuecomment-1075800626

Ezward:
> I'm running 4.3.8 and it's saving the data correctly for me

@zlite What controller are you using when you test with 4.3.8?

DocGarbanzo:
@Ezward - do you suggest that I should use the patch in #996 in the 4.3.6 branch instead of PR #989? I can do that.

Ezward:
The problem is addressed by this pull request https://github.com/autorope/donkeycar/pull/1007 which is now merged.

Ezward:
[...]

PR #1009, opened 2022-04-10
1005 fix recording yet again

Ezward:
Fixes issue https://github.com/autorope/donkeycar/issues/1005  I tested this extensively with xbox controller.   
- while I was at it I changed logging to use a logger rather than print statements.


DocGarbanzo (approved):
LGTM

BrianHenryIE:
Seems to be working for me with PS4 controller. (just a quick test with the car on my desk, but the .jpgs are in the folder, whereas before they were not :)

PR #1011, opened 2022-04-22
Fixed bug introduced in PR #997

DocGarbanzo:
The change was missing the addition of the `--myconfig` to the argument of the `train` command.


Ezward (approved):
Glad you caught this.  Thanks.

PR #1014, opened 2022-05-15
Update NMEA parser to handle bad chars

Ezward:
- it turns out that it is common for the first set of
  data returned by a gps module to include incorrectly
  framed characters that break the unicode conversion.
- This change ignores exceptions thrown by the
  unicode decoder and keeps on going until it
  can read a correctly framed line.

### DIFF
--- a/donkeycar/parts/gps.py
+++ b/donkeycar/parts/gps.py
@@ -181,7 +181,12 @@ def getGpsPosition(line, debug=False):
     nmea_msg = line[1:-3]      # msg without $ and *## checksum
     nmea_parts = nmea_msg.split(",")
     message = nmea_parts[0]
-    if (message == "GPRMC") or (message == "GNRMC"):   
+
+    if (message == "GPRMC") or (message == "GNRMC"):
+        if True or debug:
+            print()
+            print(line)
+
         #     
         # like '$GPRMC,003918.00,A,3806.92281,N,12235.64362,W,0.090,,060322,,,D*67'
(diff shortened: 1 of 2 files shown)


donkeycar/parts/gps.py
```diff
@@ -178,7 +181,12 @@ def getGpsPosition(line, debug=False):
     nmea_msg = line[1:-3]      # msg without $ and *## checksum
     nmea_parts = nmea_msg.split(",")
     message = nmea_parts[0]
-    if (message == "GPRMC") or (message == "GNRMC"):   
+
+    if (message == "GPRMC") or (message == "GNRMC"):
+        if True or debug:
```
DocGarbanzo, line 186: This is always True - why do we need the `if`. Also we are printing, not logging into the shell, is this intended. I would expect to just have:
```python
logger.debug(line)
```
so if we set global logging to Debug, it will log, otherwise is won't.

donkeycar/parts/gps.py
```diff
@@ -57,6 +57,9 @@ def _readline(self) -> str:
                     return self.gps.readline().decode()
             except serial.serialutil.SerialException:
                 pass
+            except UnicodeDecodeError:
```
DocGarbanzo, line 60: This is the real fix, this looks good.

donkeycar/parts/gps.py
Ezward: I deleted it; it was left over from finding the problem.

PR #1017, opened 2022-05-31
Fix passing of tub metadata

DocGarbanzo:
# Template complete.py does not support passing tub metadata

According to the docopt string, the complete template should allow to read tub metadata as 
```
--meta=foo:bar --meta=baz:44
```
but it doesn't. This change fixes it.

### DIFF
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -848,8 +848,9 @@ if __name__ == '__main__':
     if args['drive']:
         model_type = args['--type']
         camera_type = args['--camera']
+        meta_kv = [tuple(s.split(':')) for s in args['--meta']]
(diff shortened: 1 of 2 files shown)


cfox570:
Instead of changing the complete.py template, I recommend you change Tub_v2.   Look at datastore.py from Version 3.  The split code is in the init procedure for Tub.    If the meta conversion is handled here, the meta data could be setup in config.py or myconfig.py and passed to the tub part instead of only relying on the command line parameter.

DocGarbanzo:
@cfox570 - I have changed the code to allow what you suggested, i.e. merging `METADATA` from the config with the template arguments. I hope you find that useful.

cfox570:
Perfect.  Thanks for the update.

PR #1018, opened 2022-06-01
Adding VESC support to DonkeyCar

sisaha9:
Still pending final test run (previous test runs were more bruteforced). Is there anything else that needs to be added to this PR? What's the best place to add documentation? 

From Triton AI (Specifically @ivanferrier 's effort)


Ezward:
This is very exciting; thank you for creating this.  What will we get for documentation?  With just a donkey part we will get a tremendous number of questions in discord on how to make this work.  We need docs.donkeycar.com to have good documentation for this drive train.  My methodology is to create a parallel branch in the [donkeydocs repo](https://github.com/autorope/donkeydocs) that we can merge when we merge this PR.   It should include whatever setup kind of setup the user needs to make with the vesc_tool to make this work.  Are there any good videos or other online resources that we can point to?   Thanks again.

sisaha9:
@Ezward just opened a PR in https://github.com/autorope/donkeydocs/pull/39

sisaha9:
Also tested the changes on the car and we can smoothly control with VESC using this code

donkeycar/parts/actuator.py
```diff
@@ -166,6 +166,60 @@ def set_pulse(self, pulse):
     def run(self, pulse):
         self.set_pulse(pulse)
 
+class VESC:
+    ''' 
+    VESC Motor controler using pyvesc
```
Ezward, line 171: Thanks for the great header.

donkeycar/templates/cfg_complete.py
```diff
@@ -221,6 +221,16 @@
...
 STEERING_RIGHT_PWM = 290   # pwm value for full right steering (use `donkey calibrate` to measure value for your car)
 
+#VESC controller, primarily need to change VESC_SERIAL_PORT  and VESC_MAX_SPEED_PERCENT
+VESC_MAX_SPEED_PERCENT =.2  # Max speed as a percent of the actual speed
+VESC_SERIAL_PORT= "/dev/ttyACM0" # Serial device to use for communication. Can check with ls /dev/tty*
+VESC_HAS_SENSOR= True # Whether or not the bldc motor is using a hall effect sensor
+VESC_START_HEARTBEAT= True # Whether or not to automatically start the heartbeat thread that will keep commands alive.
+VESC_BAUDRATE= 115200 # baudrate for the serial communication. Shouldn't need to change this.
+VESC_TIMEOUT= 0.05 # timeout for the serial communication
+VESC_STEERING_SCALE= 0.5 # VESC accepts steering inputs from 0 to 1. Joystick is usually -1 to 1. This changes it to -0.5 to 0.5
```
Ezward, line 231: I understand from the comments that VESC_STEERING_SCALE and VESC_STEERING_OFFSET work to get to scale and translate the joystick input from -1..1 to 0..1 so it is compatible with VESC.  My question is, should this even be configurable?  Can we just hard code this (or at assign default to the related arguments and don't expose any configuration).

Ezward (changes requested):
I have a question regarding steering scaling.  Also, the PR indicates this has not been fully tested, 'Still pending final test run'.  How can we get it tested?  Finally, how much of this assumes an F1Tenth setup?  What if I bring my own car with my own motor?

donkeycar/parts/actuator.py
sisaha9: No problem! It was written by @ivanferrier

donkeycar/templates/cfg_complete.py
sisaha9: Mentioned in the other PR but this should be configurable since the web UI has a default mapping of 0 to 1 which is what the VESC expects (which means the scale and offset will both be 0)

sisaha9:
> I have a question regarding steering scaling. Also, the PR indicates this has not been fully tested, 'Still pending final test run'. How can we get it tested? Finally, how much of this assumes an F1Tenth setup? What if I bring my own car with my own motor?

Will be seeing if we can modify it to non F1Tenth cars (we usually don't use their XML file). That being said this was tested on our cars and it worked. So it is no longer pending the final test run. Had mentioned it here: https://github.com/autorope/donkeycar/pull/1018#issuecomment-1145128788

donkeycar/templates/cfg_complete.py
Ezward: What web UI are you talking about?  If the http://<yourcar>.local::8887/drive is producing only 0..1 values for throttle then that is a bug.  It should act like a game controller and produce -1 to 1.

Ezward (approved):
Thanks for creating this part.  I see in the docs it is clear what it has been tested on, so that is good enough for me.  If I find anything different when I setup my car I'll add it to the docs.

sisaha9:
@DocGarbanzo I have rebased and bumped the version number. For the `donkeydocs` repo we will be making updated instructions to work on any car during the summer and will make a PR there later

PR #1019, opened 2022-06-09
Added 4WD for DC Motor with L298N Configuration

Vrroom:
4 DC Motors controlled by 2 L298N motor controllers is one of the easiest and cheapest configurations to build. For example, in this build, I simply stuck the motors and screwed the motor controllers on a wooden plank. On the other side, I stuck a raspberry pi, a battery and a voltage converter (using double sided tape, **not** glue). I know the end result doesn't look pretty but it works. 

IMG_0422
IMG_0421

With this teeny-tiny contribution, new users will be able to build the cheapest RC car and be able to use it  with the rest of donkey car awesomeness right of the bat. If you guys think this is useful and would like me to improve the commit, please let me know. I'm willing to spend time documenting this (wherever the right place is).

### DIFF
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -656,6 +656,36 @@ def drive(cfg, model_path=None, use_joystick=False, model_type=None,
         V.add(left_motor, inputs=['left_motor_speed'])
         V.add(right_motor, inputs=['right_motor_speed'])
 
+    elif cfg.DRIVE_TRAIN_TYPE == "DC_FOUR_WHEEL_L298N":
+        dt = cfg.DC_FOUR_WHEEL_L298N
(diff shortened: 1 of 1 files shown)


DocGarbanzo:
@Vrroom - I think this is a great idea, because here you build the car from scratch. We have documentation in the `donkey_docs` repo and it should be added there. It would also help if you can show the whole thing works - you might link to a short video seeing the car drives autonomously for example.

Vrroom:
@DocGarbanzo - sorry for the late response. I'm currently training the autopilot. It doesn't work well in practice. I tried it on a track in a nearby park and it kept going off it. I'll get that done and contribute the details to `donkey_docs`. Thanks for your response!!

PR #1029, opened 2022-07-07
add "sudo" to create symbolic link for locale.h in /usr/lib

dinesh-rt:


Ezward (approved):
Looks good.  Thanks for this.

PR #1031, opened 2022-07-26
Add support for TFMini Lidar

mgagvani:
Works with [TFMini](https://www.sparkfun.com/products/retired/14588) and [TFMini Plus](https://www.sparkfun.com/products/15179). Tested on Raspberry Pi. 

Let me know if anything should be added/changed.

### DIFF
--- a/donkeycar/parts/tfmini.py
+++ b/donkeycar/parts/tfmini.py
@@ -1,5 +1,6 @@
 import time
 import serial
+import logging
 
 class TFMini:
     """
@@ -14,13 +15,14 @@ class TFMini:
         self.poll_delay = poll_delay
 
         self.dist = 0
-        self.strength = 0
 
-        if self.ser.is_open == False:
+        if not self.ser.is_open:
             self.ser.close() # in case it is still open, we do not want to open it twice
             self.ser.open()
 
-        print("Init TFMini")
+        self.logger = logging.Logger("TFMini Logger")
+
+        self.logger.info("Init TFMini")
         time.sleep(init_delay)
 
     def update(self):
@@ -42,12 +44,11 @@ class TFMini:
 
                     if strength > 0:
                         self.dist = dist
-                        self.strength = strength
 
                     self.ser.reset_input_buffer()
 
         except Exception as e:
-            print(e)
+            self.logger.error(e)
 
 
     def run_threaded(self):
@@ -68,3 +69,4 @@ if __name__ == "__main__":
         print(i, data)
 
     lidar.shutdown()
+    
--- a/donkeycar/parts/tfmini.py
+++ b/donkeycar/parts/tfmini.py
@@ -1,0 +1,68 @@
+import time
+import serial
+
+class TFMini:
+    """
+    Class for TFMini and TFMini-Plus distance sensors.
+    See wiring and installation instructions at https://github.com/TFmini/TFmini-RaspberryPi
+    """
+
+    def __init__(self, port="/dev/serial0", baudrate=115200, poll_delay=0.01, init_delay=0.1):
+        self.ser = serial.Serial(port, baudrate)
+        self.poll_delay = poll_delay
+
+        self.dist = 0
+        self.strength = 0
+
+        if self.ser.is_open == False:
+            self.ser.close() # in case it is still open, we do not want to open it twice
+            self.ser.open()
+
+        print("Init TFMini")
+        time.sleep(init_delay)
+
+    def update(self):
+        while self.ser.is_open:
+            self.poll()
+            if self.poll_delay > 0:
+                time.sleep(self.poll_delay)
+
+    def poll(self):
+        try:
+            count = self.ser.in_waiting
+            if count > 8:
+                recv = self.ser.read(9)   
+                self.ser.reset_input_buffer() 
+
+                if recv[0] == 0x59 and recv[1] == 0x59:     
+                    dist = recv[2] + recv[3] * 256
+                    strength = recv[4] + recv[5] * 256
+
+                    if strength > 0:
+                        self.dist = dist
+                        self.strength = strength
+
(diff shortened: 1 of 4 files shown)


donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+        self.poll_delay = poll_delay
+
+        self.dist = 0
+        self.strength = 0
+
+        if self.ser.is_open == False:
+            self.ser.close() # in case it is still open, we do not want to open it twice
+            self.ser.open()
+
+        print("Init TFMini")
```
DocGarbanzo, line 23: Please convert print statements into logger statements, e.g. here:
```python
logger.info("Init TFMini")
```

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Returns distance in centimeters. 
+    """
+
+    def __init__(self, port="/dev/serial0", baudrate=115200, poll_delay=0.01, init_delay=0.1):
+        self.ser = serial.Serial(port, baudrate)
+        self.poll_delay = poll_delay
+
+        self.dist = 0
+        self.strength = 0
```
DocGarbanzo, line 17: This member variable is not really used. Does it make sense to return it - is this valuable information? Below you only take readings if `strength > 0`. Otherwise we should probably remove it.

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+
+        if self.ser.is_open == False:
```
DocGarbanzo, line 19:
```python
if not self.ser.is_open
```

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+            self.ser.open()
+
+        print("Init TFMini")
+        time.sleep(init_delay)
+
+    def update(self):
+        while self.ser.is_open:
+            self.poll()
+            if self.poll_delay > 0:
+                time.sleep(self.poll_delay)
```
DocGarbanzo, line 30: Why do we want a poll delay here? At a car speed of 4m/s this automatically creates as systematic error of 4cm, using your default value of 10ms in the constructor. I haven't seen any problems with a threaded part running the thread continuously. Did you see any problems when running w/ 0 poll delay? If not, we could safely remove that delay.

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+                recv = self.ser.read(9)   
+                self.ser.reset_input_buffer() 
+
+                if recv[0] == 0x59 and recv[1] == 0x59:     
+                    dist = recv[2] + recv[3] * 256
+                    strength = recv[4] + recv[5] * 256
+
+                    if strength > 0:
+                        self.dist = dist
+                        self.strength = strength
```
DocGarbanzo, line 45: As mentioned above, we are not doing anything with that member variable.

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+
+                    self.ser.reset_input_buffer()
+
+        except Exception as e:
+            print(e)
```
DocGarbanzo, line 50: Please use 
```python
logger.error(e)
```

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,70 @@
...
+        self.ser.close()
+
+if __name__ == "__main__":
+    lidar = TFMini(poll_delay=0.1)
+
+    for i in range(100):
+        data = lidar.run()
+        print(i, data)
+
+    lidar.shutdown()
```
DocGarbanzo, line 70: missing new line at eof

donkeycar/templates/cfg_complete.py
```diff
@@ -350,6 +350,10 @@
 LIDAR_LOWER_LIMIT = 90 # angles that will be recorded. Use this to block out obstructed areas on your car, or looking backwards. Note that for the RP A1M8 Lidar, "0" is in the direction of the motor
 LIDAR_UPPER_LIMIT = 270
 
+# TFMINI
+HAVE_TFMINI = True
```
DocGarbanzo, line 354: This should be `False` as the majority of users will not have that part installed.

DocGarbanzo (changes requested):
This is a great addition for a high quality distance sensor part. The change looks good in general. I only have a couple of minor comments.

donkeycar/parts/tfmini.py
mgagvani: Yes, without poll delay the part would be able to output only a few values before failing. I'm not entirely sure why this is but I suspect it's because the serial port's input buffer is not clearing fast enough to run without a poll delay. The `self.ser.reset_input_buffer()` helps but doesn't completely solve it.

DocGarbanzo: @mgagvani - thanks for explaining, this is very valuable. Can you add the above explanation as a comment into the code, so we don't forget why this is important?

donkeycar/parts/tfmini.py
```diff
@@ -0,0 +1,72 @@
...
+        self.ser = serial.Serial(port, baudrate)
+        self.poll_delay = poll_delay
+
+        self.dist = 0
+
+        if not self.ser.is_open:
+            self.ser.close() # in case it is still open, we do not want to open it twice
+            self.ser.open()
+
+        self.logger = logging.Logger("TFMini Logger")
```
DocGarbanzo, line 23: Can you please move the logger outside of the class into the module scope as we have done for most of the other modules. You create the logger with:
```python
logger = logging.getLogger(__name__)
```

DocGarbanzo (changes requested):

DocGarbanzo (approved):
LGTM

PR #1033, opened 2022-08-04
Fix loss of steering in path_follow template

Ezward:
When using the path_follow template with GPS I would often lose all steering control.  I had occasionally seen this in the standard complete.py template as well.  It turns out that the steering part would throw an exception on an out of range value and would kill the steering thread.  This PR fixes that problem.

- the CTE PID algorithm in path follow was predicting steering
  values that were way out of the -1 to 1 range.
- This triggered a check that threw an exception.
- Since the part was being run in a thread, throwing the exception
  kill the thread, so so the user would lose all steering control.
- Even the auto-pilot would lose steering.
- Part 1 of fix is to not throw an exception, but instead log
  and error and return.
- Part 2 of the fix is to clamp all out of range values so
  that if the PID predicted an out of range value, it would be
  brought back into range.

### DIFF
--- a/donkeycar/parts/actuator.py
+++ b/donkeycar/parts/actuator.py
@@ -10,6 +10,18 @@ import logging
 from typing import Tuple
 
 import donkeycar as dk
+from donkeycar import utils
+from donkeycar.parts.kinematics import differential_steering
+from donkeycar.utils import clamp
+
+logger = logging.getLogger(__name__)
+
+try:
+    import RPi.GPIO as GPIO
+except ImportError as e:
+    logger.warn(f"RPi.GPIO was not imported. {e}")
+    globals()["GPIO"] = None
+
 from donkeycar.parts.pins import OutputPin, PwmPin, PinState
 from donkeycar.utilities.deprecated import deprecated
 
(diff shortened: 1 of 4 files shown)


donkeycar/parts/actuator.py
```diff
@@ -10,6 +10,18 @@
 from typing import Tuple
 
 import donkeycar as dk
+from donkeycar import utils
+from donkeycar.parts.kinematics import differential_steering
```
DocGarbanzo, line 14: Could it be that you missed to check in that module - this seems to cause the break of 3 tests: test_actuator, test_template, test_web_socket.

donkeycar/parts/actuator.py
```diff
@@ -915,19 +887,18 @@ def shutdown(self):
...
 # so this is redundant
 #
 @deprecated("This will be removed in a future release in favor of PulseController")
 class RPi_GPIO_Servo(object):
     '''
     Servo controlled from the gpio pins on Rpi
     '''
-    def __init__(self, pin, freq = 50, min=5.0, max=7.8):
-        import RPi.GPIO as GPIO
+    def __init__(self, pin, pin_scheme=GPIO.BCM, freq = 50, min=5.0, max=7.8):
```
DocGarbanzo, line 899: Why do you use frequency 50 here, whereas we use 60 everywhere else for servos, ESCs?

DocGarbanzo (changes requested):
Please also bump the version in `setup.py`

donkeycar/parts/actuator.py
Ezward: I don't use 50 there.  That is legacy code that will eventually be removed.

donkeycar/parts/actuator.py
Ezward: Thanks for catching this.  the import of differential_steering is not necessary; in the branch I am picking from it have been move to the kinematics parts, but not in this branch.  I've removed the import.

Ezward:
I don't think I broke those tests.  It seems to be some async issue.  I can see the value is set correctly in the web socket application, but at the time it is asserted in the test it is still the Mock.

donkeycar/parts/actuator.py
```diff
@@ -166,60 +180,6 @@ def set_pulse(self, pulse):
     def run(self, pulse):
         self.set_pulse(pulse)
 
-class VESC:
```
Ezward, line 169: I rebased against main.  For some reason it shows me deleting the VESC class.  What is up with that?

Ezward: I just copied it back in from main to fix this issue.

Ezward:
The mqtt test passes on my mac with python 3.9.  
```
======================================================================== test session starts ========================================================================
platform darwin -- Python 3.9.10, pytest-7.1.1, pluggy-1.0.0
rootdir: /Volumes/APFS256/projects/donkeycar/donkeycar/tests, configfile: pytest.ini
plugins: cov-3.0.0
collected 1 item                                                                                                                                                    

tests/test_telemetry.py::test_mqtt_telemetry PASSED                                                                                                           [100%]
```

donkeycar/parts/actuator.py
DocGarbanzo: Ok, thanks you're right.

DocGarbanzo (approved):
LGTM - I think we have to ignore the mamba warnings for now. We know it's not as safe as conda, but the download speed using conda has become so slow as you remember, so we don't have a choice atm.

PR #1034, opened 2022-08-13
Fix calibrate script

DocGarbanzo:
This fixes the `calibrate.py` script that is copied into the car directory and connects to the web server's calibration page. The script was still using the old specification of PWM pins, but the complete template has been upgraded to the new definitions already. This PR addresses that inconsistency. We also remove some unfortunate code form the script. There is still a bad issue w/ monkeypatching going on.

### DIFF
--- a/donkeycar/templates/calibrate.py
+++ b/donkeycar/templates/calibrate.py
@@ -62,22 +62,23 @@ def drive(cfg ):
         # and as second PwmPin for throttle (ESC)
         #
         from donkeycar.parts.actuator import PWMSteering, PWMThrottle, PulseController
+        dt = cfg.PWM_STEERING_THROTTLE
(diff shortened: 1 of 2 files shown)


donkeycar/templates/calibrate.py
```diff
@@ -113,22 +115,14 @@ def drive(cfg ):
         drive_train = RoboHATDriver(cfg)
         V.add(drive_train, inputs=['angle', 'throttle'])
 
-
+    # TODO: monkeypatching is bad!!!
```
Ezward, line 118: I think those values are not even used anywhere in the controller or here in the calibration template.  Maybe some other code uses them; you right - this is very hacky.

Ezward (approved):
This looks great.  Thank you for finding and fixing this.

PR #1035, opened 2022-08-13
Fix tensorflow training problem on RTX 30 series card.

briansp2020:
* Update ubuntu.yml and windows.yml to install latest version of tensorflow (2.9.1)

* Update setup.py to remove tensorflow version

### DIFF
--- a/install/envs/ubuntu.yml
+++ b/install/envs/ubuntu.yml
@@ -8,6 +8,8 @@ channels:
 
 dependencies:
   - python=3.7
+  - cudatoolkit=11.2
+  - cudnn=8.1.0
   - h5py
   - pillow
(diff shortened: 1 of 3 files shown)


briansp2020:
This fixed my problem with training using RTX 3080ti
https://discord.com/channels/662098530411741184/671604367706554388/1006776313585082408

DocGarbanzo:
A TF upgrade will require to move to py3.9. This requires also an upgrade of RPi OS to bullseye which still has some open issues regarding the PiCamera support on the new OS. The py/tf upgrade is definitely on the Todo list. However, your problem mentioned above is very likely unrelated to neither the python nor tf version, pls see my reply on discord.

Issue #991, opened 2022-03-06
Add gps part and integrate into path_follow template


Ezward:
The path follow template is currently dedicated to the T265 tracking camera.  We would like to make it a more general localization template.  We are also running outdoor races now, so the perfect addition would be a gps part that can produce positions in meters.
- [x] create gps part that can produced timestamped positions in meters
- [x] Integrate the gps readings into the path follow template
  - [x] use gps positions as pos/x and pos/y  
  - [x] create a mode in path_follow to record waypoints that allows capturing of multiiple gps positions, then calculates a centroid and guassian.   Recording of the waypoint as a centroid and a  guassian, so we can use the probability to decide if we have reached the endpoint.
- [x] merge into main branch (see  https://github.com/autorope/donkeycar/issues/991#issuecomment-1242763724)
  - [x] https://github.com/autorope/donkeycar/pull/1042
  - [x] https://github.com/autorope/donkeycar/pull/1046

Ezward:
Here is a pull request for the gps part https://github.com/autorope/donkeycar/pull/992

Ezward:
There is more work on gps in this branch; https://github.com/autorope/donkeycar/tree/921-path-follow-in-simulator.  We've added another [gps part](https://github.com/autorope/donkeycar/blob/050009fbf09675e5034bbe04763187ca234528db/donkeycar/parts/gps.py#L18)  which delivers the most recent position reading, rather the list of readings since the last update.  This is then integrated into the [path follow](https://github.com/autorope/donkeycar/blob/050009fbf09675e5034bbe04763187ca234528db/donkeycar/templates/path_follow.py#L345) template.   I've gotten a standard donkeycar (Ackerman steering, servo/esc) to do a shitty job of recording and following a path, but it does follow it.  The path looks correct; I probably just need better PID parameters.

Ezward:
I have the gps following the path (still need to do a better job on PID, but it is clearly following the line), but it get's lost when it get's back to the start.  This gives a little insight.  The only device that does not get lost when it get's back to the start is the T265; I believe that is because the T265 has loop closure build into it's algorithm.  So we need to fix this; if we are back at the origin point then we should restart the path explicitly.

Ezward:
Here is an issue to improve the path recording and path follow.  https://github.com/autorope/donkeycar/issues/1025

Ezward:
Path follow with GPS has been successfully implemented and tested.  Now it is time to get it merged into main.

[...]

PR #1042, opened 2022-09-16
Move adding camera, controller, drivetrain into functions

Ezward:
This is the first step in updating the path_follow.py template to use GPS.  This step is just to make the same sensor available in complete.py available to any template, and to simplify the complete.py template as well.  The full process for merging in the GPS branch can be found in [Issue 991}(https://github.com/autorope/donkeycar/issues/991#issuecomment-1242763724)
 
- refactored complete.py to move adding of common parts into functions that can be shared with other templates (path follow in particular).
     - add_drivetrain() adds the configured drive train to the vehicle.
     - add_odometry() adds configured encoders to the vehicle.
     - get_camera()/add_camera() adds configured camera(s) to the vehicle.
     - add_simulator() adds donkey gym if it is configured.
     - add_user_controller() adds the webui and any configured game controller, or the donkey hat.

### DIFF
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -14,14 +14,19 @@ Options:
     --myconfig=filename     Specify myconfig file to use. 
                             [default: myconfig.py]
 """
-import os
-import time
-import logging
 from docopt import docopt
+
+#
+# import cv2 early to avoid issue with importing after tensorflow
+# see https://github.com/opencv/opencv/issues/14884#issuecomment-599852128
+#
+try:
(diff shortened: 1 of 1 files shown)


DocGarbanzo (approved):
Looks good to me. It's probably hard to test, because you can have all possible combinations of controllers, drive-trains, cameras, etc. Did you test some of the common combinations?

Ezward:
I am going to run this branch this weekend at the diyrobocars race and if all goes well I will merge.  Thank you for the approval @DocGarbanzo

PR #1045, opened 2022-10-10
UI bug fixes and improvements

DocGarbanzo:
# Improvements and fixes in Donkey UI
* When no tub or config is loaded, clicking transport buttons in the tub manager or pilot screen lead to ungraceful program aborts. Similarly clicking on augmentation controls. This is now handled better.
* Fixes in logger statements that lead to exceptions
* Improvement in the handling of the filter functionality in the tub manager screen: This filter will now be exported into the config such that it gets picked up in the training. This allows to use the filter in training without manually editing the config file.


Ezward (approved):
LGTM

PR #1049, opened 2022-10-16
Fix test_web_socket on OSX by creating mock outside of assignment.

DocGarbanzo:
It seems that creating the `Mock` outside of the assignment fixes the instabilities in pytest.


Ezward (approved):
LGTM.  Thanks.

PR #1052, opened 2022-10-18
fix donkey gym bug

sctse999:
Simulator is broken because SIM_RECORD_DISTANCE and SIM_RECORD_ORIENTATION is not present in config.py

Temp comment out until further fix


Ezward (approved):
That's a merge bug on my part.  I have an experimental branch of the simulator that outputs accumulated distances so we can simulate odometry.  So those values are not present in the main simulator.  This fix is fine for removing them.

Issue #1053, opened 2022-10-22
fix auto-record on throttle for robothat MM1 controller


Ezward:
Reported by user `fooOO` in [discord]( https://discord.com/channels/662098530411741184/671604048218292233/1032996525497929748)

"Hi, Has anyone using RoboHAT MM1 and donkeycar 4.3.23? As I have configed the myconfig to record with throttle, but seems like it's not triggering the recording.
...
I did follow the instructions to give the right permission and disabled the nvgetty and I could use RC to control the car and in the Jetson nano serial monitor also output correct pwm info, I think it might be caused by something updated in the donkeycar 4.x, as it used to work with donkeycar 3.x. 

BTW I have tried by using the start record button on the web console it seems working properly, so I might just stick to this method."

Ezward:
I reviewed the code and it looks like we did not address the robothat controller when we added the code to make the game controllers and web controller cooperate with setting drive mode (user/autosteering/autopilot) and recording mode.  So I ported the code from the joystick controller into the robothat controller.  That is in branch [robothat-auto-record-on-throttle](https://github.com/autorope/donkeycar/tree/robothat-auto-record-on-throttle)

wallarug:
Hey @Ezward ,

It's been a few years now since I added this functionality.   So what exactly is still broken that may need my review?

It looks like you have already got a patch/fix in place.

CC: @peterpanstechland (foo00)

Ezward:
@wallarug Thanks for commenting.  We added code a while back to make the game controllers and web controller cooperate when changing the drive mode and recording mode.  We did not address the robothat at that time.  This PR https://github.com/autorope/donkeycar/pull/1055 fixes that by porting the code from the game controller to the robothat controller.  Basically we need to latch any  change that is made by the user so it can overrwrite the incoming state in run_threaded() sometime later.  This branch has been tested and shown to fix the issue.

Ezward:
Original poster reports that is is fixed https://discord.com/channels/662098530411741184/671604048218292233/1034481825202049157

PR #1055, opened 2022-10-22
Robothat auto record on throttle

Ezward:
Address issue https://github.com/autorope/donkeycar/issues/1053 

Robothat controller was not updated when we added code to make the game controllers and web controller cooperate.  I ported the recording latching code to the robot hat controller.

### DIFF
--- a/donkeycar/parts/robohat.py
+++ b/donkeycar/parts/robohat.py
@@ -32,7 +32,6 @@ class RoboHATController:
         self.angle = 0.0
         self.throttle = 0.0
         self.mode = 'user'
-        self.mode_latch = None
         self.recording = False
         self.recording_latch = None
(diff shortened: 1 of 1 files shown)


TCIII:
@Ezward,

I switched to the `robothat-auto-record-on-throttle` branch, performed a `git pull`, completed a `donkey createcar --path ~/mycar --overwrite` and proceeded to test if `AUTO_RECORD_ON_THROTTLE = True` would result in waypoints (records) to be recorded when the RC controller throttle initiated forward motion.

Yes, the `robothat-auto-record-on-throttle` branch does record waypoints when `AUTO_RECORD_ON_THROTTLE = True`.

TCIII

DocGarbanzo (approved):
LGTM

Issue #1043, opened 2022-10-09
Update path_follow algorithm to handle cross-over paths


Ezward:
The current path_follow algorithm searches for the nearest waypoint to it's current position, then chooses the next waypoint in the path and uses those two points to create a line; it then determines which side of the line it is on and how far from the line it is and uses that as the input to a PID that decides which direction to turn in and how much to turn.  

If the path crosses over itself or comes near itself (with the waypoint recording distance) then it is possible for the algorithm to get confused.  You can verify this by recording a simple '8' shaped path and try to follow it.  Sometimes it will do the wrong thing at the cross-over.

Modify the algorithm so that it can handle cross over paths.

Ezward:
merged https://github.com/autorope/donkeycar/pull/1058

PR #1058, opened 2022-11-07
1043 Handle crossing paths in path follow

Ezward:
Issue https://github.com/autorope/donkeycar/issues/1043

Update the path_follow template so that it can handle paths that have cross-overs.  The current code tests every point in the path to find the closest point.  This means that it is possible that a point on a cross-over going in the opposite direction may get selected.  The approach to fix this is to make the search start at the most recent closest point and limit the search to a configurable number of points.  So for instance, consider a perfectly symmetrical figure 8 path with 1000 waypoints in the path whose origin is at the top of the figure 8; then the intersection as at point 250 and point 750.  If we limit the search to 250 points then only 1/4 of the path will be searched.  If we have driving to the intersection from then origin, then if we are at the the 250th point.  We would then search from point 250 to point 500, and so we never get close the other intersection point at 750.

To implement this we updated these parts:

Update CTE part
- uses a different algorithm to find the closest point.  
- It allows for an offset to start the search and a maximum number of points to search.
- It allows the number of points to look ahead and look behind when returning the track that should be followed.  This allows the algorithm to adjust smoothing and allows a longer look ahead to handle shorter distances between waypoints and/or faster driving speeds.
- The run method now returns the index of the nearest point so we can keep track of it and use it as the offset when searching for the nearest point.

Update OriginOffset part
- run method takes the index of the closest point and returns it, so that it can reset it to the start of the path when the origin is reset.

We then updated the path_follow.py template to use the updated parts and we updated cfg_path_follow.py to include configuration for number of search points, as well as the number of look ahead and look behind points used to calculate the track to follow.


donkeycar/tests/test_path.py
```diff
@@ -56,3 +56,175 @@ def test_csvpath_load(self):
...
+        self.assertEqual(((0, 1), 0, 1), cte.nearest_pt(path, 1, 1))
+        self.assertEqual(((0, 50), 49, 1), cte.nearest_pt(path, 1, 50))
+        self.assertEqual(((0, 100), 99, 1), cte.nearest_pt(path, 1, 100))
+
+    def test_nearest_pt_from_pt(self):
+        # path containing [(0,1)..(0,100),(0,1)..(0,100)]
+        path = []
+        for i in range(100):
+            path.append((0, i+1))
+        for i in range(100):
```
DocGarbanzo, line 103: `path += path`?

DocGarbanzo: Actually you have this path in each test. You can make this a pytest fixture.

DocGarbanzo (commented):
This looks good to me, I have only very minor comments.

donkeycar/parts/path.py
```diff
@@ -113,15 +113,21 @@ class OriginOffset(object):
...
+        self.ox = 0.0
+        self.oy = 0.0
         self.last_x = 0.0
         self.last_y = 0.0
+        self.reset = None
 
-    def run(self, x, y):
+    def run(self, x, y, closest_pt):
+        """
+        :param:x is current horizontal position
```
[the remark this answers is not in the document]
Ezward: Thanks.

donkeycar/parts/path.py
```diff
@@ -136,7 +142,16 @@ def run(self, x, y):
             pos = (self.last_x - self.ox, self.last_y - self.oy)
         if self.debug:
             print(f"pos/x = {pos[0]}, pos/y = {pos[1]}")
```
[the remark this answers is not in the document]
Ezward: Yes.

donkeycar/parts/path.py
```diff
@@ -136,7 +142,16 @@ def run(self, x, y):
...
-        return pos
+
+        # reset the starting search index for cte algorithm
+        if self.reset:
+            print(f"cte/closest_pt = {closest_pt} -> None")
```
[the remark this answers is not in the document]
Ezward: Yes, also.

donkeycar/parts/path.py
```diff
@@ -227,6 +243,38 @@ def run(self, img, x, y):
...
+
+    #
+    # Find the index of the path element with minimal distance to (x,y).
+    # This prefers the first element with the minimum distance if there
+    # are more then one.
+    #
+    def nearest_pt(self, path, x, y, from_pt=0, num_pts=None):
+        from_pt = from_pt if from_pt is not None else 0
+        num_pts = num_pts if num_pts is not None else len(path)
+        num_pts = num_pts if num_pts <= len(path) else len(path)
```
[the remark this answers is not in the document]
Ezward: Dun

donkeycar/tests/test_path.py
Ezward: It' not worth the effort to make a fixture; my rule is once something is used in 3 places then you have enough info to decide how to generalize.  But even then, this is 5 lines of code so 1) is it worth the effort, what are we gaining, 2) does it couple code in a way that makes it more understandable (the coupling is a logical outcome of the code's structure and relationship) or does it couple code that is otherwise unrelated, in which case the coupling makes the code harder to maintain, not easier.  In this case the fact that the two tests use the same path is incidental.

Ezward: I did the path += path.

Ezward:
@DocGarbanzo thank you for those comments.  I updated the code and responded in comments.

donkeycar/tests/test_path.py
DocGarbanzo: Yeah I see, but now comes my point above, because you have the same code in 3 places, it's now changed in two places only. You can actually write the whole thing like `path = list((0, i+1) for i in range(100)) * 2` to please the gods of python.

DocGarbanzo (approved):

PR #1064, opened 2022-11-26
Add IMU_ADDRESS configuration

Ezward:
The MPU6050 and MPU9250 allow for two different I2C addresses based on how an input line is set.  Our code does not support the secondary address, which is how the MM1 hat is setup.  This code allows those IMUs to be used with their secondary address.
- move construction of IMU part and addition to vehicle into a method in complete.py and make complete.py and path_follow.py use it.
- add IMU_ADDRESS defaulted to 0x68 to cfg_complete.py and cfg_path_follow.py
- make path follow use add_imu()


DocGarbanzo (approved):
This looks good to me. I suggest we bump the version in your other commit if we merge this as well today.

PR #1066, opened 2022-11-30
separate pytorch from default conda installation

sctse999:
pytorch and fastai dependencies could be installed later by issuing

`conda env update --file install/envs/pytorch.yml --prune`

I will update the doc after this PR go through.


DocGarbanzo:
We have to think about setting up GH actions for that, as atm it doesn't seem to be possible to create the GH action conda env from multiple yaml files. In order to be able to continue supporting ptyorch based models we cannot neglect the corresponding CI tests.

PR #1067, opened 2022-11-30
Pin protobuf to 3.20.3

sctse999:


DocGarbanzo (approved):

sisaha9:
@DocGarbanzo @sctse999 what was the reason for this PR? In a GPU based container restricted to Ubuntu 18.04 and CUDA 10.1 as required by tf 2.2.0 the pinned version does not exist

PR #1069, opened 2022-12-11
Fix imu input in the complete template. 

DocGarbanzo:
The imu keras part expects an array for the imu data as a second input to the model. Creating a 3-liner vectorizer part to do this. This is a follow up from the discussion on discord: https://discord.com/channels/662098530411741184/1046926296116101180/1047206276066717746.


Ezward (approved):
This is a good solution.  Thanks Dirk.

PR #1075, opened 2022-12-17
Develop

salamisodikiolawale:
Add docker compose feature and update dockerfile with command **RUN mkdir notebooks**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,9 @@
+version: '3.8'
+
+services:
+
+  donkeycar:
+    build: .
+    ports:
+      - "8887:8887"
+      - "8888:8888"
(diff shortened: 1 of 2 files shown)


DocGarbanzo:
@Heavy02011 - can you have a look at that?

Ezward (changes requested):
Latest donkeycar release (5.x) requires python 3.9.x, so we should change the base image in the Dockerfile.

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3.8'
```
haslinghuis, line 1:
```suggestion
version: '3.9'
```

docker-compose.yml
```diff
@@ -0,0 +1,9 @@
...
+
+services:
+
+  donkeycar:
+    build: .
+    ports:
+      - "8887:8887"
+      - "8888:8888"
```
haslinghuis, line 9:
```suggestion
      - "8888:8888"

```

PR #1078, opened 2022-12-26
Support for Picamera2

Vrroom:
[Picamera2](https://github.com/raspberrypi/picamera2) is the library for camera interop on 64bit RPi OS. I created a part for this because the current PiCamera part crashes. 

I tested this part with `python manage.py drive` and got the expected result in the web ui on the browser. An added benefit of moving to 64bit OS is that we can run faster inference of PyTorch models (see [example](https://pytorch.org/tutorials/intermediate/realtime_rpi.html)). 

I'll be around to polish this pull request if this is something desired by you guys. Thanks for this project!

### DIFF
--- a/donkeycar/parts/camera.py
+++ b/donkeycar/parts/camera.py
@@ -86,6 +86,68 @@ class PiCamera(BaseCamera):
         self.camera = None
 
 
+class PiCamera2(BaseCamera):
+    def __init__(self, image_w=160, image_h=120, image_d=3, framerate=20, vflip=False, hflip=False):
+        from picamera2 import Picamera2
+        import libcamera
+
+        self.camera = Picamera2()
+        # create configuration
+        image_format = 'BGR888' if (image_d == 3 or image_d == 1) else 'XBGR8888'
+        config = self.camera.create_preview_configuration({
+            'format': image_format, 
+            'size': (image_w, image_h)
+        })
+        # set camera transformations
+        if vflip and hflip : 
+            config['transform'] = libcamera.Transform(hflip=1, vflip=1)
+        elif vflip: 
+            config['transform'] = libcamera.Transform(vflip=1)
+        elif hflip:
+            config['transform'] = libcamera.Transform(hflip=1)
+        
+        self.camera.configure(config)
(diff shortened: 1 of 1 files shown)


Vrroom:
@Ezward, @DocGarbanzo - requesting feedback

DocGarbanzo:
Hi @Vrroom,

Thanks for your contribution. We are actually very close to pushing a full upgrade of DC to py 3.9, tf 2.9 and RPi bullseye 64 bit with the new PiCamera2 support. It would be great if you could check/compare our Picamera2 implementation with yours, and create a PR against the `tf_2_9` branch, if you find we missed something or could improve the current version. The branch contains all the changes.

Ezward:
@DocGarbanzo I like this part; the issue with the tf_2_9 branches version is that it replaces the old Picamera part that works on older OS with the new part; so if a user wants to run on an older OS they cannot.  Might we want both?

DocGarbanzo:
@Ezward - we currently only support one version of RPi OS for one version of DC. If you want to run an older RPi OS, you will have a an older version of python as well and won't be able to upgrade to python 3.9 and tf 2.9 anyway. So you would be staying on DC 4.4.

Vrroom:
@DocGarbanzo I went through the code you referenced and it is quite similar to mine. Great to know that this is already out there. What is stopping merging this branch with main? Perhaps I can help out. 

Also, I like the idea of setting expectations on DonkeyCar development. I agree that there should be one version of PiCamera Part only.

DocGarbanzo:
@Vrroom - We are still in the late stages of making changes to the branch, because it will contain support for Nvidia Jetpack 5.0.2 and 4.6.2, and the latter one has turned out to be more tricky than we thought.  We are trying to address all issues that we have seen from users trying to run Donkey Car on the Jetson. In particular w/ JP 5.0.2 we will be forward-compatible to the new Jetson Orin (at least, that's the plan!). Feel free to check out that branch and test yourself, as long as you don't install it on Jetson Nano, everything should be working.

Vrroom:
Ok, thanks! I'll try it out

PR #1079, opened 2022-12-31
Train script missing transfer parameter

cfox570:
The train.py script was missing the parameter to pass a transfer model.  This caused the comment to be interpreted as a transfer model and caused an error.  

1. Added a command line parameter to optionally specify transfer model
2. Updated docopt usage script
3. Updated call to pipeline.training.train()
4. Also added code to specify tubs as "data/" if not specified on command line.  The usage string specifies that the tubs parameter is optional.

### DIFF
--- a/donkeycar/templates/train.py
+++ b/donkeycar/templates/train.py
@@ -4,9 +4,11 @@ Scripts to train a keras model using tensorflow.
 Basic usage should feel familiar: train.py --tubs data/ --model models/mypilot.h5
 
 Usage:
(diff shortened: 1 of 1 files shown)


DocGarbanzo:
@cfox570 - thanks for pointing out the issues. Actually the `train.py` script is legacy and will be deleted in the next version. The official training has been switched to the command `donkey train` some time ago, where the transfer model can be passed, please see: https://docs.donkeycar.com/guide/deep_learning/train_autopilot/

cfox570:
Okay. Then no need to merge.

Issue #1086, opened 2023-01-18
Unable to use Lidar in DonkeyCar Sim


sisaha9:
I get the following error whenever I try enabling recording Lidar.

```
Traceback (most recent call last):
  File "/donkeycar/donkeycar/vehicle.py", line 154, in start
    self.update_parts()
  File "/donkeycar/donkeycar/vehicle.py", line 200, in update_parts
    outputs = p.run_threaded(*inputs)
  File "/donkeycar/donkeycar/parts/dgym.py", line 89, in run_threaded
    outputs += [self.info['lidar']]
KeyError: 'lidar'
```

Here is my config file

```
DONKEY_GYM = True
# DONKEY_SIM_PATH = "remote"
DONKEY_SIM_PATH = "remote"
SIM_RECORD_LOCATION = True
SIM_RECORD_GYROACCEL = True
SIM_RECORD_VELOCITY = True
SIM_RECORD_LIDAR = True
DONKEY_GYM_ENV_NAME = "donkey-warren-track-v0"
#DONKEY_GYM_ENV_NAME = “donkey-mountain-track-v0”
# UCSD yellow color in RGB = 255, 205, 0
# UCSD blue color in RGB = 0, 106, 150
GYM_CONF = { "body_style" : "car01", "body_rgb" : (255, 205, 0), "car_name" : "UCSD-148-YourName", "font_size" : 30} # body style(donkey|bare|car01) body rgb 0-255
GYM_CONF["racer_name"] = "UCSD-148-YourName"
GYM_CONF["country"] = "USA"
GYM_CONF["bio"] = "Something_about_you, ex: Made in Brazil"
#
# SIM_HOST = "donkey-sim.roboticist.dev"
SIM_ARTIFICIAL_LATENCY = 0
SIM_HOST = "127.0.0.1"              # when racing on virtual-race-league use host "roboticists.dev"
#SIM_ARTIFICIAL_LATENCY = 30          # Use the value when you ping roboticists.dev. When racing on virtual-race league, use 0 (zero)

# When racing, to give the ai a boost, configure these values.
AI_LAUNCH_DURATION = 3            # the ai will output throttle for this many seconds
AI_LAUNCH_THROTTLE = 1            # the ai will output this throttle value
AI_LAUNCH_KEEP_ENABLED = True      # when False ( default) you will need to hit the AI_LAUNCH_ENABLE_BUTTON for each use. This is safest. When this True, is active on each trip into "local" ai mode.
#
[...]

PR #1088, opened 2023-01-18
Fixing initialization of self.info

sisaha9:
Fixes https://github.com/autorope/donkeycar/issues/1086

### DIFF
--- a/donkeycar/parts/dgym.py
+++ b/donkeycar/parts/dgym.py
@@ -34,7 +34,8 @@ class DonkeyGymEnv(object):
                      'cte': 0,
                      'gyro': (0., 0., 0.),
                      'accel': (0., 0., 0.),
-                     'vel': (0., 0., 0.)}
+                     'vel': (0., 0., 0.),
+                     'lidar': []}
         self.delay = float(delay) / 1000
         self.record_location = record_location
         self.record_gyroaccel = record_gyroaccel


DocGarbanzo:
@Maximellerbach - can you have a look at this?

Maximellerbach:
@DocGarbanzo the type of self.info['lidar'] should be a np.array, I'm not sure of the expected behavior when the array is empty ? Are we checking somewhere if the array of point is empty / matches the expected shape the model needs ?

DocGarbanzo:
> @DocGarbanzo the type of self.info['lidar'] should be a np.array, I'm not sure of the expected behavior when the array is empty ? Are we checking somewhere if the array of point is empty / matches the expected shape the model needs ?

Currently the lidar data is not used in any models. For now it's just being recorded into the tub, as it would be returned in the `run_threaded` method. Here, we would need a native python list, but not an `np.array`.

Maximellerbach:
> 

Maybe I'm missing something, why should it be a python list ? I am currently returning from GymDonkeycar a numpy array, if you confirm it should be a list then I will also change it on this side

Ezward:
I think the older drivers return an np.array() and the newer drivers return a python list.  I would go with the newer drivers.  This is what the new driver puts out: https://github.com/autorope/donkeycar/blob/c4d90910e0e1b5056dc8961a09c7e5eb1ba18ffe/donkeycar/parts/lidar.py#L171

Maximellerbach:
Ok ! I will then align with those.
What does the time and scan values look like ?

Maximellerbach (approved):

DocGarbanzo (approved):
Thanks @sisaha9

Issue #1085, opened 2023-01-14
Upgrade the computer vision template


Ezward:
The cv_control.py/cfg_cv_control.py template files implement a line following algorithm using opencv and traditional computer vision techniques.  Currently the template only supports the raspberrypi camera and the a pwm steering/throttle drivetrain.  So it does not support the various cameras and drivetrains supported by the complete.py and path_follow.py template.  In addition the LineFollower part is hard-coded into the template, so it is not easy to replace.  

We would like to upgrade the template;
- [x] use the add_camera() method in complete.py to add support for the full range of supported cameras.
- [x] use the add_drivetrain() method in complete.py to add support for the full range of supported drivetrains
- [x] use the add_controller() method in complete.py to add support for the full range of game controllers, including the web controller.
- [x] separate out the LineFollower part into it's own source file in the parts folder.
- [x] add logic that allows a computer vision part to be selected and loaded at runtime, so the user can create their own part without having to edit the LineFollower part.  We do something similar with the custom joystick; if a user has created a custom joystick using the Joystick Wizard then they have a file named MyJoystick.py in their mycar application folder.  If they choose "custom" as the JOYSTICK_TYPE in myconfig.py, then it will load that part and use it as the joystick.
- [x] add configuration for the LineFollower part; you can see the hard-coded parameters the constructor
```
        self.scan_y = 60   # num pixels from the top to start horiz scan
        self.scan_height = 10 # num pixels high to grab from horiz scan
        self.color_thr_low = np.asarray((0, 50, 50)) # hsv dark yellow
        self.color_thr_hi = np.asarray((50, 255, 255)) # hsv light yellow
        self.target_pixel = None # of the N slots above, which is the ideal relationship target
        self.steering = 0.0 # from -1 to 1
        self.throttle = 0.15 # from -1 to 1
        self.recording = False # Set to true if desired to save camera frames
        self.delta_th = 0.1 # how much to change throttle when off
        self.throttle_max = 0.3
        self.throttle_min = 0.15

        # These three PID constants are crucial to the way the car drives. If you are tuning them
        # start by setting the others zero and focus on first Kp, then Kd, and then Ki.
        self.pid_st = PID(Kp=-0.01, Ki=0.00, Kd=-0.001)

```
[...]

PR #1091, opened 2023-02-10
1085 upgrade computer vision template

Ezward:
Issue https://github.com/autorope/donkeycar/issues/1085

The cv_control.py/cfg_cv_control.py template files implement a line following algorithm using opencv and traditional computer vision techniques.  Currently the template only supports the raspberrypi camera and the a pwm steering/throttle drivetrain.  So it does not support the various cameras and drivetrains supported by the complete.py and path_follow.py template.  In addition the LineFollower part is hard-coded into the template, so it is not easy to replace.  

We would like to upgrade the template;
- [x] use the add_camera() method in complete.py to add support for the full range of supported cameras.
- [x] use the add_drivetrain() method in complete.py to add support for the full range of supported drivetrains
- [x] use the add_controller() method in complete.py to add support for the full range of game controllers, including the web controller.
- [x] separate out the LineFollower part into it's own source file in the parts folder.
- [x] add logic that allows a computer vision part to be selected and loaded at runtime, so the user can create their own part without having to edit the LineFollower part.  We do something similar with the custom joystick; if a user has created a custom joystick using the Joystick Wizard then they have a file named MyJoystick.py in their mycar application folder.  If they choose "custom" as the JOYSTICK_TYPE in myconfig.py, then it will load that part and use it as the joystick.
- [x] add configuration for the LineFollower part; you can see the hard-coded parameters the constructor
- [x] add configuration to optionally draw the algorithm's output on an output image; see LineFollower.debug_display().  We need to make the image an output parameter.  If the user has chosen to update the image, then do it and return it; if not then just return the input image.  Note that the current code just draws on the image then shows it in an opencv windows; it does not produce a new image and return it. it is better to create a new image and leave the original intact and then decide which image will be shown in the webui, so the user can see it in realtime.
- [x] support the assignable web ui buttons; allow the user to assign various functions to the web ui buttons via configuration; like we do in path_follow.py template.  This is particularly important if the user is using a standard RC controller that does not have buttons.  It is also useful to allow the user to remap the controls for any joystick/game controller.
  - [x] Add functions to control various aspects/configuration of the template; like the pid parameters, the min and max throttle, etc.


donkeycar/parts/line_follower.py
```diff
@@ -0,0 +1,124 @@
...
+        '''
+
+        mask_exp = np.stack((mask, ) *3, axis=-1)
+        iSlice = self.scan_y
+        img = np.copy(cam_img)
+        img[iSlice : iSlice + self.scan_height, :, :] = mask_exp
+        img = cv2.cvtColor(img, cv2.COLOR_RGB2BGR)
+
+        display_str = []
+        display_str.append("STEERING:{:.1f}".format(self.steering))
```
DocGarbanzo, line 111: In general, we prefer f-strings.

donkeycar/parts/web_controller/web.py
```diff
@@ -365,10 +367,16 @@ async def get(self):
         while True:
 
             interval = .01
-            if served_image_timestamp + interval < time.time() and \
-                    hasattr(self.application, 'img_arr'):
```
DocGarbanzo, line 369: I believe you are doing this, because two different web servers are using the VideoAPI. Wouldn't it be better to make `img_arr` a member in both of them?

donkeycar/templates/cfg_cv_control.py
```diff
@@ -1,58 +1,497 @@
...
+TELEMETRY_DONKEY_NAME = 'my_robot1234'
+TELEMETRY_MQTT_TOPIC_TEMPLATE = 'donkey/%s/telemetry'
+TELEMETRY_MQTT_JSON_ENABLE = False
+TELEMETRY_MQTT_BROKER_HOST = 'broker.hivemq.com'
+TELEMETRY_MQTT_BROKER_PORT = 1883
+TELEMETRY_PUBLISH_PERIOD = 1
+TELEMETRY_LOGGING_ENABLE = True
+TELEMETRY_LOGGING_LEVEL = 'INFO' # (Python logging level) 'NOTSET' / 'DEBUG' / 'INFO' / 'WARNING' / 'ERROR' / 'FATAL' / 'CRITICAL'
+TELEMETRY_LOGGING_FORMAT = '%(message)s'  # (Python logging format - https://docs.python.org/3/library/logging.html#formatter-objects
+TELEMETRY_DEFAULT_INPUTS = 'pilot/angle,pilot/throttle,recording'
```
DocGarbanzo, line 431: Did you intend to change this to `pilot/steering` - will it show as a conflict if we merge the other PR first?

donkeycar/templates/cfg_path_follow.py
```diff
@@ -41,6 +41,7 @@
 CAMERA_INDEX = 0  # used for 'WEBCAM' and 'CVCAM' when there is more than one camera connected
 # For CSIC camera - If the camera is mounted in a rotated position, changing the below parameter will correct the output frame orientation
 CSIC_CAM_GSTREAMER_FLIP_PARM = 0 # (0 => none , 4 => Flip horizontally, 6 => Flip vertically)
+BGR2RGB=False  # true to convert from BRG format to RGB format; requires opencv
```
DocGarbanzo, line 44: Maybe `BGR2RGB = False`

donkeycar/templates/cv_control.py
```diff
@@ -158,48 +47,174 @@ def drive(cfg, args):
...
+            elif mode == 'local_angle':
+                return pilot_angle if pilot_angle else 0.0, user_throttle
 
+            else:
+                return pilot_angle if pilot_angle else 0.0, \
+                       pilot_throttle * cfg.AI_THROTTLE_MULT \
+                           if pilot_throttle else 0.0
+
+    V.add(DriveMode(),
+          inputs=['user/mode', 'user/angle', 'user/throttle',
```
DocGarbanzo, line 187: `user/steering` rather?

donkeycar/utils.py
```diff
@@ -190,20 +190,46 @@ def load_image(filename, cfg):
...
+        # If the PIL image is greyscale, the np array will have shape (H, W)
+        # Need to add a depth channel by expanding to (H, W, 1)
+        if img.mode == 'L':
+            h, w = img_arr.shape[:2]
+            img_arr = img_arr.reshape(h, w, 1)
+
+        return img_arr
+
+    except Exception as e:
+        print(e)
```
DocGarbanzo, line 229: Use logging instead of printing?

donkeycar/parts/line_follower.py
Ezward: I just left what was there.

donkeycar/parts/web_controller/web.py
Ezward: I don't remember doing anything here.  Let me look into it.  I don't know what this is supposed to be doing, but it is also a delete.  So this must have come from a rebase  somehow.

donkeycar/templates/cfg_cv_control.py
Ezward: I did mean to make is steering.  The term angle is a misnomer and it get's confusing when when try to actually model a real steering angle, so I renamed it steering.  I will rebase this, then I will change this terminology in all the templates.

donkeycar/templates/cv_control.py
Ezward: The version of add_user_control() in this branch outputs 'user/angle', so we need to do that here.

donkeycar/utils.py
Ezward: done.  thanks.

donkeycar/parts/web_controller/web.py
Ezward: Ok, I see.  I changed the code to show a placeholder image if not camera image is available.  The full code change is:
```
            if served_image_timestamp + interval < time.time():
                #
                # if we have an image, then use it.
                # otherwise show placeholder
                #
                if hasattr(self.application, 'img_arr') and self.application.img_arr is not None:
                    img = utils.arr_to_binary(self.application.img_arr)
                else:
                    img = utils.arr_to_binary(placeholder_image)
```

donkeycar/templates/cfg_cv_control.py
DocGarbanzo: OK

DocGarbanzo (approved):
Thanks @Ezward - looks all good to me.

DocGarbanzo (approved):

PR #1102, opened 2023-02-26
Added hsv_picker.sh script to support cv template

Ezward:
This adds a script that allows a user to select an area on an image and sample the low and high HSV value to create a mask, like is used in the cv_control template.

We have found in testing that it can be difficult to figure out the HSV value range to use to detect the line and reject background.  This script makes it easier by allow the user to sample the video image.


DocGarbanzo (approved):
Thanks Ed.

PR #1105, opened 2023-02-26
1104 fix user angle in deep learning

Ezward:
I added a serious bug in complete.py when I renamed the steering output from `'user/angle'` to `'user/steering'`.  It turns out the the literal string 'user/angle' is used all over the place in the deep learning pipeline and DonkeyUI, so we can't change it.  

This fix the fixes complete.py so it uses 'user/angle' and 'pilot/angle'.  This has been tested doing data collection, training and inference in autopilot mode.

The fix works by renaming the 'user/steering' value to 'user/angle' so both are in memory.  I found and fixed a bug in the Pipe() part that happens when renaming a single argument; Python's variable argument facility outputs a tuple with two elements when only a single argument was passed; I now check for that and explicitly send back the single argument.

### DIFF
--- a/donkeycar/parts/pipe.py
+++ b/donkeycar/parts/pipe.py
@@ -2,12 +2,7 @@ class Pipe:
     """
     Just pipe all inputs to the output, so they can be renamed.
     """
-    def __init__(self):
-        self.running = True
-
     def run(self, *args):
-        if self.running:
-            return args
-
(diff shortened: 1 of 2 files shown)


zlite (approved):

zlite:
LGTM. Nice use of Pipe to copy between two namespaces

PR #1107, opened 2023-02-27
Dgym refactor

Maximellerbach:


donkeycar/parts/dgym.py
```diff
@@ -20,28 +20,64 @@ def __init__(self, sim_path, host="127.0.0.1", port=9091, headless=0, env_name="
...
+            'lap_count': 'lap_count',
+        }
+
+        self.output_keys = {}
+
+        try:
+            for key, val in cfg.SIM_RECORD.items():
+                if cfg.SIM_RECORD[key]:
+                    outputs_key = self.info_keys[key]
+                    if isinstance(outputs_key, list):
```
DocGarbanzo, line 72: I would prefer to not unravel these vectors/lists but to return them as such. So vel, gyro, etc will be 3-d vectors. This would keep the data better together and will also making it much cleaner in passing it through the memory and into the tub.

donkeycar/templates/complete.py
```diff
@@ -748,30 +748,12 @@ def add_simulator(V, cfg):
...
-            outputs += ['lidar/dist_array']
-        # if cfg.SIM_RECORD_DISTANCE:
-        #     outputs += ['dist/left', 'dist/right']
-        # if cfg.SIM_RECORD_ORIENTATION:
-        #     outputs += ['roll', 'pitch', 'yaw']
+        outputs = ['cam/image_array'] # modified in DonkeyGymEnv constructor according to the config
+        gym = DonkeyGymEnv(cfg, outputs)
 
-        V.add(gym, inputs=inputs, outputs=outputs, threaded=threaded)
+        V.add(gym, inputs=inputs, outputs=outputs, threaded=True)
```
DocGarbanzo, line 756: If we add the gym with only `outputs = ['cam/image_array'], how can we access all the other gym outputs in the car memory? They would not be there, right?

DocGarbanzo (commented):
This is a great change improving the interface to the donkey gym. A couple of comments how to improve the return values from the gym to stay in vectors instead of components.

donkeycar/templates/complete.py
Maximellerbach: the output list is actually modified in the __init__ of the DonkeyGymEnv part, I know it is kind of a bad practice, should I put this in a separate function to be able to build seperately the output list ?

donkeycar/parts/dgym.py
Maximellerbach: Yep we could do that ! for the sake of backward compability I unpacked them, but I can easily change that

Maximellerbach: how would you prefer the vectors ? np.arrays or lists ?

donkeycar/templates/cfg_complete.py
```diff
@@ -567,19 +567,28 @@
 DONKEY_GYM = False
 DONKEY_SIM_PATH = "path to sim" #"/home/tkramer/projects/sdsandbox/sdsim/build/DonkeySimLinux/donkey_sim.x86_64" when racing on virtual-race-league use "remote", or user "remote" when you want to start the sim manually first.
 DONKEY_GYM_ENV_NAME = "donkey-generated-track-v0" # ("donkey-generated-track-v0"|"donkey-generated-roads-v0"|"donkey-warehouse-v0"|"donkey-avc-sparkfun-v0")
```
Maximellerbach, line 569: @DocGarbanzo while we are changing things, should we unify the names used in the myconfig file ? Some are GYM_ other SIM_, other DONKEY_GYM or both, I think it's quite confusing.
I guess changing this would break people's config ? is there a way to update them ?

setup.py
```diff
@@ -78,7 +78,9 @@ def package_files(directory, strip_leading):
               'pandas',
               'pyyaml',
               'plotly',
-              'imgaug'
+              'imgaug',
+              'gym',
+              'gym-donkeycar @ git+ssh://git@github.com/tawnkramer/gym-donkeycar.git@master#egg=gym-donkeycar',
```
Maximellerbach, line 83: This doesn't work in the CI, need to fix

donkeycar/templates/cfg_complete.py
DocGarbanzo: I would say that `SIM` should be used for all data referring to the simulation app and `GYM` should be used for everything interacting with the donkey gym. So the above looks reasonable to me. But we should clearly distinguish between them and correct it where it seems to be mixed up.

setup.py
DocGarbanzo: Can you try removing the trailing `@master...` and just put:
'gym-donkeycar @ git+ssh://git@github.com/tawnkramer/gym-donkeycar.git`

donkeycar/parts/dgym.py
DocGarbanzo: Can you make them python lists?

donkeycar/parts/dgym.py
```diff
@@ -48,34 +48,33 @@ def __init__(self, cfg, outputs):
 
         # output keys corresponding to info dict values
         self.info_keys = {
```
DocGarbanzo, line 50: Because this dict has now the structure { k: v } with v == k, we can simply turn it into a list of keys only [k].

DocGarbanzo: I see we can't there is still a map from `odom` to something else.

donkeycar/templates/complete.py
DocGarbanzo: I understand. I see your comment in 755. It's a bit tricky but I think this is fine for now.

Issue #1109, opened 2023-03-04
Refactor Image Augmentations/Transformation to use parts


Ezward:
Currently image augmentations and transformation are a fixed set.  There is no way to add a custom augmentation or transformation without changing core code.  We can do better.

This was initiated after talking to a user on Discord that wanted to add Canny edge detection as a transformation.  They ended up just hacking the camera part to apply the Canny algorithm on the way out; so they end up with transformed data in their data sets, which is not ideal.  

Goal:
- make it easy to add a custom augmentation or transformation to the training and inference pipelines.
- make it easier to maintain the training and inference pipelines; pulling code out of it and into a place where it can be more easily maintained.  As an example, we have a nasty bug caused by imgaug on the Jetson that makes it impossible to use cropping; it should be easy to work around this for even fix this.
- make code more composable and reusable.  We have a lot of parts in `cv.py` that could be used for augmentation or transformation. For instance, we have a part that will apply Canny edge detection to an image.  It takes an image in the run method and returns an image.  It was designed as a useful part for the computer vision template.  However, that should also be very easy to reuse as transformation.  
- Make it easier for community to contribute.  Regarding hacking the camera part to apply the Canny algorithm; there is no way we would accept a PR that did that, so we cannot leverage that user's innovation.
- It would be ideal is user's did not have to change their configuration that uses the current standard set of augmentations and transformations.
- This could make it easy to do this in a declarative way; like we intend to do for vehicle construction.

Tasks:
- refactor the current augmentations and transformations into parts.  The constructor will take in whatever configuration is needed.  The `run()` method will take an image and return an image.
- implement a way in the pipeline code (`augmentations.py`) to add a part with a name to the pipeline.  So a part would be constructed, then `add(name, part)` method would be called with a name and the part instance, similar to how we add a part to the vehicle loop.  The pipeline would keep a dictionary of named parts that had had been added.
- Change how we manage the augmentations and transformations.  Current the pipeline uses what is effectively a switch/case statement in [`donkeycar/pipeline/augmentations.py`](https://github.com/autorope/donkeycar/blob/08adc76ec184dd6a0dde1c8f53aa82aaa8b5204c/donkeycar/pipeline/augmentations.py#L93).  Instead we will have a dictionary and we will lookup the part using it's name.
[...]

Issue #968, opened 2021-12-17
Protect CVCAM from returning None image


Ezward:
We want to make sure cameras initialize correctly and do not return None images because that causes hard to understand errors downstream.  Instead we want to initalize the camera and check that it is returning valid frames at intialization time.  This is like the work done in https://github.com/autorope/donkeycar/pull/967

Ezward:
See PR https://github.com/autorope/donkeycar/pull/1111

PR #1111, opened 2023-03-08
1109 refactor augmentations/transformations

Ezward:
Issue https://github.com/autorope/donkeycar/issues/1109

## Transformations
Use standard parts for image transformations.
- Refactor transformations to be simple parts that take an image and return an image.  
  - add a number of parts to cv.py to support a lot of transformations.
  - image_transformations.py has a factory to create parts from a transformation name and the Config object and an class ImageTransformations that can take an array of transformation names and turn that into a transformation pipeline part.
- Update train so that it can handle a set of transformation before augmentations are appliend and another after augmentations are applied.  For instance, CROP or CANNY are better applied after augmentation.
- Update complete.py to merge transform pipelines for use at inference time.
- Can create one or more custom transformations and add them to the pipeline.  See comments in https://github.com/autorope/donkeycar/blob/1ecc26e95ed1afe4d1a6840dfb0f26902ad97504/donkeycar/parts/image_transformations.py#L105 and https://github.com/autorope/donkeycar/blob/7ed5849fb1bd5e286a8b1122ee43c0e7c246c4fa/donkeycar/templates/cfg_complete.py#L427

## Augmentations 
augmentations.py
- Refactor augmentations to use the albumentations library; this is a port of Dirk's branch.
- This requires updating to latest opencv, so install yaml and setup.py are updated.

## Deep Learning Template 
complete.py
- added the ability to show the transformed image that is used for inference when the autopilot is active.  This is similar to what we do in the computer vision template in autopilot mode.  By default this is turned off, but can be enabled by setting `SHOW_PILOT_IMAGE` to True.

## Computer Vision Parts 
cv.py
- fixed a bug in CvCam where some cameras will produce an image that is not the size requested.  Now the camera detects this and resizes if necessary.
- fixed CvCam to wait for a image during warm up; addresses Issue https://github.com/autorope/donkeycar/issues/968
- added parts to do color conversions.
- added a part that will output a given image; used in self-test but could also be used as a mock camera
- added a main() that can run most of the parts for testing

>> This should fix the imgaug bug https://github.com/autorope/donkeycar/issues/970 but that still needs to be tested in a Jetson.


Ezward:
This works on my Mac with Python 3.8, but the package test fail with;
```
  Looking for: ['python=3.7', 'numpy==1.23.4', 'h5py', 'pillow', 'opencv', 'matplotlib', 'tornado', 'docopt', 'pandas', 'pylint', 'pytest', 'pytest-cov', 'codecov', 'pip', 'imgaug', 'progress', 'paho-mqtt', 'prettytable', 'pyfiglet', 'mypy', 'pytorch', 'torchvision=0.12', 'torchaudio', 'pytorch-lightning', 'psutil', 'kivy=2.0.0', 'plotly', 'pyyaml', 'fastai', 'pynmea2', 'pyserial', 'utm', 'albumentations', 'python=3.7']
  
  
  Could not solve for environment specs
  Encountered problems while solving:
    - package numpy-1.23.4-py310h9638375_0 requires python >=3.10,<3.11.0a0, but none of the providers can be installed
  
  The environment can't be solved, aborting the operation
  

Error: The process '/usr/local/miniconda/condabin/mamba' failed with exit code 1
```

install/envs/ubuntu.yml
```diff
@@ -8,7 +8,7 @@ channels:
 
 dependencies:
   - python=3.7
-  - numpy=1.19
+  - numpy==1.23.4
```
DocGarbanzo, line 11: Why do we need this version? There seems to be no python 3.7 compatible version of numpy 1.23, this is also the reson for the CI failure.

DocGarbanzo (commented):
I will need to update the UI on the branch, otherwise the Brightness transformation will likely not work.

install/envs/ubuntu.yml
DocGarbanzo: I have now create a PR #1113 against this branch that takes care of the right usage of the new augmentations in the UI and in addition makes above changes to the conda files. Also removes opencv-headless, which is causing a clash with opencv in particular on the nano.

Ezward: I was getting runtime errors with numpy 1.19 which indicated I needed a newer version; I saw those issues after I added albumentations.  I'll go back to 1.19 and see if it works with albumentations.

Ezward: I removed the pinning of numpy and the install of opencv headless in my branch and that works.

Ezward: I have a bug in the way I instantiate the custom transformer.  It works great when driving from the mycar folder `python manage.py drive --model=models/linear.h5` (I have the python file in the root of the mycar folder), but it does not find it correctly when trying to train using the donkey command, 
```
Traceback (most recent call last):
  File "/Users/edmurphy/projects/temp/donkeycar/env/bin/donkey", line 11, in <module>
    load_entry_point('donkeycar', 'console_scripts', 'donkey')()
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/management/base.py", line 626, in execute_from_command_line
    c.run(args[2:])
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/management/base.py", line 563, in run
    args.comment)
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/pipeline/training.py", line 139, in train
    training_pipe = BatchSequence(kl, cfg, training_records, is_train=True)
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/pipeline/training.py", line 40, in __init__
    self.post_transformation = ImageTransformations(config, getattr(config, 'POST_TRANSFORMATIONS', []))
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/parts/image_transformations.py", line 15, in __init__
    self.transformations = [image_transformer(name, config) for name in names]
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/parts/image_transformations.py", line 15, in <listcomp>
    self.transformations = [image_transformer(name, config) for name in names]
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/parts/image_transformations.py", line 98, in image_transformer
    return custom_transformer(name, config)
  File "/Users/edmurphy/projects/temp/donkeycar/donkeycar/parts/image_transformations.py", line 156, in custom_transformer
    module = __import__(module_name)
ModuleNotFoundError: No module named 'my_custom_transformer'
```
That happens even when running the training from the `mycar` folder.  I'm not sure what to do about that.

Ezward: If I create a symbolic link from the python file in the `mycar` folder into the donkeycar source folder, so from the root of the donkecar repo:
```
ln -s /Users/edmurphy/projects/temp/mycar_deep_learning/my_custom_transformer.py donkeycar/my_custom_transformer.py
```

and change the module configuration to:
```
CUSTOM_CROP_MODULE = "donkeycar.my_custom_transformer"
```
Then that will work when training using 
```
donkey train --tub=data --type=linear
```
After the training is done I'll check to see if that also works when driving.  It is a pain however.

Ezward: That does work for driving as well.

DocGarbanzo:
I see the CI failing in the pipeline test where a tf tensor to numpy conversion has some problems. Looking in the logs, I see numpy version 1.21 here, whereas our last commit on main was using 1.19, and that ran successfully. We actually might need to pin that. I’m wondering if albumentations are the reason for the numpy upgrade, I will check that. I’m sure its all fine once we are on tf 2.9.

donkeycar/templates/cfg_complete.py
```diff
@@ -389,29 +390,163 @@
...
 
+# "CANNY" Canny Edge Detection tranformation
+CANNY_LOW_THRESHOLD = 60    # Canny edge detection low threshold value of intensity gradient
+CANNY_HIGH_THRESHOLD = 110  # Canny edge detection high threshold value of intensity gradient
+CANNY_APERTURE = 3          # Canny edge detect aperture in pixels, must be odd; choices=[3, 5, 7]
+
+# "BLUR" transformation (not this is SEPARATE from the blur augmentation)
+BLUR_KERNAL = 5        # blur kernal horizontal size in pixels
+BLUR_KERNAL_Y = None   # blur kernal vertical size in pixels or None for square kernal
+BLUR_GUASSIAN = True   # blur is gaussian if True, simple if False
```
DocGarbanzo, line 540: This is a typo but I fixed it in my PR

donkeycar/templates/cfg_complete.py
```diff
@@ -389,29 +390,163 @@
...
 ROI_TRAPEZE_MAX_Y = 120
 
+# "CANNY" Canny Edge Detection tranformation
+CANNY_LOW_THRESHOLD = 60    # Canny edge detection low threshold value of intensity gradient
+CANNY_HIGH_THRESHOLD = 110  # Canny edge detection high threshold value of intensity gradient
+CANNY_APERTURE = 3          # Canny edge detect aperture in pixels, must be odd; choices=[3, 5, 7]
+
+# "BLUR" transformation (not this is SEPARATE from the blur augmentation)
+BLUR_KERNAL = 5        # blur kernal horizontal size in pixels
+BLUR_KERNAL_Y = None   # blur kernal vertical size in pixels or None for square kernal
```
DocGarbanzo, line 539: These two should be "BLUR_KERNEL"

donkeycar/parts/cv.py
```diff
@@ -141,27 +316,30 @@ def run(self, img_arr):
...
             return None
 
     def shutdown(self):
         pass
-    
+
 
 class ImgGaussianBlur:
 
-    def __init__(self, kernal_size=5):
```
DocGarbanzo, line 155: These variable should be called "kernel_..."

DocGarbanzo (commented):
We need to get the CI tests to pass, I'm trying in my PR, too.

DocGarbanzo (approved):
Thanks @Ezward - great feature!

PR #1113, opened 2023-03-10
UI changes and yaml file updates

DocGarbanzo:
# Support albumentations in UI and small proposed correction to yaml files

* Change imgaug -> albumentations for image augmentations like blur and brightnessin the UI. 
* Update conda yaml files to remove opencv-headless which clashes w/ opencv. Also Unfreeze numpy version, as 1.23 is not supported in conda channels for py3.7


donkeycar/management/kivy_ui.py
```diff
@@ -853,14 +856,16 @@ def on_aug_list(self, obj, aug_list):
...
+        self.augmentation = ImageAugmentation(self.config, 'AUGMENTATIONS',
+                                              always_apply=True)
         self.on_current_record(None, self.current_record)
 
     def on_trans_list(self, obj, trans_list):
         if not self.config:
             return
         self.config.TRANSFORMATIONS = self.trans_list
-        self.transformation = ImageAugmentation(self.config, 'TRANSFORMATIONS')
+        self.transformation = ImageAugmentation(self.config, 'TRANSFORMATIONS',
```
Ezward, line 867: This doesn't look correct since transformations are no longer handled by ImageAugmention, but rather by ImageTransformations.  But I don't know what this method does; I don't see it called anywhere.

Ezward: Also there are now pre-augmentation and post-augmentation transformations.  So the complete list of transformations is given by 
```
transforms = getattr(cfg, 'TRANSFORMATIONS', []) + getattr(cfg, 'POST_TRANSFORMATIONS', [])
```

DocGarbanzo: `ImageAugmentation` is the class that performs the transformations. Did you rename it or create another one? This one works, I have tested it. But of course we don’t want to have two classes doing the same thing. Ill add the post transformations too.

Ezward: The point of my PR is to refactor the transformation pipeline so it can use standard parts and we can allow users to add a custom transform through configuration.  So to do this I removed transforms from ImageAugmentation and made a new class called ImageTransformations.  There are now two lists for transformations, one that runs prior to augmentation and one that runs post-augmentation.

Ezward:
Note the test failures.  I get these in my branch as well.  If you have insight into this that would be helpful.  I've looked but it is not obvious what is wrong.

Ezward:
Pinning numpy==1.19 fixes the tests.

donkeycar/management/kivy_ui.py
Ezward: See https://github.com/DocGarbanzo/donkeycar/blob/1109-refactor-augmentations-dg/donkeycar/parts/image_transformations.py

donkeycar/management/kivy_ui.py
```diff
@@ -856,16 +861,24 @@ def on_aug_list(self, obj, aug_list):
...
         self.on_current_record(None, self.current_record)
 
     def on_trans_list(self, obj, trans_list):
         if not self.config:
             return
         self.config.TRANSFORMATIONS = self.trans_list
-        self.transformation = ImageAugmentation(self.config, 'TRANSFORMATIONS',
-                                                always_apply=True)
+        self.transformation = ImageTransformations(
+            self.config, 'TRANSFORMATIONS')
```
Ezward, line 873: The way I wrote this you have to pass in the list of transformations, not just the name of the config.  I did this so I could merge the lists to get the complete list for inference;
```
transforms = getattr(cfg, 'TRANSFORMATIONS', []) + getattr(cfg, 'POST_TRANSFORMATIONS', [])
self.transformation = ImageTransformations(self.config, transforms)
```
Even here, since there are two list of tranformations, we have to merge them somehow.  So the above code would use the merged list of transformations.

Ezward: Ok, I see below there is a separate on_post_trans_list().

donkeycar/management/kivy_ui.py
```diff
@@ -856,16 +861,24 @@ def on_aug_list(self, obj, aug_list):
...
+        self.on_current_record(None, self.current_record)
+
+    def on_post_trans_list(self, obj, trans_list):
```
Ezward, line 876: Ok, I see what is going on.  Yes, then we can change the ImageTransformations to take the config name.  I can change the inference to use two parts rather than one.

Ezward:
I has missed pinning numpy==1.19 in ubunty.yaml.  That is probably why the tests are failing.  so if you rebase that should fix the tests.

Ezward:
@DocGarbanzo if the only thing hanging this up is the desire to add all the new transformations into DonkeyUI then I would say we punt that work to another PR.

Ezward:
@DocGarbanzo  I made a change to the way custom transformations are loaded.  It is in image_tranformations.py so you will need to do a rebase and merge that file before you can merge this pull request.

donkeycar/management/kivy_ui.py
DocGarbanzo: Yeah I found this a bit clunky to pass in the config and some attributes of the config. Actually we don't really need the to have flexibility in the attributes, we can just agree on `TRANSFORMATIONS` and `POST_TRANSFORMATIONS` as we are doing it in many other places, too. And doing the logic in the constructor of ImageTransformations.

donkeycar/management/kivy_ui.py
DocGarbanzo: I think it's fine to have one part that uses both lists together.

DocGarbanzo:
@Ezward - I have added the functionality in the UI and this is really useful. Say you apply CROP as a pre-transformation and CANNY as a post. You can see nicely what happens with increased/decreased Brightness and Blur. I've pulled your latest changes but I still need to resolve the CI errors.

https://user-images.githubusercontent.com/47540921/226045236-79d5ca5b-1fa9-4138-afdf-a2e4c4f01af3.mp4

Ezward:
That is seriously cool.  Thanks @DocGarbanzo.  The tests are failing because of an issue with the protobufs; presumably and tensorflow issue.  I'm ok with merging this into my branch I can see if I can deal with there.
```
/usr/local/miniconda/envs/donkey/lib/python3.7/site-packages/google/protobuf/descriptor.py:561: in __new__
    _message.Message._CheckCalledFromGeneratedFile()
E   TypeError: Descriptors cannot not be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
E   
E   More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
________________ ERROR collecting donkeycar/tests/test_torch.py ________________
```

PR #1119, opened 2023-04-04
Upgrade to python 3.9 and TF 2.9

DocGarbanzo:
# Upgrade to python 3.9 and TF 2.9 and bullseye on RPi

This PR is upgrading the DC software on all platforms:
* Host PC
* RPi
* Jetson 

Why do we chose this combination? There are some further code changes required if we move to python 3.10. In addition, TF > 2.9 will not support GPU support on native windows any longer. Hence in the next upgrade we have strongly reduced windows support. 

The code change addresses all required changes to enable the newer version of TF and the camera on RPi as well as some small general improvements. For example the Model now takes an `*args` argument as a second entry in the `run` method so we can handle model with arbitrary input data and not only two. Also the `Interpreter` classes got streamlined and simplified. 

## PC

On the PC we mainly perform the upgrade to py 3.9, TF 2.9 and kivy 2.1.

## Raspberry Pi

We now install Raspberry Pi OS Bullseye 64-bit which comes with improvements in performance and the new camera software. Inferencing the linear model also using Tensorflow 2.9 brings a speedup of about 20% compared to the previous version.

## Jetson

We are supporting Jetson Nano for DC <= 4.5, i.e. the latest version before this. This will require JP 4.5. DC version 5.0 and newer require JP 4.6 on the Jetson Nano and JP 5.0 on the Jetson Xavier. Of these three installations the JP 4.6 is the most tricky one, because we need self built tensorflow wheel, to be installed from the donkeycar/jetson repo and also self build and install our opencv with gstreamer support in order to get the camera working. Not for the faint-harted.


Ezward:
@DocGarbanzo is there a corresponding branch in the docs repo that I can use to test the install instructions?

DocGarbanzo:
> @DocGarbanzo is there a corresponding branch in the docs repo that I can use to test the install instructions? Nevermind, I see it https://github.com/autorope/donkeydocs/tree/Update_to_tf_29

Yes thats the one.

Ezward (approved):
This looks great.  Thank you @DocGarbanzo for all the hard work to get this done.  Also thanks to @TCIII for his testing and work on the Nano.

TCIII:
@DocGarbanzo,

 Some of these supported versions are incorrect in the table at the frontend of the DC Nano/Xavier installation instructions:

Jetson                    Jetpack           Python           Donkey              Tensorflow
Nano                         4.5.2                 3.6             <= 4.5.X                   2.3.1
Xavier/Orin               4.6.2                 3.8              >= 5.X                      2.9

1) JP4.5.2 is not available for downloading for the Nano, only JP4.5.1.
2) Attempting to upgrade JP4.5.1 to JP4.5.2 with the Linux upgrade command results in a corrupted bootloader which causes the Ubuntu Desktop not loading at startup.
3) I assume we are sticking with the NIDIA tensorflow 2.3.1 and not upgrading to tf2.5 as tf2.3.1 requires numpy <1.19.0,>=1.16.0 while tf2.5 requires numpy >=1.19?
4) The Xavier NX, Orin Xavier NX, and Orin Nano can use JP 5.0.x and are not limited to JP4.6.1/2.

Additionally, you still have albumentations in the main branch `setpup.py` file for the Nano running JP4.5.1/2:

`'nano45': [
              'Adafruit_PCA9685',
              'adafruit-circuitpython-ssd1306',
              'adafruit-circuitpython-rplidar',
              'Jetson.GPIO',
              'albumentations',
              'matplotlib',
              'kivy-jetson',
              'pyyaml',
              'plotly'`

Albumentations requires Python >=3.7 and JP4.5.1 and JP4.6.1/2 are using Python 3.6.9.
Also, albumentations installs opencv-python-headless which will override the opencv version installed with JP.

TCIII

PR #1127, opened 2023-05-06
Fix missing data handling

francisduvivier:
This PR is for handling the case of some missing data during data loading better.

In our case we a lot of catalogs with really bad data so we just deleted that from the manifest.

Since the start index is specified in the catalogs, it seems only correct to use it.


francisduvivier:
Tests are failing, so need some more work first. I'll reopen when ready

PR #1132, opened 2023-06-02
Update config.py

phantom-philosopher:
Added ELSE in line 56 to prevent error which occurs when the passed path is not None. If path is not None, cfg tries to load directory, because it is missing the file name at the end.

To test try running the following code (changing constants as needed): MY_PATH = r'/home/pi/mycar/'
MY_FILE = r'myconfig.py'

cfg = dk.load_config(config_path=MY_PATH, myconfig=MY_FILE)

### DIFF
--- a/donkeycar/config.py
+++ b/donkeycar/config.py
@@ -53,6 +53,8 @@ def load_config(config_path=None, myconfig="myconfig.py"):
             local_config = os.path.join(os.path.curdir, 'config.py')
             if os.path.exists(local_config):
                 config_path = local_config
+    else:
+        config_path = os.path.join(config_path, 'config.py')
     
     logger.info(f'loading config file: {config_path}')
     cfg = Config()


donkeycar/config.py
```diff
@@ -53,6 +53,8 @@ def load_config(config_path=None, myconfig="myconfig.py"):
             local_config = os.path.join(os.path.curdir, 'config.py')
             if os.path.exists(local_config):
                 config_path = local_config
+    else:
+        config_path = os.path.join(config_path, 'config.py')
```
DocGarbanzo, line 57: This doesn't look correct, as the variable `config_path` is meant to contain the full path of the `config.py` file. Either you pass this to the function, or it gets auto-generated for you.

DocGarbanzo (changes requested):

PR #1133, opened 2023-06-24
fix cv color conversions

Ezward:
- several of the color conversions were not correctly returning the converted image; returning None instead.  This caused a crash when they were used.
- this commit fixes those and makes all color conversions similar in how they return that converted image
- This also adds a new transform; a version of the trapezoid mask that takes values from the edges of the image like the crop transform does (rather than absoluted image coordinates as the pre-existing trapezoid transform does).

### DIFF
--- a/donkeycar/parts/cv.py
+++ b/donkeycar/parts/cv.py
@@ -24,8 +24,7 @@ class ImgGreyscale:
             return None
 
         try:
-            img_arr = cv2.cvtColor(img_arr, cv2.COLOR_RGB2GRAY)
-            return img_arr
+            return cv2.cvtColor(img_arr, cv2.COLOR_RGB2GRAY)
         except:
             logger.error("Unable to convert RGB image to greyscale")
             return None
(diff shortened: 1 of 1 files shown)


donkeycar/parts/cv.py
```diff
@@ -422,6 +413,62 @@ def shutdown(self):
...
+                mask = np.zeros(image.shape, dtype=np.int32)
+                points = [
+                    [self.upper_left, self.top],
+                    [width - self.upper_right, self.top],
+                    [width - self.lower_right, height - self.bottom],
+                    [self.lower_left, height - self.bottom]
+                ]
+                cv2.fillConvexPoly(mask,
+                                    np.array(points, dtype=np.int32),
+                                    self.fill)
```
haslinghuis, line 459: :space_invader: 

```suggestion
                cv2.fillConvexPoly(
                    mask,
                    np.array(points, dtype=np.int32),
                    self.fill
                )
```



```suggestion
                cv2.fillConvexPoly(mask, np.array(points, dtype=np.int32), self.fill)
```

DocGarbanzo (approved):
LGTM.

PR #1135, opened 2023-07-01
Fixing setup.py for jetson

DocGarbanzo:
Remove albumentations from jetson setup and move pyyaml into the required section, so it also installs on RPi. Also bump version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,8 +73,6 @@ setup(name='donkeycar',
               'adafruit-circuitpython-rplidar',
               'Jetson.GPIO',
               'matplotlib',
-              'kivy-jetson',
-              'plotly'
           ],
           'nano': [
               'Adafruit_PCA9685',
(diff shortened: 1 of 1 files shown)


Ezward:
Should we just replace the two augmentations with our own code?  One is guassian blur and we already have that in https://github.com/autorope/donkeycar/blob/1398eec8309499c4d54fc45d510348bcb819daff/donkeycar/parts/cv.py#L323 and the other is a randomized brightness/contrast that we could easily implement.

Ezward:
However, for now this is a good enough fix.  Thanks @DocGarbanzo

Ezward (approved):

DocGarbanzo:
> Should we just replace the two augmentations with our own code? One is guassian blur and we already have that in
> 
> https://github.com/autorope/donkeycar/blob/1398eec8309499c4d54fc45d510348bcb819daff/donkeycar/parts/cv.py#L323
> and the other is a randomized brightness/contrast that we could easily implement.
> 
> I added an issue for this; #1136

Yes, I was thinking the exact same thing. I want to google around a bit if there are other augmentations that might be useful for use, but for blur / brightness we can just use the one you already created.

PR #1145, opened 2023-09-23
Commented out unicode error logging

zlite:


zlite:
Error was due to GPS misconfiguration. This fix is no longer necessary

PR #1146, opened 2023-09-24
Updated default PIDs to those best for standard car

zlite:
Optimized and tested on standard Exceed Magnet


Ezward (approved):
Thanks for doing this; the original defaults are mental.

haslinghuis (approved):

PR #1153, opened 2023-12-12
Release 5.0 and installation with pip

DocGarbanzo:
# Release 5.0 and installation with pip

This PR is in preparation for the Donkey Car 5.0 release. Compared to current HEAD, this release has the following major changes:
* a more modern `setup.cfg` and `pyproject.toml` installation
* a migration from a full conda install towards a conda/pip install where we will be publishing a wheel file to pypi, such that users can install Donkey Car with `pip install donkeycar[XX]`, where XX contains the platform.
* removal of the `keras-viz` dependency from which we only required two utility functions which have now been copied into the code directly
* improve consistency in tub, such that adding tub with different input or metadata will not be allowed 

As part of the 5.X release we will also be updating the documentation accordingly.  The changes for the documentation are in https://github.com/autorope/donkeydocs/tree/install_with_pip.


cfox570:
I have reviewed the changes included for Release 5.0 and would like to encourage you to include changes to the MacOS setup to leverage the Mac GPU. METAL is the GPU technology that powers hardware-accelerated graphics and AI on Apple platforms. METAL  is supported by Mac computers from the year 2012.  See [METAL on Mac Computers](https://support.apple.com/en-us/102894).   Apple has released tensorflow-macos and tensorflow-metal to speedup tensorflow on the Mac. I have used this installation on both an INTEL and ARM mac for the last couple of years.   Here is a set of instructions I have used for installation on the Macbook Air M1. [Install Tensorflow-macos](https://gist.github.com/cfox570/e89603392fd48e1a60ef2f1c667e2050).    There is one difference for the ARM Mac.  For mac computers prior to 2012, the setup would use standard tensorflow.   Once you merge the files into the MAIN branch I will work through suggested changes to the code and instructions for each type of Mac. Maybe you could solicit others with Macs to test.

DocGarbanzo:
> I have reviewed the changes included for Release 5.0 and would like to encourage you to include changes to the MacOS setup to leverage the Mac GPU. METAL is the GPU technology that powers hardware-accelerated graphics and AI on Apple platforms. METAL is supported by Mac computers from the year 2012. See [METAL on Mac Computers](https://support.apple.com/en-us/102894). Apple has released tensorflow-macos and tensorflow-metal to speedup tensorflow on the Mac. I have used this installation on both an INTEL and ARM mac for the last couple of years. Here is a set of instructions I have used for installation on the Macbook Air M1. [Install Tensorflow-macos](https://gist.github.com/cfox570/e89603392fd48e1a60ef2f1c667e2050). There is one difference for the ARM Mac. For mac computers prior to 2012, the setup would use standard tensorflow. Once you merge the files into the MAIN branch I will work through suggested changes to the code and instructions for each type of Mac. Maybe you could solicit others with Macs to test.

@cfox570 - this is a very good comment. Having a quick look at your gist above, it looks like we should simply replace `tensorflow-macos==2.9` with `tensorflow-metal==0.5` in the setup.cfg for the `macos` target (matching version from here: https://pypi.org/project/tensorflow-metal/). Let's get this merged and then we can do a 5.1 release with the tensorflow metal upgrade thereafter.

cfox570:
> this is a very good comment. Having a quick look at your gist above, it looks like we should simply replace `tensorflow-macos==2.9` with `tensorflow-metal==0.5` in the setup.cfg for the `macos` target (matching version from here: https://pypi.org/project/tensorflow-metal/). Let's get this merged and then we can do a 5.1 release with the tensorflow metal upgrade thereafter.

_For Conda, the ARM Mac (M1/M2/M3) needs tensorflow-deps installed via conda not pip._ Not sure how you change your setup to accomplish.

DocGarbanzo:
> > this is a very good comment. Having a quick look at your gist above, it looks like we should simply replace `tensorflow-macos==2.9` with `tensorflow-metal==0.5` in the setup.cfg for the `macos` target (matching version from here: https://pypi.org/project/tensorflow-metal/). Let's get this merged and then we can do a 5.1 release with the tensorflow metal upgrade thereafter.
> 
> _For Conda, the ARM Mac (M1/M2/M3) needs tensorflow-deps installed via conda not pip._ Not sure how you change your setup to accomplish.

This is already part of this PR. Everything now gets installed via pip. Going from `tensorflow-macos` to `tensorflow-metal` is a one line change in `setup.cfg`.

cfox570:
To clarify, both tensorflow-macos and tensorflow-metal are required.

On Dec 13, 2023, at 3:55 PM, DocGarbanzo ***@***.***> wrote:


this is a very good comment. Having a quick look at your gist above, it looks like we should simply replace tensorflow-macos==2.9 with tensorflow-metal==0.5 in the setup.cfg for the macos target (matching version from here: https://pypi.org/project/tensorflow-metal/). Let's get this merged and then we can do a 5.1 release with the tensorflow metal upgrade thereafter.

For Conda, the ARM Mac (M1/M2/M3) needs tensorflow-deps installed via conda not pip. Not sure how you change your setup to accomplish.

This is already part of this PR. Everything now gets installed via pip. Going from tensorflow-macos to tensorflow-metal is a one line change in setup.cfg.

cfox570:
I tried out the installation of [torch].  I found two issues: 

1. replace PyTorch with just torch
2. pytorch-lightning is missing

Also I suggest you update the mac instructions to install [macos] instead of [pc]

DocGarbanzo:
> I tried out the installation of [torch]. I found two issues:
> 
>     1. replace PyTorch with just torch
> 
>     2. pytorch-lightning is missing
> 
> 
> Also I suggest you update the mac instructions to install [macos] instead of [pc]

Thanks @cfox570 - we have tested on Mac Intel with the standard TF install that is in the `[pc]` configuration and everything works fine with that. I'll give it a try to with the `[macos]` target and see if this works and makes a difference, but this is independent of the code release here.

.github/workflows/python-package-conda.yml
```diff
@@ -28,16 +24,16 @@ jobs:
...
           channel-priority: true
       - name: Conda info and list
         run: |
-          echo Environment file ${{matrix.ENV_FILE}}
           conda info
           conda list
       - name: Install donkey
-        run: pip install -e .[pc]
+        run: | 
+          pip install -e .[pc,dev]
```
Ezward, line 36: should this list include `macos`?

donkeycar/parts/tub_v2.py
```diff
@@ -78,8 +78,7 @@ def write_record(self, record=None):
         # Private properties
         contents['_timestamp_ms'] = int(round(time.time() * 1000))
         contents['_index'] = self.manifest.current_index
-        contents['_session_id'] = self.manifest.session_id
-
+        contents['_session_id'] = self.manifest.session_id[1]
```
Ezward, line 81: why the index of 1 here?  It's magic-number-ish.  Maybe worth a comment.

Ezward (approved):
Look good.  Thanks so much for this.  I'm sorry I took so long to review.  I added a couple of question comments, but nothing that blocks release.

.github/workflows/python-package-conda.yml
DocGarbanzo: Not yet. I will work on a dedicated apple silicon CI run afterwards. For now we only test on apple x86.

donkeycar/parts/tub_v2.py
DocGarbanzo: Sorry, saw that comment too late... `session_id` now returns a tuple of the (number, session_id_string) which allows to simplify the code a lot, so we are picking the string here.

PR #1168, opened 2024-03-24
Update setup for Bookworm

zlite:
Updated TF version to 2.15 to support Raspberry OS Bookworm and thus the RP5


zlite:
Replaced with a cleaner PR

PR #1169, opened 2024-03-24
Update setup.cfg to support Bookworm

zlite:
Updated TF to 2.15 on the Pi


DocGarbanzo:
@zlite - it is not that easy to do the upgrade that way because we also want to upgrade python in sync. I've create a PR #1170 for which passes the CI. Will close this here. If you can approve the other then we need to update the docs as well. Happy to do that.

PR #1170, opened 2024-03-26
Upgrade to python 3.11 and TensorFlow 2.15.X

DocGarbanzo:
# Upgrade Python and TensorFlow

After upgrading of the default Raspberry Pi OS version to bookworm which is using Python 3.11 as system python version we need to lock into that upgrade for Donkey Car. The changes here consist of:

* Upgrading the CI to use python 3.11: the CI now runs about twice as fast as before
* Upgrading and unfreezing certain python package in setup.cfg as they were tagging too old versions
* Fixing minor compatibility issues in the pytorch, telemetry and ui code as a consequence of above package upgrades

The python upgrade is improving performance by a measurable amount because 3.11 has a much faster Cpython implementation than 3.9.

### DIFF
--- a/donkeycar/parts/pytorch/torch_train.py
+++ b/donkeycar/parts/pytorch/torch_train.py
@@ -46,7 +46,7 @@ def train(cfg, tub_paths, model_output_path, model_type, checkpoint_path=None):
 
     if cfg.PRINT_MODEL_SUMMARY:
         summarize(model)
-    trainer = pl.Trainer(gpus=gpus, logger=logger, max_epochs=cfg.MAX_EPOCHS,
+    trainer = pl.Trainer(logger=logger, max_epochs=cfg.MAX_EPOCHS,
                          default_root_dir=output_dir)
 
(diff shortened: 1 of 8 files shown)


donkeycar/parts/pytorch/torch_train.py
```diff
@@ -46,8 +46,8 @@ def train(cfg, tub_paths, model_output_path, model_type, checkpoint_path=None):
 
     if cfg.PRINT_MODEL_SUMMARY:
         summarize(model)
-    trainer = pl.Trainer(gpus=gpus, logger=logger, max_epochs=cfg.MAX_EPOCHS,
-                         default_root_dir=output_dir)
+    trainer = pl.Trainer(accelerator='cpu', logger=logger,
```
haslinghuis, line 49: Could perhaps use `auto` instead of `cpu` ?

haslinghuis: Also used in `tests/test_torch.py`

PR #1172, opened 2024-03-31
Small fixes and updates for the python 3.11 upgrade

DocGarbanzo:
# Fixes and improvement for python 3.11 and TensorFlow 2.15

This PR contains a couple of small fixes for the latest python 3.11 and tensorflow 2.15.* upgrade
* The tensorrt interpreter now returns the model output as a list and not a dictionary
* Removed multiple warning from legacy logging warn function by switching to warning
* Pegged version of pytorch to version 2.1.* that works also when a graphics card is installed 
* Removed switching of gpu evaluations for fastai model, because that segfaults

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -18,10 +18,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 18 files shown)


Ezward (approved):
Looks great.  Thank you.

Issue #1177, opened 2024-05-12
I2c package needs to be update to work on bookworm


Ezward:
Users have indicated that PCA9685 is not working on RaspberryPi OS Bookworm.  This is likely because we are using a very old an deprecated library that was not updated to handle the change in how gpio is addressed in RaspberryPi OS.  

The old dependency is [Adafruit_PCA9685](https://github.com/adafruit/Adafruit_Python_PCA9685), which is marked as deprecated and suggests using [Adafruit_CircuitPython_PCA9685](https://github.com/adafruit/Adafruit_CircuitPython_PCA9685).

Adafruit_CircuitPython_PCA9685 has two pieces we need, the [PCA9685](https://github.com/adafruit/Adafruit_CircuitPython_PCA9685/blob/874545d908077baed0be8055cbbc63280a00c034/adafruit_pca9685.py#L123) class which initialized the PCA9685 bus, address and frequency, and the [PWMChannel](https://github.com/adafruit/Adafruit_CircuitPython_PCA9685/blob/874545d908077baed0be8055cbbc63280a00c034/adafruit_pca9685.py#L51C7-L51C17) class, which controls a specific servo channel on the board.

We need to update the code in two places to use the new classes.  The first is in the pin specifiers code, [pins.py](https://github.com/autorope/donkeycar/blob/main/donkeycar/parts/pins.py).  The [pins.py:PCA9685](https://github.com/autorope/donkeycar/blob/45dc516d918db177ee5e49dc75a7d6e23e6d9978/donkeycar/parts/pins.py#L563) class should now import the [Adafruit_CircuitPython_PCA9685:PCA9685](https://github.com/adafruit/Adafruit_CircuitPython_PCA9685/blob/874545d908077baed0be8055cbbc63280a00c034) class to initialize the board.  Then for each pin that is created, it will instantiate a [Adafruit_CircuitPython_PCA9685:PWMChannel]((https://github.com/adafruit/Adafruit_CircuitPython_PCA9685/blob/874545d908077baed0be8055cbbc63280a00c034/adafruit_pca9685.py#L51C7-L51C17) to control the pin.

[...]

PR #1178, opened 2024-05-13
Update PCA9685 driver to latest Adafruit CircuitPython driver

Ezward:
Fixes https://github.com/autorope/donkeycar/issues/1177

- change pi and nano install to install the new driver.
- update pins.py to use new driver to provide output and pwm pins on a PCA9685 board.
- Change the PCA9685 actuator to be a subclass of PulseController, so it inherits the underlying pin provider and pulse controller api.  This elimnates the duplicate code in pins.py and actuator.py.

- [ ] Test PCA9685 drivetrains on RaspberryPi 4
- [ ] Test PCA9685 drivetrains on RaspberryPi 5
- [ ] Test I2c OLED


Ezward:
I tested this with a RaspberryPi 5 and PCA9685 and a servo using the calibrate command.  It worked as expected.  I think we still want to test on an actual car before we release it.

Ezward:
The code as-is (commmit bca136abf130601088f89760bf635c9841084c80) has a bug in the throttle somewhere.  Steering works correctly, but throttle does not.  This is likely in the PwmThrottle part and how it interacts with the pin, possibly how the pin is started.  However, I have also found that we don't really need this update; the current code works perfectly fine on a RaspberrPi 5 using the latest RaspberryPi OS (bookworm), so I'm pausing work on this branch.

PR #1173, opened 2024-04-15
UI update

DocGarbanzo:
# Summary
## UI changes

The UI code has been in a single python and single kv file which is hard to maintain. Also, the kivy version 2.0 is by now quite outdated. The UI was crashing on RPi. Here is a refactoring that resolves these issues and adds a lot more features in addition:

1. The new ui rewrite is now separated into each source file (py and kv) for each screen
2. The ui design has been greatly reworked with a more modern layout and there is more explanation about the different features directly in the app. 
3. Several features have been upgraded and improved:
  - In the tub manager screen the timestamp record can now be viewed as human readable
  - In the trainer screen where the user can overwrite config variables one can now choose multiple variables at once and this setting is stored. So if one likes to have CACHE_IMAGES, MAX_EPOCHS and BATCH_SIZE available in the screen to fine-tune the training, this is now possible.
  -  Config variable that are overwritten can now be saved back into the myconfig.py file to be permanent
  -  In the trainer screen the database columns can be selected/deselected for a cleaner view
  -  The pilot arena screen allows to load even more pilots and allows to configure the panel view by selecting the number of columns. Also very useful if only 1 pilot is chosen. 
  -  The pilot arena screen now also offers to run Tub plot, and the command `donkey tubplot` has been removed as it was not working anyway.
  - The pilot arena colors the user and model data matching with the lines in the images
  -  Graphs in the UI are now produced by matplotlib
  
4. The UI will now by default be installed also on the RPi. The donkey installation requires a `sudo apt install libhdf5-dev` before.    

## Other changes
We have introduced a more versatile image caching policy to support a binary caching format in addition to the no cache and the numpy array which we have now. The binary format stores the jpeg compression which gives a good compromise between performance and memory footprint. Training with 100k records and binary caching using 32GB or ram w/o swapping is possible.  

## To install this code
I pushed this code to test-pypi and it can be installed with `pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple donkeycar-pypi\[XX\]==5.2.dev2`, where `XX` is either `pc` or `pi` or `nano`. No escaping of `[` or `]` needed if you are on bash.

# Screenshots

## Tub manager
image

## Trainer
image

## Pilot arena
image

## Car connector
image

PR #1179, opened 2024-05-21
Fix augmentations and matplotlib

DocGarbanzo:
# Fix recent compatibility issues with augmentations and matplotlib
* There seems to be a non-backward compatible change in matplotlib 3.9.* which we can resolve for the time being, by pegging it to version 3.8.* (https://discord.com/channels/662098530411741184/671604367706554388/1242225275941425193)
* There also seems to be a non-back compatible change in albumentations for empty compositions, for which we have a code fix here
* There is another backward incompatible change in `pytest`, hence pinning this to version 8.1 (https://github.com/mtreinish/stestr/issues/363)
* Also updated layout height for certain widgets in the UI under OSX which are too compressed
* Fixed tensor flow package name under OSX
* Fixed an older `__main__.__file__` expression in `load_config` which I found to be not working

### DIFF
--- a/donkeycar/pipeline/training.py
+++ b/donkeycar/pipeline/training.py
@@ -153,7 +153,7 @@ def train(cfg: Config, tub_paths: str, model: str = None,
     assert val_size > 0, "Not enough validation data, decrease the batch " \
                          "size or add more data."
     logger.info(f'Train with image caching: '
-                f'{getattr(cfg, "CACHE_IMAGES", True)}')
+                f'{getattr(cfg, "CACHE_IMAGES", "ARRAY")}')
     history = kl.train(model_path=model_path,
                        train_data=dataset_train,
                        train_steps=train_size,
--- a/donkeycar/config.py
+++ b/donkeycar/config.py
@@ -66,8 +66,7 @@ class Config:
 def load_config(config_path=None, myconfig="myconfig.py"):
     
     if config_path is None:
-        import __main__ as main
-        main_path = os.path.dirname(os.path.realpath(main.__file__))
+        main_path = os.getcwd()
         config_path = os.path.join(main_path, 'config.py')
         if not os.path.exists(config_path):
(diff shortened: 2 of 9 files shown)


donkeycar/pipeline/training.py
```diff
@@ -153,7 +153,7 @@ def train(cfg: Config, tub_paths: str, model: str = None,
     assert val_size > 0, "Not enough validation data, decrease the batch " \
                          "size or add more data."
     logger.info(f'Train with image caching: '
-                f'{getattr(cfg, "CACHE_IMAGES", True)}')
+                f'{getattr(cfg, "CACHE_IMAGES", "ARRAY")}')
```
Ezward, line 156: @DocGarbanzo I don't know that this means.  Does this limit the size of the cache?

Ezward (commented):
I have one question; see the comment.

donkeycar/pipeline/training.py
DocGarbanzo: In the previous PR, we have upgraded the CACHE_IMAGES config parameter to allow for 3 values, i.e NOCACHE, ARRAY and BINARY, please see here: https://github.com/autorope/donkeycar/pull/1173/files#diff-8a130617f883f549bf089e3ab0c46cabd4c3e4a69b50568c3b27ef73414e7c28. ARRAY represents the original setting where the whole image is cached as a `np.uint8` array (consuming about 58k of mem ~ 120 * 160 * 3). With BINARY we can now cache the image as binary jpeg encoded which uses only 4k of memory. This caching still provides significantly faster access to the image than re-reading it from disk, even with a fast M2 SSD. With that much smaller memory footprint we can now train >> 100k images with caching, speeding up the training on large datasets by a lot. In this line here, I just fixed a left-over from the previous PR and here, it only affects logging the caching status into the shell, because the default changed from the binary `True` to the enum/string `'ARRAY'`.

Ezward (approved):
Thanks for this.

PR #1186, opened 2024-06-11
Add support from Hokuyo lidar

mgagvani:
- Part in `lidar.py` to interface with Hokuyo
- Fixes to `LidarPlot2` to be used with Hokuyo so it can act as a `Camera` in the web interface
- Integration of `LidarPlot2` with existing templates/configs

**Notes:** 
To set it up you need to configure the Ethernet port on the Pi by creating a file in `/etc/interfaces.d`. This is well-described by many online tutorials, and googling "f1tenth hokuyo" also brings up a lot of helpful info. 

Also, unlike the RPLidar(2) parts, instead of returning a generator of individual measurements this part returns the entire scan at once. Thus in `complete.py` it is called `lidar/dist_scan` instead of `lidar/dist_array`. 

**Pictures:** 
image
IMG_8345
IMG_8347

### DIFF
--- a/donkeycar/parts/lidar.py
+++ b/donkeycar/parts/lidar.py
@@ -402,17 +402,17 @@ class HokuyoLidar(object):
         self.DMAX = max_dist
 
     def poll(self):
-        # scan is list of (theta, r) - e.g. shape of 1000, 2
         timestamp, scan = self.laser.get_filtered_dist(dmax=self.DMAX)
 
-        # flip so it's (r, theta) --> compatible with LidarPlot
-        distances = scan[,:1]
-        angles = scan[,:0]
-        self.scan = np.hstack(distances, angles)
+        # shape (n, 2) --> list of (theta, r, _, _, _)
+        angles, distances = scan[:,0], scan[:,1]
+        angles = np.rad2deg(angles)
+        filler = np.zeros_like(angles) # for LidarPlot
+        self.scan = np.stack((distances, angles, filler, filler, filler), axis=-1)
 
     def update(self):
         self.poll()
-        time.sleep(0) # copied from RPLidar2
+        time.sleep(0) # copied from RPLidar2 (release to other threads)
 
     def run_threaded(self):
         return self.scan
@@ -736,6 +736,11 @@ class LidarPlot2(object):
                           self.angle_direction, self.rotate_plot)
         plot_polar_angle(draw, bounds, self.border_color, 0,
                          self.angle_direction, self.rotate_plot)
+        
+        # handle no data yet
+        if measurements is None:
+            print("bruh")
+            return np.asarray(self.frame)
         
         # data points
         plot_polar_points(
--- a/donkeycar/parts/lidar.py
+++ b/donkeycar/parts/lidar.py
@@ -389,6 +389,41 @@ class YDLidar(object):
         self.lidar.StopScanning()
         self.lidar.Disconnect()
 
+class HokuyoLidar(object):
+    '''
+    Class for ethernet-based Hokuyo lidars (e.g. UST-10LX in F1Tenth)
+    '''
+    def __init__(self, max_dist):
+        '''
+        max_dist: maximum distance in mm (e.g. 20,000 for UST-20LX)
+        '''
+        from hokuyolx import HokuyoLX
+        self.laser = HokuyoLX()
+        self.DMAX = max_dist
+
+    def poll(self):
+        # scan is list of (theta, r) - e.g. shape of 1000, 2
+        timestamp, scan = self.laser.get_filtered_dist(dmax=self.DMAX)
+
+        # flip so it's (r, theta) --> compatible with LidarPlot
+        distances = scan[,:1]
+        angles = scan[,:0]
+        self.scan = np.hstack(distances, angles)
+
+    def update(self):
+        self.poll()
+        time.sleep(0) # copied from RPLidar2
+
+    def run_threaded(self):
+        return self.scan
+
+    def run(self):
+        self.poll()
+        return self.scan
+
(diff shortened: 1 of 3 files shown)


donkeycar/parts/lidar.py
```diff
@@ -737,6 +737,11 @@ def run(self, measurements):
         plot_polar_angle(draw, bounds, self.border_color, 0,
                          self.angle_direction, self.rotate_plot)
         
+        # handle no data yet
+        if measurements is None:
+            print("bruh")
```
Ezward, line 742: Do you want to add a debug log statement here or delete this?

donkeycar/templates/complete.py
```diff
@@ -880,7 +880,31 @@ def add_camera(V, cfg, camera_type):
               outputs=['cam/image_array', 'cam/depth_array',
                        'imu/acl_x', 'imu/acl_y', 'imu/acl_z',
                        'imu/gyr_x', 'imu/gyr_y', 'imu/gyr_z'],
+              threaded=False)
+        
+    elif cfg.CAMERA_TYPE == "OAKD":
```
Ezward, line 885: I don't thing the oakd part is in main yet; so should not be in this PR.  I'll go look at the oakd tomorrow, but it should not be mixed in this PR.

donkeycar/templates/complete.py
```diff
@@ -880,7 +880,31 @@ def add_camera(V, cfg, camera_type):
...
+        from donkeycar.parts.oak_d import OakD
+        cam = OakD(
+            enable_rgb=cfg.OAKD_RGB,
+            enable_depth=cfg.OAKD_DEPTH,
+            device_id=cfg.OAKD_ID)
+        V.add(cam, inputs=[],
+              outputs=['cam/image_array', 'cam/depth_array'],
               threaded=True)
+    
+    elif cfg.CAMERA_TYPE == "LIDAR_PLOT":
```
Ezward, line 895: This is very interesting.  So the idea here is that the lidar plot would be used as the input to the CNN?  Separate from the code concerns, have you tried this?  Does it produce a reasonable autopilot?

donkeycar/parts/lidar.py
```diff
@@ -737,6 +737,11 @@ def run(self, measurements):
         plot_polar_angle(draw, bounds, self.border_color, 0,
                          self.angle_direction, self.rotate_plot)
         
+        # handle no data yet
+        if measurements is None:
```
Ezward, line 741: maybe this can be 
```python
# data points
if measurements:
  plot_polar_points(
      draw, bounds, self.mark_fn, self.point_color, self.mark_px,
      [(distance, angle) for distance, angle, _, _, _ in measurements],
      self.max_distance, self.angle_direction, self.rotate_plot)

```

donkeycar/templates/cfg_complete.py
```diff
@@ -349,9 +349,11 @@
 
 # #LIDAR
 USE_LIDAR = False
-LIDAR_TYPE = 'RP' #(RP|YD)
+LIDAR_TYPE = 'RP' #(RP|YD|HOKUYO)
 LIDAR_LOWER_LIMIT = 90 # angles that will be recorded. Use this to block out obstructed areas on your car, or looking backwards. Note that for the RP A1M8 Lidar, "0" is in the direction of the motor
 LIDAR_UPPER_LIMIT = 270
+LIDAR_MAX_DIST = 10_000 # Maximum distance for LiDAR. Measured in mm
```
Ezward, line 355: These values need to have defaults based on the chose lidar type.  RP/YD would be 0 for angle and I'm not sure for distance.

Ezward (changes requested):
Thank you for this PR.  It would be great to be able to use the Lidar on an F1Tenth car.  The the code looks very good; I have left a few comments.  There are other concerns:  
- is there a parallel PR for the docs?  Without documentation on how to setup the Lidar the software is not that useful and it becomes a maintenance burden.  Links out to reliable, stable content can do most of the work.  It would also be great to have a video of the webui output.  
- we should also mention in the camera section of the docs about how to setup the 'camera' to be the lidar scan.  That is a really good idea.
- we should rebase/squash these two commits into a single commit.

mgagvani:
Thanks @Ezward for the review, I'll probably get to the fixes/writing docs in a week or so.

donkeycar/templates/complete.py
mgagvani: I was thinking more along the lines of being able to see what the lidar sees. I haven't tried to train an autopilot with it although that does sound like an interesting idea.

donkeycar/templates/cfg_complete.py
mgagvani: Do you mean these should be set to zero for the user to put in themselves? I wanted to give this example to show that the back half of scans should likely be discarded in most set-ups. As for the max distance, I think this is reasonable for the entry level RP Lidar/ YD Lidar. If someone buys a fancier model then I would assume they would also know how to change this value to reflect the capabilities of their lidar.

mgagvani:
Hi @Ezward I addresssed your review and made a PR for documentation: https://github.com/autorope/donkeydocs/pull/56

donkeycar/templates/complete.py
```diff
@@ -872,7 +880,22 @@ def add_camera(V, cfg, camera_type):
               outputs=['cam/image_array', 'cam/depth_array',
                        'imu/acl_x', 'imu/acl_y', 'imu/acl_z',
                        'imu/gyr_x', 'imu/gyr_y', 'imu/gyr_z'],
+              threaded=False)
```
DocGarbanzo, line 883: This should be left a threaded part.

PR #1189, opened 2024-06-21
Update github README.md with more detail and links to docs

Ezward:
Update and reorganize the readme with a lot more explanation of Donkeycar and links to key parts of the documentation.  It also includes some helpful links out to resources that will help a new user be prepared for Donkeycar.


zlite (approved):
Nice work!

PR #1193, opened 2024-07-18
Add Support for Video Games as a Medium of Control/Recording

mgagvani:
This PR introduces two parts:

1. `ScreenCamera` is a part in `camera.py` which takes screenshots and uses those as camera images.
2. `VirtualController` in `actuator.py` is a part for an emulated Xbox controller which can be used to control a car in videogames.

I used these two parts to make a (mostly) successful pilot in Forza Horizon 3: https://youtu.be/LIuH6aonYWU

### DIFF
--- a/donkeycar/parts/camera.py
+++ b/donkeycar/parts/camera.py
@@ -1,8 +1,10 @@
 import logging
 import os
 import time
+import threading
 import numpy as np
 from PIL import Image
+import mss
 import glob
 from donkeycar.utils import rgb2gray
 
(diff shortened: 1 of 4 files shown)


donkeycar/parts/camera.py
```diff
@@ -1,8 +1,10 @@
 import logging
 import os
 import time
+import threading
 import numpy as np
 from PIL import Image
+import mss
```
DocGarbanzo, line 7: Please have a look at the failed tests, you need to add the python packages into setup.cfg in the right place. Likely we only want to support that in the PC installation but not on the robot.

donkeycar/parts/camera.py
```diff
@@ -329,6 +331,61 @@ def shutdown(self):
...
+                 vflip=False, hflip=False):
+        self.image_w = image_w
+        self.image_h = image_h
+        self.image_d = image_d
+        self.vflip = vflip
+        self.hflip = hflip
+        self.sct = mss.mss()
+        self.running = True
+
+        self._monitor_thread = threading.Thread(target=self.take_screenshot, args=())
```
DocGarbanzo, line 350: Please have a look at other threaded parts. Threads are not started in the constructor but centrally in the car application, when adding the part as a threaded part - which is what you want to do for your part.

donkeycar/parts/camera.py
```diff
@@ -329,6 +331,61 @@ def shutdown(self):
...
+        img = img.resize((self.image_w, self.image_h))
+        img_arr = np.asarray(img)
+        if self.vflip:
+            img_arr = np.flipud(img_arr)
+        if self.hflip:
+            img_arr = np.fliplr(img_arr)
+        self.frame = img_arr
+        return img
+
+    def update(self):
```
DocGarbanzo, line 372: The method should not return but refresh the class member `self.frame`.

donkeycar/parts/camera.py
```diff
@@ -329,6 +331,61 @@ def shutdown(self):
...
+        if self.running:
+            img = self.take_screenshot()
+            return img
+
+    def run(self):
+        self.update()
+        assert self.frame is not None
+        return self.frame
+
+    def run_threaded(self):
```
DocGarbanzo, line 382: No need to overwrite.

DocGarbanzo (changes requested):
Please see my comments in the files.

PR #1198, opened 2024-09-14
Fix  bug in path.py

zlite:
Should return three values, but in the case of None it only returned two, causing an error. Thanks to Dirk for debugging


DocGarbanzo (approved):
Great, that's what we checked during the call.

PR #1217, opened 2026-01-02
Updated to handle Pi Zero properly

zlite:
Because the Pi Zero doesn't have an Ethernet port, the call to eth0 fails. This fixes it


DocGarbanzo (approved):
@zlite  - I don't understand why we are not seeing a proper diff as I believe most of the lines have not changed. But we are seeing a big block removed and added instead.

PR #1219, opened 2026-01-04
Switches Pi-side NN framework to TFLite (AKA LiteRT) for smaller install and faster performance

zlite:
On the Pi, we only do inference so TFLite is sufficient. It also allows installation on the Pi Zero 2, which chokes on a full Tensorflow installation. 

This PR keeps full Tensorflow on the PC, but switches to TFLite on the Pi.

It should also be faster at inference than full Tensorflow since it's designed for edge devices like the Pi (although I have not benchmarked that yet)

### DIFF
--- a/donkeycar/parts/interpreter.py
+++ b/donkeycar/parts/interpreter.py
@@ -1,79 +1,42 @@
-from __future__ import annotations
-
 import os
 from abc import ABC, abstractmethod
 import logging
 import numpy as np
-from typing import Union, Sequence, List, Optional
+from typing import Union, Sequence, List
 
-tf = None
-keras = None
-tag_constants = None
-signature_constants = None
-trt = None
+try:
+    import tensorflow as tf
+    from tensorflow import keras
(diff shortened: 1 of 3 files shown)


donkeycar/parts/litert.py
DocGarbanzo: I believe, this file should be deleted, no need to create a parallel universe of the model interfaces.

donkeycar/parts/interpreter.py
```diff
@@ -262,8 +327,14 @@ def load(self, model_path):
         assert os.path.splitext(model_path)[1] == '.tflite', \
             'TFlitePilot should load only .tflite files'
         logger.info(f'Loading model {model_path}')
+        interpreter_cls = _get_tflite_interpreter()
```
DocGarbanzo: Basically here is the place where we need to decide between loading the runtime-only interpreter alternatively.

donkeycar/utils.py
DocGarbanzo: There should be no change required here.

PR #1223, opened 2026-02-07
Modified the web interface so the assigned buttons have labels

zlite:
Screenshot 2026-02-06 195800

It always bugged me that the buttons just said (1, 2, 3...) and you had to remember what you assigned to each. Now the web interface displays them

### DIFF
--- a/donkeycar/__init__.py
+++ b/donkeycar/__init__.py
@@ -3,7 +3,7 @@ import sys
 from pyfiglet import Figlet
 import logging
 
-__version__ = '5.2.dev5'
+__version__ = '5.2.dev6'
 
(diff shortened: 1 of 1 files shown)


DocGarbanzo:
@zlite - there are no changed files in the PR. Can you please check?

DocGarbanzo:
@zlite - I merged your other PR and now this has a conflict. Can you pull from main and resolve that? I think it needs some care because we don't want to lose your documentation of the config parameter.

donkeycar/parts/interpreter.py
Ezward, line 7: The PR is about adding the labels to the web interface buttons (awesome, thank you), but this commit is regarding TFLite support.  This should be in a separate PR.  I would just delete this commit and recreate it is a separate PR.

PR #1230, opened 2026-03-27
Fix numpy/opencv dependencies and enable CI training tests

DocGarbanzo:
- Pin numpy>=1.23,<2.0 to prevent incompatibility with tensorflow 2.15 (numpy 2.x breaks tensorflow 2.15 at runtime)
- Add opencv-contrib-python-headless==4.9.* to pc extras so opencv is available in CI (headless for server/CI environments)
- Pin opencv-contrib-python==4.9.* in pi extras for consistency
- Add pytest-rerunfailures to dev extras (pytest.ini already uses reruns=3)
- Remove GITHUB_ACTIONS skip from test_train.py keras training tests; these tests work fine with tensorflow installed and should run in CI
- Replace GITHUB_ACTIONS skip in test_torch.py with torch availability check so tests skip gracefully when torch is not installed rather than hard-coding the CI environment
- Handle fastai_linear test case in test_train.py with importorskip so it skips cleanly when torch/fastai are not installed

All 164 tests pass locally (16 skipped due to torch not being installed).

https://claude.ai/code/session_01EHRf479YqUALrA4t2YAzC3

### DIFF
--- a/donkeycar/__init__.py
+++ b/donkeycar/__init__.py
@@ -3,7 +3,7 @@ import sys
 from pyfiglet import Figlet
 import logging
 
-__version__ = '5.2.dev6'
+__version__ = '5.2.dev7'
 
 logging.basicConfig(level=os.environ.get('LOGLEVEL', 'INFO').upper())
 
--- a/donkeycar/templates/complete.py
+++ b/donkeycar/templates/complete.py
@@ -1110,7 +1110,7 @@ def add_drivetrain(V, cfg):
         elif cfg.DRIVE_TRAIN_TYPE == "MM1":
             from donkeycar.parts.robohat import RoboHATDriver
             # Share serial port with controller to avoid opening the same port twice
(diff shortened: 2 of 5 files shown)


DocGarbanzo:
@zlite - You mentioned tflite-runtime to be faster than tensorflow at inferencing on the pi. I couldn't originally reproduce this, but I have re-done the exercise and I'm seeing a massive improvement in performance, producing around 265 fps vs 165 fps.

zlite:
Such good news... and that is in addition to the faster loading time.
Thanks for testing that!

zlite:
That all works great! But I seem to have created a small bug in my MM1 drivetrain call in the earlier PR that throws an error when trying to open the serial port (it's only an issue if you're using the RC hat). 

Can you please change this line in complete.py :
V.add(RoboHATDriver(cfg, serial_port=ctr.serial), inputs=['steering', 'throttle'])

to this:
V.add(RoboHATDriver(cfg), inputs=['steering', 'throttle'])

I tested that and it fixes it.

zlite (approved):
LGTM! Many thanks

PR #1234, opened 2026-04-20
Replace import gym with gymnasium and update step() return handling

lavar-rnf:
- Switched import from 'import gym' to 'import gymnasium as gym'
- Updated env.step() handling to support 5-tuple return (obs, reward, terminated, truncated, info)
- Fixes compatibility with newer gymnasium-based Donkey Simulator

### DIFF
--- a/donkeycar/parts/dgym.py
+++ b/donkeycar/parts/dgym.py
@@ -1,6 +1,9 @@
 import os
 import time
-import gymnasium as gym
+try:
+    import gymnasium as gym
+except ImportError:  # pragma: no cover
+    import gym
 import gym_donkeycar
 
 
--- a/donkeycar/parts/dgym.py
+++ b/donkeycar/parts/dgym.py
@@ -1,6 +1,6 @@
 import os
 import time
-import gym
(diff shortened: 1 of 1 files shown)


DocGarbanzo:
@lavar-rnf - thanks for this PR. I think you are right, we have to upgrade from `gym` to `gymnasium`, but it probably needs to happen in a couple more places. We try to get @Maximellerbach, to help with that.

lavar-rnf:
Sure! I have no problem collaborating on that if you think I can help.
Copilot's suggestion for putting "import gymnasium as gym" in a try
statement was spot on! We should also do that for the calls to
self.env.step(self.action), since they return a 5-tuple in Gymnasium rather
than a 4-tuple in Gym. Gymnasium isn't exactly a drop-in replacement for
Gym.

<http://readynowfoundation.org>

Lavar Askew

Founder, Executive Director

***@***.***

619-777-6438

Calendly <https://calendly.com/lavar-readynowfoundation> | LinkedIn
<https://www.linkedin.com/in/lavaraskew/>

PR #1235, opened 2026-04-25
Fix OAK-D RGB output not respecting IMAGE_W/IMAGE_H by using preview stream

k8yuan:
The OAK-D RGB camera output did not respect IMAGE_W and IMAGE_H settings when using the default video stream. This caused mismatches between configured resolution, web UI display, and model input expectations.

This change switches the pipeline from the video stream to the preview stream, which correctly applies setPreviewSize() in hardware. As a result, the camera output now matches the configured dimensions, improves consistency between training and inference, and reduces unnecessary resizing overhead.

I verified this behavior by comparing images saved to disk and transferred to a local machine with the live preview shown in the web UI. After this change, the saved images match the preview output and respect the configured resolution.

### DIFF
--- a/donkeycar/parts/oak_d.py
+++ b/donkeycar/parts/oak_d.py
@@ -138,7 +138,9 @@ class OakD(object):
         res = depthai.ColorCameraProperties.SensorResolution.THE_1080_P
 
         cam_rgb.setResolution(res)
-        cam_rgb.setVideoSize(width, height)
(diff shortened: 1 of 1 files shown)


DocGarbanzo (approved):

DocGarbanzo:
@k8yuan - thank you for addressing this issue.

Issue #1236, opened 2026-05-27
Adding support for new GPS and IMU fusion


aokhader:
We are working on adding support for the SparkFun NEO-F10N GPS and Adafruit BNO08x IMU to offer an alternative to the Point One Nav GNSS module

TCIII:
@aokhader,

It might be smart to allow more than just the NEO-F10N GPS as many DC users have RTK GPSs and the NEO-F10N GPS does not have RTK capability.
Regards,
TCIII

aokhader:
Sounds good. We are looking to make the IMU a standalone part that can integrate with the other GPS. We're currently testing on the NEON-F10N GPS, but after that we'll look into the Point One Nav module after that.

PR #1237, opened 2026-06-12
Add a GPS + IMU Fusion Part

Phill310:
This is the addition mentioned in #1236.
- adds the BNO08x IMU into IMU.py
- creates a fusion part in gps_imu_fusion.py
This fusion part uses a 5 state Kalman filter to predict your position based on your velocity and acceleration data. The 5th variable is an acceleration bias to prevent drift when the car isn't moving. There is also a low pass EMA filter to clean out noise. Finally we added a friction factor to slowly bleed velocity.

We tested with a SparkFun NEO-F10N GPS which had a frequency of 10Hz. I believe the BNO08x IMU has a frequency of 50 Hz so in theory the fusion part can go up to 50 Hz. The main limiting factor is the drive loop frequency since the fusion part recalculates position once every drive loop.

This should support any GPS since the fusion part only takes the x and y position data from the gps.

Config Options added:
- USE_FUSION
- FUSION_DEBUG

I hope this explains everything we did. If not please ask questions so I can expand on our thinking.
Should I also make a pr in the docs repo to add a section for the fusion part and update the IMU section?


TCIII:
@Phill310,

Are the x and y GPS positions in degrees or in meters from the Greenwich Meridian and the Equator?

Regards,
TCIII

Phill310:
> @Phill310,
> 
> Are the x and y GPS positions in degrees or in meters from the Greenwich Meridian and the Equator?
> 
> Regards, TCIII

I believe that the Fusion part outputs the position in UTM coordinates (same as the gps part as far as I know). So yes it give the distance from the equator in meters but for easting it outputs the number of meters from its specific zone's central meridian. Is this something we should describe in the fusion class?

donkeycar/parts/gps_imu_fusion.py
```diff
@@ -0,0 +1,156 @@
...
+        self.last_time = time.time()
+        self.last_gps = (0.0, 0.0)
+        self.initialized = False
+
+    def _quat_to_yaw(self, qi, qj, qk, qr):
+        siny_cosp = 2 * (qr * qk + qi * qj)
+        cosy_cosp = 1 - 2 * (qj * qj + qk * qk)
+        return math.atan2(siny_cosp, cosy_cosp)
+
+    def run(self, gps_x, gps_y, ax, ay, az, gx, gy, gz, qi, qj, qk, qr):
```
DocGarbanzo: Can you change the interface to pass vectors instead of single coordinates:
[gps_x, gpy_y] -> gps
[ax, ay, az] -> a or accel?
[gx, gy, gz] ->  g or gyro?
[qi, qj, qk, qr] -> q or quarternion?

donkeycar/parts/imu.py
```diff
@@ -20,21 +22,21 @@ class IMU:
     sudo python setup.py install
 
     pip install mpu6050-raspberrypi
-    
+
     - MPU9250
     pip install mpu9250-jmdev
```
DocGarbanzo, line 27: These install commands are partially wrong (`sudo...`). can you add the python dependencies into `setup.cfg` so they get installed as part of the standard donkey install?

donkeycar/parts/imu.py
```diff
@@ -48,15 +50,14 @@ def __init__(self, addr=0x68, poll_delay=0.0166, sensor=SENSOR_MPU6050, dlp_sett
...
                 self.sensor.writeSlave(CONFIG_REGISTER, dlp_setting)
             self.sensor.calibrateMPU6500()
             self.sensor.configure()
 
-        
-        self.accel = { 'x' : 0., 'y' : 0., 'z' : 0. }
-        self.gyro = { 'x' : 0., 'y' : 0., 'z' : 0. }
+        self.accel = {'x': 0., 'y': 0., 'z': 0.}
+        self.gyro = {'x': 0., 'y': 0., 'z': 0.}
         self.mag = {'x': 0., 'y': 0., 'z': 0.}
```
DocGarbanzo: Dictionaries are unnecessary here, please just use vectors (or lists, if you prefer python native over numpy. Both types would be fine).

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
                 self.accel, self.gyro, self.temp = self.sensor.get_all_data()
             else:
                 from mpu9250_jmdev.registers import GRAVITY
                 ret = self.sensor.getAllData()
-                self.accel = { 'x' : ret[1] * GRAVITY, 'y' : ret[2] * GRAVITY, 'z' : ret[3] * GRAVITY }
-                self.gyro = { 'x' : ret[4], 'y' : ret[5], 'z' : ret[6] }
-                self.mag = { 'x' : ret[13], 'y' : ret[14], 'z' : ret[15] }
+                self.accel = {'x': ret[1] * GRAVITY, 'y': ret[2] * GRAVITY, 'z': ret[3] * GRAVITY}
+                self.gyro = {'x': ret[4], 'y': ret[5], 'z': ret[6]}
+                self.mag = {'x': ret[13], 'y': ret[14], 'z': ret[15]}
```
DocGarbanzo: Same here, vectors are better than dicts.

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
                 self.temp = ret[16]
         except:
             print('failed to read imu!!')
-            
+
+    def run_threaded(self):
+        return self.accel['x'], self.accel['y'], self.accel['z'], self.gyro['x'], self.gyro['y'], self.gyro[
```
DocGarbanzo: It would also be much cleaner to return a tuple of two vectors (accel and gyro) and a float (temp), instead of unravelling the vectors.

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
+            'z'], self.temp
+
+    def run(self):
+        self.poll()
+        return self.accel['x'], self.accel['y'], self.accel['z'], self.gyro['x'], self.gyro['y'], self.gyro[
```
DocGarbanzo: Same observation here.

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
+        # Initialize I2C bus
+        self.i2c = busio.I2C(board.SCL, board.SDA)
+        self.sensor = BNO08X_I2C(self.i2c, address=addr)
+
+        # Explicitly enable the streams needed for navigation and fusion
+        self.sensor.enable_feature(BNO_REPORT_ACCELEROMETER)
+        self.sensor.enable_feature(BNO_REPORT_GYROSCOPE)
+        self.sensor.enable_feature(BNO_REPORT_ROTATION_VECTOR)
+
+        self.accel = {'x': 0., 'y': 0., 'z': 0.}
```
DocGarbanzo: Pls see comment above.

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
+
+    def update(self):
+        import time
+        while self.on:
+            self.poll()
+            time.sleep(self.poll_delay)
+
+    def poll(self):
+        try:
+            ax, ay, az = self.sensor.acceleration
```
DocGarbanzo, line 136: You see how all this stuff simplifies dramatically if you work with vectors directly.

donkeycar/parts/imu.py
```diff
@@ -66,27 +67,89 @@ def update(self):
...
+            qi, qj, qk, qr = self.sensor.quaternion
+            self.quat['i'], self.quat['j'], self.quat['k'], self.quat['real'] = qi, qj, qk, qr
+        except Exception as e:
+            print(f"Failed to read BNO08x: {e}")
+
     def run_threaded(self):
-        return self.accel['x'], self.accel['y'], self.accel['z'], self.gyro['x'], self.gyro['y'], self.gyro['z'], self.temp
+        return (self.accel['x'], self.accel['y'], self.accel['z'],
+                self.gyro['x'], self.gyro['y'], self.gyro['z'],
+                self.quat['i'], self.quat['j'], self.quat['k'], self.quat['real'])
```
DocGarbanzo: Here as well, return a tuple of (accel, gyro, quarternion).

donkeycar/templates/complete.py
```diff
@@ -505,25 +501,25 @@ def run(self, *components):
 
     if cfg.HAVE_IMU or (cfg.CAMERA_TYPE == "D435" and cfg.REALSENSE_D435_IMU):
         inputs += ['imu/acl_x', 'imu/acl_y', 'imu/acl_z',
```
DocGarbanzo: this simplifies with vectors

donkeycar/templates/complete.py
```diff
@@ -505,25 +501,25 @@ def run(self, *components):
...
-           'float', 'float', 'float']
+        types += ['float', 'float', 'float',
+                  'float', 'float', 'float']
 
     # rbx
     if cfg.DONKEY_GYM:
         if cfg.SIM_RECORD_LOCATION:
             inputs += ['pos/pos_x', 'pos/pos_y', 'pos/pos_z', 'pos/speed', 'pos/cte']
-            types  += ['float', 'float', 'float', 'float', 'float']
+            types += ['float', 'float', 'float', 'float', 'float']
```
DocGarbanzo, line 511: And this...

donkeycar/templates/complete.py
```diff
@@ -505,25 +501,25 @@ def run(self, *components):
...
         if cfg.SIM_RECORD_GYROACCEL:
             inputs += ['gyro/gyro_x', 'gyro/gyro_y', 'gyro/gyro_z', 'accel/accel_x', 'accel/accel_y', 'accel/accel_z']
```
DocGarbanzo, line 513: And this...

donkeycar/templates/complete.py
```diff
@@ -505,25 +501,25 @@ def run(self, *components):
...
-            types  += ['float', 'float', 'float', 'float', 'float', 'float']
+            types += ['float', 'float', 'float', 'float', 'float', 'float']
         if cfg.SIM_RECORD_VELOCITY:
             inputs += ['vel/vel_x', 'vel/vel_y', 'vel/vel_z']
```
DocGarbanzo, line 516: ... and this as well.

donkeycar/templates/complete.py
```diff
@@ -877,7 +873,7 @@ def add_camera(V, cfg, camera_type):
                        'imu/acl_x', 'imu/acl_y', 'imu/acl_z',
```
DocGarbanzo: If you don't want to touch this class, but want to work with vectors, then you can write a one-liner class that combines x, y, z values into a vector, like:

```python 
class Vectoriser:
    def run(x, y, z):
        return [x, y, z]
```

donkeycar/templates/path_follow.py
```diff
@@ -464,20 +464,41 @@ def add_gps(V, cfg):
...
+
+        if cfg.USE_FUSION:
+            if cfg.HAVE_IMU:
+                from donkeycar.parts.gps_imu_fusion import EKFFusion
+                fusion = EKFFusion(debug=cfg.FUSION_DEBUG)
+
+                V.add(fusion,
+                      inputs=[
+                          'gps/utm/longitude', 'gps/utm/latitude',  # From GPS
+                          'imu/acl_x', 'imu/acl_y', 'imu/acl_z',  # From IMU
```
DocGarbanzo: Also, use vectors here.

DocGarbanzo (changes requested):
Thank you @Phill310 - this is good PR that we would like to merge. I have two general feedback points:
1) We should not have to call `pip install some_package` to make use of the BNO085 or other sensors. The relevant python packages should be added to `setup.cfg` so everything works out of the box.
2) The IMU interfaces use scalars instead of vectors for acceleration, gyro, quaternions, etc. That is bad, but it is of course not your fault as such code was there before. It would be great if instead of making this even worse, we would clean this up as part of this PR as for the moment, no-one is really using IMU data yet. - Let me know if this exceeds your intended contribution and we can do a refactoring PR afterwards.

aokhader:
We are working on refactoring our code to use vectors instead of scalars, and we added the package install to ``setup.cfg``. For the refactoring of the IMU code in general and for the code that wasn't directly from our fusion module, we left it as-is since we thought it was the base setup for the code. We are going to try to refactor the the code for all the IMU modules, but I'm not sure if we can verify if everything works the way it should for all the components including the camera since it would be out of our scope. Right now it is a PR in our repo if it helps.

aokhader:
I've refactored the IMU code to use vectors and changed all the dependencies for the IMU to do the same, as well as verified that it works properly on path following. Is there anything else that needs fixing in this code?
Just to reiterate, the IMU fusion module is GPS-agnostic, so it can be used with a GPS with RTK corrections.

DocGarbanzo (approved):
Looks good to me.

DocGarbanzo:
Thanks, @aokhader - merged your PR.

PR #1238, opened 2026-07-09
Fixed OAK-D FOV cropping and single-stream RuntimeError from previous PR

YashTandon05:
## Summary
Fixes OAK-D RGB camera FOV cropping and a `RuntimeError` when only one stream (RGB or depth) is enabled. Also fixes the crash introduced in #1235.

## Problem
- The RGB stream was linked from `cam_rgb.video`, which is hard-cropped to 16:9 regardless of `setResolution()`, clipping the sensor's native ~4:3 FOV (near-ground view was cut off).
- `#1235` attempted to fix this by switching to `setPreviewSize()`, but referenced `self.image_w`/`self.image_h`, which are never defined on `OakD` — causing an `AttributeError` on every camera init. It also left `xout_rgb` linked to `.video` instead of `.preview`, so even patched, the resolution fix would have had no effect.
- Separately, `_poll()` unconditionally queried both the `"rgb"` and `"depth"` output queues whenever either stream was enabled (`if self.enable_rgb or self.enable_depth`). If only one stream was enabled in config, the other's `XLinkOut` was never wired into the pipeline, and querying its queue raised a `RuntimeError`.

## Fix
- `setup_rgb_camera()`: use `THE_13_MP` resolution + `setIspScale(1, 6)`, linking `cam_rgb.isp` instead of `.video`. `isp` preserves the sensor's full FOV; `_poll()` resizes the (approximately-scaled) output down to the exact configured `IMAGE_W`/`IMAGE_H` with `cv2.resize`.
- `_poll()`: initialize the `rgb_queue`/`depth_queue` independently based on `self.enable_rgb`/`self.enable_depth` rather than on their `or`, so a single-stream config no longer touches the disabled stream's (nonexistent) queue.

## Testing
Ran live on a physical rig (OAK-D Lite, `CAMERA_TYPE=OAKD`, `OAKD_RGB=True`, `OAKD_DEPTH=False`) via `manage.py drive`. Camera initializes without error, full FOV is restored (previously only saw ~4ft+ ahead), and output resolution matches configured `IMAGE_W`/`IMAGE_H`.


YashTandon05:
@claude review

DocGarbanzo (approved):

PR #1250, opened 2026-09-14
Run donkeycar on the Arduino UNO Q, with the MCU driving the servo and ESC

Ezward:
Runs donkeycar on the **Arduino UNO Q**, with the board's own STM32U585 MCU
driving the steering servo and ESC.

## Why this shape

The UNO Q's Arduino-shaped header belongs to the STM32, **not** to the
Qualcomm SoC running Linux. A PCA9685 wired to the header SDA/SCL is invisible
to `/dev/i2c-*`, and `adafruit-platformdetect` does not recognise the board at
all, so `import board` raises and no CircuitPython driver will load. The MCU
therefore owns the real-time work — reading an RC receiver and emitting the
servo pulses — and Linux does vision and inference. This is the same split as
the Robo HAT MM1 and the DIY Robocars RC hat.

`analogWrite()` cannot produce servo pulses on this board: it calls
`pwm_set_pulse_dt()` and the devicetree fixes every PWM pin at 500 Hz. The
sketch calls Zephyr's `pwm_set_dt()` directly instead.

**Use D2 and D5 only.** D3, D9 and D10 *appear* to work but their 16-bit
counters cannot hold a 20 ms frame at 32 MHz, and the driver does not reject
the request — it returns success and emits the wrong waveform. D9 and D10 are
what Arduino habit reaches for first.

## Why the sketch is C++

donkeycar is a Python project, and the MM1 and the DIY Robocars RC hat both run
CircuitPython on the microcontroller, so this is a fair thing to ask.

Those boards have **no Linux** — the MCU is the only computer, so pulse maths,
scaling, trim and the serial protocol all have to live on it, and CircuitPython
is what makes that bearable. The UNO Q has a Linux side, and Arduino's model for
it is ["Python code running on Linux with C code running as a Sketch on the
microcontroller"](https://docs.arduino.cc/tutorials/uno-q/user-manual/). The
Python didn't go away — it moved to where the board puts it:

| | MM1 / DIY hat | UNO Q |
|---|---|---|
| Pulse maths, scaling, trim | CircuitPython on MCU | **Python** — `parts/unoq.py` |
| Transport to the host | CircuitPython UART | **Python** — `parts/unoq_bridge.py` |
| Pin timing, PWM generation | CircuitPython on MCU | C++ sketch |

What stays in C++ is only what must be on the MCU. The failsafe is the clearest
case: it neutralises the throttle when the host stops commanding, and was
verified by `kill -9`-ing the drive process so `shutdown()` never ran — the MCU
stopped the car on its own. In Python it would die alongside the thing it
protects against.

It is also not currently possible. On the board, `arduino-cli core search
micropython` returns *"No platforms matching your search"*; all 27 App Lab
example sketches are `.ino`; and CircuitPython has no STM32U5 port at all — the
draft PR for the UNO Q (adafruit/circuitpython#10674) was closed without
merging.

**The MicroPython alternative**, for the record: upstream MicroPython v1.28 does
support STM32U5, and flashing it over the Zephyr firmware would let donkeycar's
existing `parts/robohat.py` drive this car over `/dev/ttyHS1` with *no new Python
part at all*. That is the most Pythonic end state and reuses the most code — but
it gives up the Arduino core, `Arduino_RouterBridge` and App Lab, and requires
disabling `arduino-router`, which also drives MCU reset. A spike before a plan.

Fuller version in `ARDUINO_UNO_Q_SETUP.md`.

## What's here

| | |
|---|---|
| `arduino/unoq_rc_hat/` | MCU sketch: servo + ESC out, 3 RC channels in, failsafe |
| `arduino/unoq_i2c_scan/`, `arduino/unoq_pwm_probe/` | diagnostics used to establish the above |
| `parts/unoq_bridge.py` | MsgPack-RPC client for `arduino-router` |
| `parts/unoq.py` | `UnoQRcHatDriver` / `UnoQRcHatController` |
| `templates/complete.py` | `DRIVE_TRAIN_TYPE = "UNOQ"`, `CONTROLLER_TYPE = "UNOQ"` |
| `pyproject.toml` | `unoq` extra |

`ARDUINO_UNO_Q_PLAN.md` and `ARDUINO_UNO_Q_SETUP.md` carry the findings and the
setup guide.

No existing platform is touched: `pins.py` and `actuator.py` are byte-identical
to `main`, and the `pi`, `nano`, `pc` and `macos` extras are unchanged.

## Measured on the board

| | |
|---|---|
| Drive loop | 19.93 Hz (400 steps in 20.07 s) |
| Recording while steering | 19.8 Hz, 200 frames / 200 images in 10.1 s |
| `set_pulse` round trip | 7.4 ms median, 9.1 ms p95 |
| Pushed `rc_input` | 38–39 Hz, 26.4 ms inter-frame |
| CPU | 46% of one core |

50 hardware-free tests (`donkeycar/tests/test_unoq.py`). Suite: 168 passed,
31 skipped.

## Not verified

- **RC receiver input.** The three channels on D3/D4/D6 are written and
  compile, and report a correct no-receiver state, but no receiver was
  available to test against a transmitter.
- Everything in the test plan below.

---

# Test plan

## How to use this plan

**Where you run it.** Everything below runs in a **shell on the UNO Q**,
reached over SSH from a workstation. Not in App Lab's editor, not on your
laptop, and not in a Python prompt. Every line in a `bash` block is a shell
command; if a step needs Python it will be a `python -c` or `python -m`
command you can paste whole.

**Every session starts with this.** Nothing below works without it, and
nothing below repeats it:

```bash
ssh arduino@Edunoq.local            # from your workstation
source ~/env/bin/activate           # donkeycar's venv - no `python` exists outside it
cd ~/projects/donkeycar             # the repo
```

If you did T0.5 the `source` line is already automatic and you can skip it.
`uv` needs no setup in an interactive shell: the image's `~/.bashrc` already
sources `~/.local/bin/env`. (It is *not* on `PATH` for
`ssh host "<command>"`, which skips `.bashrc` — only relevant if you script
against the board.)

**Shorthand.** From T3 onward the bridge CLI is abbreviated. Set it once per
shell:

```bash
B="python -m donkeycar.parts.unoq_bridge"
```

**Recording results.** Tick each box. Anything that fails, record under
**Failures** at the end with the command, the output, and the relevant lines
from `journalctl -u ssh` or `~/mycar/drive.log`.

---

## T0 — Prerequisites

Do these before T1. They are not tests of this PR; they establish the starting
state the tests assume. If any fails, stop — later results will be noise.

- [ ] **T0.1** The UNO Q is **initialized per Arduino's own getting-started
      procedure** (<https://docs.arduino.cc/tutorials/uno-q/user-manual/>):
      Debian boots, the `arduino` user exists and you know its password, and
      the board is on your network. This PR assumes a working stock board and
      changes nothing about it.
- [ ] **T0.2** You can reach a shell from a **separate workstation**:
      ```bash
      ssh arduino@Edunoq.local
      ```
      If you get `Too many authentication failures` before any password
      prompt, your SSH agent is offering more keys than the board's
      `MaxAuthTries` allows — this is a client-side problem, not a board one:
      ```bash
      ssh -o PubkeyAuthentication=no arduino@Edunoq.local
      ```
      Prefer `Edunoq.local` over an IP address; the board's DHCP lease moves.
- [ ] **T0.3** The donkeycar repo is cloned and on this PR's branch:
      ```bash
      git -C ~/projects/donkeycar rev-parse --abbrev-ref HEAD    # -> arduino-uno-q
      ```
      If not: `git clone https://github.com/autorope/donkeycar.git ~/projects/donkeycar && cd ~/projects/donkeycar && git checkout arduino-uno-q`
- [ ] **T0.4** **Create the virtualenv if you do not already have one.**
      Skip if `~/env` exists.
      ```bash
      uv venv --python 3.13 --system-site-packages ~/env
      ```
      No `uv`? The standard library works just as well:
      ```bash
      python3 -m venv --system-site-packages ~/env
      ```
      `--system-site-packages` matches how the board's stock venv is built, so
      Debian-packaged modules stay visible.
- [ ] **T0.5** Verify it, and put `uv` on `PATH`:
      ```bash
      source ~/env/bin/activate && python -V      # -> Python 3.13.x
      uv --version                                # -> uv 0.x
      ```
      If `uv --version` fails, run `. ~/.local/bin/env` first, or use
      `pip` in place of `uv pip` throughout.
- [ ] **T0.6** *(optional, recommended)* **Activate the venv automatically on
      login**, so no step can fail for want of it:
      ```bash
      cat >> ~/.bashrc <<'EOF'

      # Activate the donkeycar virtualenv for interactive shells.
      if [ -f "$HOME/env/bin/activate" ]; then
          . "$HOME/env/bin/activate"
      fi
      EOF
      ```
      Log out and back in, then check:
      ```bash
      echo $VIRTUAL_ENV    # -> /home/arduino/env
      which python         # -> /home/arduino/env/bin/python
      ```
      Debian's `~/.bashrc` returns early for non-interactive shells, so this
      does **not** affect `scp`, `rsync` or `ssh host "<command>"` — verified.
      To undo, delete those lines from `~/.bashrc`.
- [ ] **T0.7** The MCU toolchain and bridge are present:
      ```bash
      arduino-cli board list | grep -i unoq       # -> arduino:zephyr:unoq
      systemctl is-active arduino-router          # -> active
      ```
- [ ] **T0.8** Hardware to hand: the car (chassis, steering servo, ESC,
      motor, battery), **a stand that keeps the wheels off the ground**, a
      separate 5–6 V supply or BEC for the servo, and jumper leads. A
      transmitter and receiver are optional — only T10 uses them.

---

## T1 — Install

Assumes the session setup above has been run.

- [ ] **T1.1** Install the extra:
      ```bash
      uv pip install -e ".[unoq]"
      ```
      → completes with no compiler errors. On a stock image with no
      `build-essential` this is the headline claim: nothing here needs a C
      toolchain.
- [ ] **T1.2** `python -c "import donkeycar"` → prints the version banner.
- [ ] **T1.3** `python -c "import msgpack, cv2; from ai_edge_litert.interpreter import Interpreter; print('ok')"` → `ok`.
- [ ] **T1.4** *(documented step, not needed by the drive train)* I2C group
      access:
      ```bash
      sudo usermod -aG i2c arduino
      ```
      Log out, log back in, then `python -c "open('/dev/i2c-1')"` → no
      `PermissionError`.

## T2 — MCU sketch

- [ ] **T2.1** Compile:
      ```bash
      cd ~/projects/donkeycar/arduino/unoq_rc_hat
      arduino-cli compile --fqbn arduino:zephyr:unoq .
      ```
      → compiles, roughly 12% of program storage.
- [ ] **T2.2** Flash it:
      ```bash
      arduino-cli upload --fqbn arduino:zephyr:unoq .
      ```
      → uploads via SWD. **This overwrites whatever sketch was on the MCU.**
- [ ] **T2.3** `systemctl is-active arduino-router` → `active`.
- [ ] **T2.4** ```bash
      cd ~/projects/donkeycar
      python -m donkeycar.parts.unoq_bridge get_last_pulse
      ```
      → `1500, 1500, 1`.
      *The third field is the failsafe flag. It reads `1` whenever nothing has
      commanded for 500 ms, which is always true after a one-shot CLI call —
      the failsafe working, not a fault. It reads `0` only while something is
      actively driving, as in T6.*
- [ ] **T2.5** `python -m donkeycar.parts.unoq_bridge get_rc`
      → `0, 0, 0, 999999` with no receiver connected. Zeros plus a sentinel
      age is the correct no-receiver reading — **this is a pass**, not an
      error.

## T3 — Servo, wheels off the ground

> **Wheels off the ground for T3 and T4.** The pulse range used here is wider
> than a steering linkage can take. Keep the motor disconnected or the battery
> out until T4.

Wire the steering signal to **D2**, servo ground common with both the board
and the servo's own supply, servo powered externally — not from the board.

```bash
source ~/env/bin/activate
B="python -m donkeycar.parts.unoq_bridge"
```

- [ ] **T3.1** `$B set_pulse 1500 1500` → prints `0`; servo moves to centre.
- [ ] **T3.2** `$B set_pulse 1200 1500` then `$B set_pulse 1800 1500`
      → servo moves each way, smoothly, no buzzing or hunting at rest.
- [ ] **T3.3** `$B set_pulse 100 1500` then `$B get_last_pulse`
      → steering reports `500` — clamped, not 100.
- [ ] **T3.4** `$B set_pulse 4000 1500` then `$B get_last_pulse`
      → steering reports `2500`.
      *(Ignore the third field in T3.3/T3.4; see T2.4. Throttle reads `1500`
      because the failsafe has neutralised it.)*
- [ ] **T3.5** `$B set_pulse 1200 1900`, wait 2 s, then `$B get_last_pulse`
      → `1200, 1500, 1`: throttle back to neutral, tripped flag set.
      **This is the failsafe. It must pass before you reconnect the motor.**

## T4 — ESC and motor, wheels still off the ground

Throttle signal to **D5**. Reconnect motor and battery.

- [ ] **T4.1** Power the ESC with nothing commanding → it arms normally (the
      usual beeps) and the motor does not spin.
- [ ] **T4.2** `$B set_pulse 1500 1600` → motor spins forward briefly;
      `$B set_pulse 1500 1500` → stops.
      *Each CLI call exits, so the failsafe neutralises throttle ~500 ms
      later. A short spin is correct behaviour.*
- [ ] **T4.3** `$B set_pulse 1500 1400` → reverse or brake, per your ESC.
- [ ] **T4.4** Start a sustained command, then kill it:
      ```bash
      python - <<'PY' &
      import time
      from donkeycar.parts.unoq_bridge import UnoQBridge
      b = UnoQBridge()
      while True:
          b.call("set_pulse", 1500, 1600); time.sleep(0.05)
      PY
      sleep 3; kill -9 %1
      ```
      → motor spins while the loop runs, and **stops within ~1 s** of the
      kill. *Record the time taken:* ________

## T5 — Calibration

- [ ] **T5.1** ```bash
      cd ~/mycar 2>/dev/null || donkey createcar --path ~/mycar && cd ~/mycar
      donkey calibrate
      ```
      → steering responds to entered pulse values.
- [ ] **T5.2** Find full-left/right values that do **not** strain the linkage
      or push the wheels past their lock.  Left: ______  Right: ______
- [ ] **T5.3** Find throttle neutral and the forward/reverse limits you are
      happy to drive at.  Fwd: ______  Stop: ______  Rev: ______
- [ ] **T5.4** Put those in `~/mycar/myconfig.py` as `UNOQ_STEERING_MID`,
      `UNOQ_MAX_FORWARD`, `UNOQ_STOPPED_PWM`, `UNOQ_MAX_REVERSE`.
- [ ] **T5.5** Restart and confirm steering sweeps its **calibrated** range
      with no mechanical strain at either end.

**Only now put the car on the ground.**

## T6 — Driving

- [ ] **T6.1** In `~/mycar/myconfig.py` set `CAMERA_TYPE = "CVCAM"` and
      `DRIVE_TRAIN_TYPE = "UNOQ"`, then:
      ```bash
      cd ~/mycar && python manage.py drive
      ```
      → starts with no traceback.
- [ ] **T6.2** Open `http://Edunoq.local:8887/drive` in a browser → live
      camera image.
- [ ] **T6.3** The steering slider moves the wheels, in the correct
      direction. *(If mirrored, note it — there is no `UNOQ_STEERING_INVERTED`
      yet.)*
- [ ] **T6.4** Throttle drives the car forward and back at walking pace.
- [ ] **T6.5** Drive a lap. Steering feels responsive, not laggy or steppy.
- [ ] **T6.6** Ctrl-C → the car stops and does not creep afterwards.
- [ ] **T6.7** Close the browser tab mid-drive while moving slowly → the car
      stops within about a second.

## T7 — Recording and autopilot

- [ ] **T7.1** Drive with recording on → `ls ~/mycar/data/images | wc -l`
      grows steadily.
- [ ] **T7.2** Over a 30 s run the recorded rate is close to 20 Hz.
      *Record:* ______ Hz
- [ ] **T7.3** `donkey tubhist --tub ~/mycar/data` → steering values span the
      range actually driven, not stuck at one value.
- [ ] **T7.4** Copy the tub to a training machine, train, convert to
      `.tflite`.
- [ ] **T7.5** Copy the model back, set `DEFAULT_MODEL_TYPE = "tflite_linear"`,
      then `python manage.py drive --model ~/mycar/models/<model>.tflite`
      → starts with no error.
- [ ] **T7.6** In autopilot the car drives a lap under its own steering.
- [ ] **T7.7** Loop rate stays near 20 Hz with inference running.
      *Record:* ______ Hz

## T8 — Health under load

- [ ] **T8.1** Over a 5-minute drive, `grep -icE "traceback|Bridge send failed|Timed out calling" ~/mycar/drive.log` → `0`.
- [ ] **T8.2** `top` shows the drive process under ~70% of one core.
- [ ] **T8.3** No brownout or reboot: `uptime` still shows the original boot.
- [ ] **T8.4** After the run, `$B get_last_pulse` reports throttle at neutral.

## T9 — Nothing else regressed

- [ ] **T9.1** ```bash
      cd ~/projects/donkeycar && python -m pytest donkeycar/tests/ -q
      ```
      → same results as on `main`, plus the new `test_unoq.py` tests.
- [ ] **T9.2** `git diff main -- donkeycar/parts/pins.py donkeycar/parts/actuator.py`
      → empty. This PR touches no existing platform.
- [ ] **T9.3** On a Raspberry Pi car, `DRIVE_TRAIN_TYPE = "PWM_STEERING_THROTTLE"`
      still drives.

## T10 — RC receiver *(optional; this half is unverified)*

Only if you have a receiver. Channels on **D3, D4, D6**, ground common.
**Wheels off the ground for this section.**

- [ ] **T10.1** With the transmitter on, `$B get_rc` → three plausible pulse
      widths (≈1000–2000) and a small `age_ms`.
- [ ] **T10.2** `$B --listen rc_input --seconds 5` while moving the steering
      stick → the first value sweeps roughly 1000–2000 and the reported rate
      is near 40 Hz.
- [ ] **T10.3** Set `CONTROLLER_TYPE = "UNOQ"` in `myconfig.py` → the
      transmitter drives the car.
- [ ] **T10.4** Turn the transmitter off → **no full reverse and no runaway.**
      Zero channels are ignored by design, but this is the dangerous case.

---

## Results

| Section | Pass / Fail | Notes |
|---|---|---|
| T0 Prerequisites | | |
| T1 Install | | |
| T2 MCU sketch | | |
| T3 Servo | | |
| T4 ESC | | |
| T5 Calibration | | |
| T6 Driving | | |
| T7 Recording / autopilot | | |
| T8 Health | | |
| T9 Regression | | |
| T10 RC *(optional)* | | |

## Failures

_Command run, output seen, and anything from `~/mycar/drive.log`._


TCIII:
When performing T2-MCU after successfully completing T1-Install, the following error occurred during T2.4/5:

T2 — MCU 
T2.1 Compiled the unoq_rc_hat.ino sketch and ~12% of program storage was indicated.
T2.2 Uploaded the compiled sketch successfully.
T2.3 Activated arduino-router and received "active" response.
T2.4/5 When executing either `UnoQBridge().call("get_last_pulse")` or `UnoQBridge().call("get_rc")` using Python, I received the following error: -bash: syntax error near unexpected token '('.

Re: Syntax error
"`UnoQBridge` is not a standard class or method in the official arduino.app_utils library. Must import and interact directly with the global Bridge object rather than instantiating an unauthorized class" according to AI.

TCIII

Ezward:
> When performing T2-MCU after successfully completing T1-Install, the following error occurred during T2.4/5:
> 
> T2 — MCU T2.1 Compiled the unoq_rc_hat.ino sketch and ~12% of program storage was indicated. T2.2 Uploaded the compiled sketch successfully. T2.3 Activated arduino-router and received "active" response. T2.4/5 When executing either `UnoQBridge().call("get_last_pulse")` or `UnoQBridge().call("get_rc")` using Python, I received the following error: -bash: syntax error near unexpected token '('.
> 
> Re: Syntax error "`UnoQBridge` is not a standard class or method in the official arduino.app_utils library. Must import and interact directly with the global Bridge object rather than instantiating an unauthorized class" according to AI.
> 
> TCIII

@TCIII I think I see what may be wrong.  That step needs to be run from within a Python interpreter.  Overall the test instructions suck; they don't indicate that the Arduino must be initialized via Arduino App Lab, they don't indicate that an environment should be created and installed, etc.  So I updated the instructions and we now have a more extensive T0 prerequisites section and each other section should be more explicit.

TCIII:
Reinstalled the UNO Q Debian 13 OS per the Arduino UNO Q User Manual using the Arduino App Lab and successfully completed T0 through T2.
Added the following to T0.4 Create the virtualenv if you do not already have one: `curl -LsSf https://astral.sh/uv/install.sh | sh` to stay consistent with the Donkey Car RPi installation.

TCIII

